PR #226, opened 2021-04-08
style: eslint warnings

vincentaudebert:
## Type

- Enhancement

#### What is expected?

Fix eslint warnings

1. Run `yarn run lint`

2. Fix warnings


.eslintrc.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     // React
     'react/jsx-filename-extension': ['error', { extensions: ['.js', '.mdx'] }],
     'react/destructuring-assignment': 'warn',
-    'react/jsx-props-no-spreading': 'warn',
+    'react/jsx-props-no-spreading': 'off',
```
chambo-e, line 26: We should re-enable that :)

src/components/CreationProgress/index.js
```diff
@@ -147,6 +147,7 @@ const CreationProgress = ({
         }
 
         return (
+          // eslint-disable-next-line react/no-array-index-key
           <React.Fragment key={index}>
```
chambo-e, line 151: Maybe prefix the key, just for the sake of possible conflict with others keys ?

src/components/Placeholder/Line.js
```diff
@@ -14,8 +14,8 @@ const style = width => theme => css`
   background-color: ${theme.colors.gray300};
 `
 
-const Line = props => (
-  <Box css={style(props?.width ?? randomSize())} {...props} />
+const Line = ({ width, ...props }) => (
+  <Box css={style(width ?? randomSize())} {...props} />
```
chambo-e, line 18: This will break when width is defined as it's not passed down to Box :)

src/components/Popper/__stories__/index.stories.mdx
```diff
@@ -94,6 +98,7 @@ import Icon from '../../Icon'
     <Box>
       <Box display="flex" flexDirection="column">
         <Popper
+          key="left-1"
```
chambo-e, line 101: We need keys there ??

src/components/UniversalLink/index.js
```diff
@@ -15,15 +15,23 @@ const needNativeLink = url => {
...
-  const href = props.to || props.href
-  const as = props.as || (needNativeLink(href) ? 'a' : linkComponent)
+  const isBlank = target === '_blank'
+  const rel = propsRel || (isBlank ? 'noopener noreferrer' : undefined)
+  const href = to || propsHref
+  const asValue = propsAs || (needNativeLink(href) ? 'a' : linkComponent)
 
   return (
-    <Box {...props} as={as} href={href} rel={rel}>
+    <Box {...props} as={asValue} href={href} rel={rel}>
```
chambo-e, line 34: we miss the target here :)

chambo-e (changes requested):

src/components/TagsPoplist/index.js
```diff
@@ -55,6 +55,7 @@ const TagsPoplist = ({ tags, threshold, maxLength, maxTagWidth, ...props }) => {
       <Box display="flex" alignItems="center" color="gray700" {...props}>
         {tags.slice(0, visibleTagsCount).map((tag, i) => (
           <Tag
+            // eslint-disable-next-line react/no-array-index-key
```
DorianMaliszewski, line 58: Is there a good solution to drop that ?

DorianMaliszewski (changes requested):

src/components/TagsPoplist/index.js
vincentaudebert: Sadly tag is just a string and we can have twice the same tag so we have to keep the index :(

.eslintrc.js
matthprost: How is it that we warn about prop spreading ? What if we need to spread props into a deep child component ?

src/components/Range/index.js
```diff
@@ -431,6 +431,7 @@ const Range = ({
           offsetTop={offsetTop}
           width={cursorWidth}
           grabbed={grabbedCursor !== undefined}
+          // eslint-disable-next-line react/no-array-index-key
           key={`cursor-${index}`}
```
matthprost, line 435: can't you use value instead of index ? As cursor value is always unique

matthprost (changes requested):

src/components/Popper/__stories__/index.stories.mdx
vincentaudebert: No sorry leftover from tests

src/components/Range/index.js
vincentaudebert: I thought about it but it will change on every cursor move. Which might not be ideal performance wise? 🤔

.eslintrc.js
vincentaudebert: We might want to disable props spreading warning when it's justified yes. @chambo-e ?

matthprost (approved):

.eslintrc.js
chambo-e: Yes, the end goal being to have explicit props on every components.

There might be some case where we'll still need to spread props (component pass through, hoc, ...) but we'll manage case by case :)

src/components/MarkDown/index.js
```diff
@@ -2,16 +2,16 @@ import { css } from '@emotion/react'
...
 import React from 'react'
 import ReactMarkDown from 'react-markdown'
-import { Box } from '../Box'
+import Box from '../Box'
 import Command from '../Command'
 import Link from '../Link'
 import { Typography } from '../Typography'
 
-const headingRenderer = ({ node, ...props }) => {
+const headingRenderer = ({ node, children, ...props }) => {
```
chambo-e, line 10: We miss the children line 20 :)

src/components/Typography/index.js
```diff
@@ -195,15 +195,15 @@ const Text = ({
...
+const TextWithTooltip = ({ children, ...props }) => {
   const isTruncated = useCallback((target = {}) => {
     // If the text is really truncated
     const { offsetWidth, scrollWidth } = target
 
     return offsetWidth < scrollWidth
   }, [])
 
-  const finalStringChildren = recursivelyGetChildrenString(props.children)
+  const finalStringChildren = recursivelyGetChildrenString(children)
```
chambo-e, line 206: We also miss the children on the Text component

chambo-e (changes requested):
Last feedback :)

Could you also change the PR title to another kind ? I don't think we need to generate a version as this does not impact the library interface

chambo-e (approved):

PR #272, opened 2021-04-19
fix(tooltip): fixed prop spreading

matthprost:
## Type

- Bug

#### What is expected?

We should spread props on ReakitTooltip not in his children. This can allow us to define a zIndex and so to be able to show a tooltip inside of a modal.

### DIFF
--- a/src/components/Tooltip/index.js
+++ b/src/components/Tooltip/index.js
@@ -55,7 +55,7 @@ const StyledTooltip = styled(Box, {
 const Tooltip = ({
   animated,
   children,
-  text = '',
+  text,
   placement,
   visible,
   variant,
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/index.js
```diff
@@ -106,8 +106,8 @@ const Tooltip = ({
       <TooltipReference {...tooltip} ref={children.ref}>
         {finalChildren}
       </TooltipReference>
-      <ReakitTooltip {...tooltip}>
-        <StyledTooltip variant={variant} {...props}>
+      <ReakitTooltip {...tooltip} {...props}>
```
chambo-e, line 109: As StyledTooltip is a Box onto which we pass styled props this will break styling no ? 

We could extract zIndex from props and only pass that to the reakit tooltip ?

chambo-e (changes requested):

src/components/Tooltip/index.js
matthprost: @chambo-e It now should be better

chambo-e (approved):

DorianMaliszewski (approved):

PR #278, opened 2021-04-21
fix(date-input): add locale props from date-fns

philibea:
## Type

- Enhancement

### Summarise concisely:

Add possibilty to register locale of date-fns

Add locale props. it's add control

1. Add register of local with currentLocale
2. Add storybook with date-fns
3. Add Storybook with i18n  hook from scaleway-lib

French Locale
Capture d’écran 2021-04-21 à 16 05 37

Spanish Locale

Capture d’écran 2021-04-21 à 16 05 51

Russian Locale
Capture d’écran 2021-04-21 à 16 06 01

I18n Provider
Capture d’écran 2021-04-21 à 16 06 23


src/components/DateInput/index.js
```diff
@@ -224,6 +226,7 @@ DateInput.propTypes = {
   error: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
   format: PropTypes.func,
   label: PropTypes.string,
+  locale: PropTypes.shape({}),
```
chambo-e, line 229: Should be string here :)

philibea: Locale is date-fns locale object and not currentLocale string

chambo-e: Gotcha thanks, just wondering if we could use `locale.code` instead of the `currentLocale` props ?

philibea: Yes you right useless props in this case.

vincentaudebert (approved):

chambo-e (approved):

PR #295, opened 2021-04-27
fix(universal-link): add to when using react-router-dom

philibea:
## Type

- Bug

### Summarize concisely:

When you are using Link from react router dom or specify Link with as props your Universal Link will not work as expected because props to was not forward to our component

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,6 +145,7 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-is": "^17.0.1",
+    "react-router-dom": "^5.2.0",
     "read-pkg": "^6.0.0",
     "regenerator-runtime": "^0.13.7",
     "rollup": "^2.38.0",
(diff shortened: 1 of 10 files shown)


vincentaudebert (approved):

package.json
```diff
@@ -145,6 +145,7 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-is": "^17.0.1",
+    "react-router-dom": "^5.2.0",
```
vincentaudebert, line 148: Weird to add a package just for tests no? But I agree I don't see how to do different :(

vincentaudebert (approved):

package.json
philibea: i can add a test with Link from @reach/router if necessary also

chambo-e (changes requested):

src/helpers/renderWithTheme.js
```diff
@@ -11,10 +11,12 @@ const emotionCache = createCache({
 
 emotionCache.compat = true
 
-export default (component, options) =>
+export default (component, options, modifyTheme = {}) =>
```
chambo-e, line 14: Here you receive theme in option

Why not override the current theme ? Instead of extending it ?

chambo-e (approved):

PR #349, opened 2021-05-12
feat(stepper): new design

matthprost:
## Type

- Feature

#### What is expected?

New design for `Stepper` component.

1. Removed side borders for + and - buttons

2. Changed colors and borders

3. Updated story

Basic:
Screenshot 2021-05-12 at 17 10 47

Input hovered:
Screenshot 2021-05-12 at 17 16 26

Input focused:
Screenshot 2021-05-12 at 17 11 08

Side button hover:
Screenshot 2021-05-12 at 17 16 20

Sizes:
Screenshot 2021-05-12 at 17 11 15

Disabled:
Screenshot 2021-05-12 at 17 11 20

### DIFF
--- a/src/components/Stepper/index.js
+++ b/src/components/Stepper/index.js
@@ -78,7 +78,7 @@ const StyledCenterTouchable = styled(Touchable)`
 const StyledInput = styled.input`
   pointer-events: none;
   color: ${({ theme }) => theme.colors.gray700};
-  background-color: ${({ theme }) => theme.colors.white};
+  background-color: transparent;
   font-size: 16;
   border: none;
   text-align: right;
@@ -183,6 +183,16 @@ const Stepper = ({
   const isMinusDisabled = inputValue <= minValue || disabled
   const isPlusDisabled = inputValue >= maxValue || disabled
 
+  const centerTouchableRef = useRef(null)
+
+  const [inputMaxWidth, setInputMaxWidth] = useState()
+
+  useEffect(() => {
+    setInputMaxWidth(
+      centerTouchableRef.current && centerTouchableRef.current.offsetWidth - 2,
+    )
+  }, [inputMaxWidth, setInputMaxWidth])
+
   return (
     <StyledContainer disabled={disabled} size={size} {...props}>
       <StyledTouchable
@@ -192,11 +202,7 @@ const Stepper = ({
         disabled={isMinusDisabled}
         aria-label="Minus"
       >
-        <StyledIcon
-          name="minus"
-          size={size === 'large' ? 28 : 18}
-          color="gray300"
-        />
+        <StyledIcon name="minus" size={iconSizes[size]} color="gray300" />
       </StyledTouchable>
 
       <StyledCenterTouchable
@@ -207,8 +213,8 @@ const Stepper = ({
             inputRef.current.focus()
           }
         }}
-        px={1}
         aria-label="Input"
+        ref={centerTouchableRef}
       >
         <StyledInput
           disabled={disabled}
@@ -218,7 +224,10 @@ const Stepper = ({
           onFocus={handleOnFocus}
           onKeyPress={onKeyOnlyNumbers}
           ref={inputRef}
-          style={{ width: inputValue.toString().length * 10 + 15 }}
+          style={{
+            width: inputValue.toString().length * 10 + 15,
+            maxWidth: inputMaxWidth || '100%',
+          }}
           value={inputValue.toString()} // A dom element can only have string attributes.
         />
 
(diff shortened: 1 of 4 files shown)


src/components/Stepper/index.js
```diff
@@ -116,21 +91,23 @@ const StyledInput = styled.input`
...
+  border: 1px solid ${({ theme }) => theme.colors.gray300};
+  border-radius: 4px;
   ${({ disabled, theme }) =>
     disabled &&
     `
     > ${StyledTouchable}, ${StyledInput}, ${StyledCenterTouchable} {
       ${disabledStyles({ disabled, theme })}
     }
-  `}
+  `};
```
chambo-e, line 110: 💩 ?

src/components/Stepper/index.js
```diff
@@ -22,80 +23,54 @@ const disabledStyles = ({ disabled, theme }) =>
...
   position: relative;
+  border-radius: 4px;
+  width: ${({ size }) => containerSizes[size] - 10}px;
 
-  :hover,
-  :focus {
-    > svg {
-      fill: ${({ disabled, theme }) => !disabled && theme.colors.primary};
-    }
+  > svg {
```
chambo-e, line 47: Possible to target the Icon instead pls ?

src/components/Stepper/index.js
```diff
@@ -22,80 +23,54 @@ const disabledStyles = ({ disabled, theme }) =>
...
+  height: calc(100% - 8px);
   align-items: center;
   outline: none;
   justify-content: center;
-  border: 1px solid ${({ theme }) => theme.colors.gray350};
-
-  border-left-width: 0;
-  border-right-width: 0;
+  border-radius: 4px;
+  :hover:not([disabled]):not(:focus) {
```
chambo-e, line 66: :wat: 😂 

There is no easier selector ?

matthprost: I can change it for `:hover:not([disabled], :focus)` if you prefer 😄

src/components/Stepper/index.js
matthprost: Should be better @chambo-e

src/components/Stepper/index.js
chambo-e: That's just slightly better but yeah 😄

src/components/Stepper/index.js
```diff
@@ -22,80 +23,56 @@ const disabledStyles = ({ disabled, theme }) =>
...
-  large: '16px',
-  medium: '10px',
-  small: '4px',
+const iconSizes = {
+  large: 26,
+  medium: 24,
+  small: 22,
 }
 
+const StyledIcon = styled(Icon)``
```
DorianMaliszewski, line 37: Just Icon will be sufficient ;)

src/components/Stepper/index.js
```diff
@@ -255,10 +234,7 @@ const Stepper = ({
         disabled={isPlusDisabled}
         aria-label="Plus"
       >
-        <StyledSeparator size={size} style={{ left: 0 }}>
-          |
-        </StyledSeparator>
-        <Icon name="plus" size={size === 'large' ? 28 : 18} color="gray300" />
+        <StyledIcon name="plus" size={iconSizes[size]} />
```
DorianMaliszewski, line 237: Why `minus` icon has not the same size prop ??

matthprost: my bad I forgot to change minus icon too

src/components/Stepper/index.js
```diff
@@ -204,6 +183,16 @@ const Stepper = ({
   const isMinusDisabled = inputValue <= minValue || disabled
   const isPlusDisabled = inputValue >= maxValue || disabled
 
+  const centerTouchableRef = useRef(null)
+
+  const [inputMaxWidth, setInputMaxWidth] = useState()
+
+  useEffect(() => {
```
chambo-e, line 190: This was to correct which behaviour ?

matthprost: Input width but I corrected it with another way thanks to @DorianMaliszewski

DorianMaliszewski (changes requested):
```js
  const [inputValue, setInputValue] = useState(parseIntOr(value, minValue))
  useEffect(() => {
    if (onChange) {
      onChange(inputValue)
    }
  }, [inputValue, onChange])
```

This is not a good way to handle controlled AND uncontrolled components. IMO, the component can have a local value but : 
- when `value` change its should change the local one too. 
- Each call to setInputValue should be followed by `onChange?.(event)` which contain the event of type `ChangeEvent<InputHtmlElement>`

DorianMaliszewski (approved):

chambo-e (approved):

PR #380, opened 2021-05-26
feat: ssr safe rendering

DorianMaliszewski:
## Type

- Enhancement

#### What is expected?

- Scaleway can work with SSR

1. Based on SCW Website and thanks to `yalc`, I successfully add the library to the website.

2. Change when needed call to Client API instead of Node API

### DIFF
--- a/src/components/Toaster/index.js
+++ b/src/components/Toaster/index.js
@@ -1,15 +1,13 @@
-import { ClassNames, css, useTheme } from '@emotion/react'
+import { ClassNames, Global, css, useTheme } from '@emotion/react'
 import PropTypes from 'prop-types'
 import React from 'react'
 import {
   ToastContainer as BaseToastContainer,
   toast as baseToast,
 } from 'react-toastify'
-import { injectStyle } from 'react-toastify/dist/inject-style'
+import style from 'react-toastify/dist/ReactToastify.min.css'
 import Icon from '../Icon'
 import NotificationBar from '../NotificationBar'
-
-injectStyle()
 
 const PREFIX = '.Toastify'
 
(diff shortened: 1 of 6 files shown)


matthprost (approved):
🎉

src/components/Toaster/index.js
```diff
@@ -5,11 +5,9 @@ import {
   ToastContainer as BaseToastContainer,
   toast as baseToast,
 } from 'react-toastify'
-import { injectStyle } from 'react-toastify/dist/inject-style'
 import Icon from '../Icon'
 import NotificationBar from '../NotificationBar'
-
-injectStyle()
+import 'react-toastify/dist/ReactToastify.css'
```
chambo-e, line 10: Does this correctly inline the css in the bundle ?

philibea, line 10: I got a doubt on this change, i think we use to be like this before but it's doesn't work properly on other projects

DorianMaliszewski: Do I need to do the same as you did with PhoneInput css ?

chambo-e: If this is not inlined yes I think you'll need to

But what's wrong with the injectStyle way ?

DorianMaliszewski: injectStyles use document object under the hood which is not supported in SSR/SSG mode. The other way to keep it is to put injectStyles() in useLayoutEffect/useEffect on mount lifecycle

DorianMaliszewski: injectStyle use document object under the hook (See https://github.com/fkhadra/style2js/blob/main/src/style2js.ts). NextJS doesn't like it 😅

chambo-e: I just checked and css seems to be correctly inlined in the bundle, so no issue on my side :)

chambo-e (approved):

matthprost (approved):

PR #383, opened 2021-05-26
docs: updated all B components doc

matthprost:
## Type

- Documentation

#### What is expected?

Update documentation of all components that start with letter `B`.

### DIFF
--- a/src/components/BorderedBox/index.js
+++ b/src/components/BorderedBox/index.js
@@ -51,6 +51,9 @@ const BorderedBox = ({ children, variant, to, href, as, ...props }) => (
 )
 
 BorderedBox.propTypes = {
+  /**
+   * Allow you to change tag type.
+   */
   as: PropTypes.string,
   children: PropTypes.node.isRequired,
   href: PropTypes.string,
--- a/src/components/BorderedBox/index.js
+++ b/src/components/BorderedBox/index.js
@@ -52,7 +52,7 @@ const BorderedBox = ({ children, variant, to, href, as, ...props }) => (
 
 BorderedBox.propTypes = {
   /**
-   * Allow you to change tag type.
+   * Allow you to use a different HTML tag or a different custom component but keep all the styling
    */
   as: PropTypes.string,
   children: PropTypes.node.isRequired,
--- a/src/components/Box/__stories__/index.stories.mdx
+++ b/src/components/Box/__stories__/index.stories.mdx
@@ -12,13 +12,23 @@ import Box from '..'
 # Box
 
 <Description>
-  System based Box, see [xstyled docs](https://xstyled.dev/docs/).
+  Box is base on [xstyled](https://xstyled.dev/docs/) `x.div`. You can use a
+  `Box` for wrapping your components in it and apply style on them.
+</Description>
+
+## Basic
+
+<Description>
+  In the example we have a `Box` with primary background, a height of `100px`,
+  margin x `auto` and padding of `16px`.
 </Description>
 
 <Canvas>
-  <Story name="basic">
+  <Story name="Basic">
     <Box backgroundColor="primary" height={100} mx="auto" p={2} />
   </Story>
 </Canvas>
+
+## API
 
 <ArgsTable of={Box} />
--- a/src/components/Box/__stories__/index.stories.mdx
+++ b/src/components/Box/__stories__/index.stories.mdx
@@ -16,8 +16,6 @@ import Box from '..'
   `Box` for wrapping your components in it and apply style on them.
 </Description>
 
-## Basic
-
 <Description>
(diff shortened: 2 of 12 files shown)


src/components/Button/index.js
```diff
@@ -364,6 +364,9 @@ const propTypes = {
   download: PropTypes.string,
   extend: PropTypes.bool,
   href: PropTypes.string,
+  /**
+   * Name of the icon. All [icons](/?path=/docs/components-icon) are supported.
+   */
   icon: PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.func]),
   iconPosition: PropTypes.oneOf(['left', 'right']),
   innerRef: PropTypes.func,
```
DorianMaliszewski, line 372: What is this used for ?

src/components/Breadcrumbs/__stories__/index.stories.mdx
```diff
@@ -84,23 +54,21 @@ import Breadcrumbs, { breadcrumbsVariants } from '..'
...
+## API
+
+### `<Breadcrumbs>`
+
 <ArgsTable of={Breadcrumbs} />
 
-<ArgsTable of={Breadcrumbs.Item} />
+### `<Breadcrumbs.Item>`
+
+<ArgsTable of={Item} />
```
chambo-e, line 74: Because if you use `Breadcrumbs.Item` it will not be able to infer the propTypes ?

src/components/Button/index.js
DorianMaliszewski: Same for : ` displayProgressOnlyl` When must say that user need to have `progress` prop too too make it works

DorianMaliszewski (changes requested):

src/components/BorderedBox/index.js
```diff
@@ -51,6 +51,9 @@ const BorderedBox = ({ children, variant, to, href, as, ...props }) => (
 )
 
 BorderedBox.propTypes = {
+  /**
+   * Allow you to change tag type.
```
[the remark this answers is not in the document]
DorianMaliszewski: `Allow you use a different HTML tag or a different custom component but keep all the styling`

src/components/Breadcrumbs/__stories__/index.stories.mdx
matthprost: Exactly, like I did for estimated cost

src/components/Button/index.js
matthprost: `innerRef` seems to be a ref you could associate with button child but as we pass props into it we can directly do `ref={aRef}`

src/components/Box/__stories__/index.stories.mdx
```diff
@@ -12,13 +12,23 @@ import Box from '..'
...
 <Canvas>
-  <Story name="basic">
+  <Story name="Basic">
     <Box backgroundColor="primary" height={100} mx="auto" p={2} />
   </Story>
 </Canvas>
 
+## API
+
 <ArgsTable of={Box} />
```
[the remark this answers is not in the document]
matthprost: Tried to, in the end it seems like we cannot get properties of x.div

DorianMaliszewski:
BorderedBox has `basic` instead of `Basic` story name, Difference between sidebar anchros and titles.
Breadcrumbs first story is Active index instead of `Basic` is sidebar.
Button difference between last story name and last sidebar anchor

src/components/Button/index.js
```diff
@@ -381,8 +382,8 @@ FwdButton.defaultProps = defaultProps
 
 FwdButton.propTypes = propTypes
 
-function forwardRef(props, ref) {
-  return <FwdButton {...props} innerRef={ref} />
+function forwardRef(props) {
```
chambo-e, line 385: Hmm, so the forwardRef is useless now ?

This will break every disclosure :)

chambo-e (changes requested):

DorianMaliszewski:
Capture d’écran 2021-06-01 à 11 57 09
Rename the following :
- As Link -> Link ? 
- Horizontal + Vertical -> Variants

DorianMaliszewski (changes requested):
See comments

DorianMaliszewski (changes requested):
I still have this : Bad order between sidebar and stories.
Capture d’écran 2021-06-01 à 15 51 44

Box still have only height and width in it's API : Use https://xstyled.dev/docs/prop-types/
Capture d’écran 2021-06-01 à 15 53 06

Boxer : Still no ArgsTable
Capture d’écran 2021-06-01 à 15 54 30

DorianMaliszewski (approved):

chambo-e (approved):

PR #385, opened 2021-05-27
feat(accessibility): enable storybook accessibility

matthprost:
## Type

- Enhancement

#### What is expected?

To be able to see accessibility review of each components.

1. Enabled canvas view with accessibility only

2. Disabled all other features for the moment (background color, responsive, grid, etc.).


philibea (approved):

DorianMaliszewski (approved):

chambo-e (commented):
Would it be possible to only enable it in dev ?

vincentaudebert (approved):

matthprost:
> Would it be possible to only enable it in dev ?

@chambo-e I added production and development environment using storybook env variables, you can check on it if config seems fine to you

chambo-e (approved):

PR #387, opened 2021-05-27
docs: updated all C components doc

matthprost:
## Type

- Documentation

#### What is expected?

To have better documentation for all starting C letter components.

⚠️ CreationProgress not include as full refacto is needed ⚠️ 

1. Updated stories

2. Fixed checkbox when progress is true not to be clickable

### DIFF
--- a/src/components/Checkbox/index.js
+++ b/src/components/Checkbox/index.js
@@ -93,7 +93,9 @@ const Checkbox = ({
           {...checkbox}
           hasChildren={hasChildren}
           size={size}
-          onChange={onChange}
+          onChange={e => {
+            if (!progress) onChange(e)
+          }}
           onFocus={onFocus}
           onBlur={onBlur}
           disabled={disabled}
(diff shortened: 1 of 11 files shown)


src/components/Checkbox/index.js
```diff
@@ -93,7 +93,9 @@ const Checkbox = ({
           {...checkbox}
           hasChildren={hasChildren}
           size={size}
-          onChange={onChange}
+          onChange={e => {
```
chambo-e, line 96: Maybe add a test for this use case ?

DorianMaliszewski (changes requested):
Differences between sidebar and story order/name

DorianMaliszewski:
Capture d’écran 2021-06-01 à 11 54 53
Still there

DorianMaliszewski (changes requested):
See comments

matthprost:
> Capture d’écran 2021-06-01 à 11 54 53
> 
> Still there

Screenshot 2021-06-01 at 12 24 01

I think you may have cache

DorianMaliszewski:
When clearing the cache and go to the chart it's still there. What did you have @chambo-e ?

chambo-e (approved):

DorianMaliszewski (approved):

PR #415, opened 2021-06-07
chore(lint): sort keys ont rollup config file

philibea:
## Type

- Refactor

#### What is expected?

- Eslint warning on sort keys
- Change report.html to index.html in .report directory


chambo-e (approved):
To be honest this makes no sense 😄 

To have plugins before presets in babel config, external before defining input or output, ...

But meh, this is only cosmetic 👍

PR #416, opened 2021-06-07
fix(button): button with as link doesn t forward disabled property

philibea:
…rrectly

## Type

- Bug

#### What is expected?

Disabled button with as=Link,a doesn't work

Checking if component is disabled before forward as property

In this case, Button is disabled.
Capture d’écran 2021-06-07 à 18 14 52

### DIFF
--- a/src/components/Button/__stories__/index.stories.mdx
+++ b/src/components/Button/__stories__/index.stories.mdx
@@ -191,6 +191,19 @@ import Button, { buttonSizes, buttonVariants } from '..'
   </Story>
 </Canvas>
 
+<Description>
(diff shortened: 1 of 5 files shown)


chambo-e (approved):

src/components/Button/__stories__/index.stories.mdx
```diff
@@ -191,6 +191,19 @@ import Button, { buttonSizes, buttonVariants } from '..'
   </Story>
 </Canvas>
 
+<Description>
+  In case, you would like to disabled a link, this link will be transform into a
```
vincentaudebert, line 195: transformed

vincentaudebert (approved):

PR #421, opened 2021-06-09
refactor(box): merge box and borderedbox

vincentaudebert:
## Type

- Refactor

#### What is expected?

Bordered doesn't exist anymore and is a variant of box

(Describe what you did)

1. Nuke BorderedBox

2. Add bordered prop to Box

3. Link variant on BorderedBox doesn't exist anymore, customize a Link

4. Variant horizontal/vertical wasn't doing any thing


src/components/RadioBorderedBox/index.js
```diff
@@ -3,12 +3,13 @@ import { transparentize } from 'polished'
...
 import React from 'react'
 import Badge from '../Badge'
-import BorderedBox from '../BorderedBox'
 import Box from '../Box'
 import Radio from '../Radio'
 import Typography from '../Typography'
 
-const StyledBorderedBox = styled(BorderedBox)`
+const StyledBox = styled(Box, {
+  shouldForwardProp: prop => !['bordered'].includes(prop),
```
chambo-e, line 11: You need to pass the bordered props to Box no ?

src/components/Box/index.js
```diff
@@ -1,19 +1,34 @@
...
+  border: 1px solid ${theme.colors.gray350};
+`
+
+const StyledBox = styled(x.div, {
+  shouldForwardProp: prop => !['bordered'].includes(prop),
+})`
+  ${({ bordered }) => (bordered ? borderedStyles : null)}
+`
+const Box = forwardRef(({ width, height, bordered, ...props }, ref) => (
+  <StyledBox ref={ref} w={width} h={height} bordered={bordered} {...props} />
```
chambo-e, line 19: If you pass the class directly would this avoid the noise in the snapshots ?

```js
    <w.div ref={ref} w={width} h={height} css={bordered && borderedStyles} {...props} />
```

chambo-e (changes requested):

src/components/Box/index.js
vincentaudebert: probleme with having the css class directly instead of styled component is the overwriting. With a css class directly then the `p, border, etc.` have no effect on what's inside the borderedStyles. So you can't overwrite the padding and border. Which is a problem I reckon :(

src/components/RadioBorderedBox/index.js
vincentaudebert: Yes of course sorry :)

src/components/Box/index.js
chambo-e: Alrighty then :)

chambo-e (approved):

philibea (approved):

PR #437, opened 2021-06-14
feat: merge Action into Button

chambo-e:
## Type

- Refactor

### Summarise concisely:

```js
<Action name="lock" /> 
// =>
<Button action icon="lock" />
```
```js
<Action variant="rounded" /> 
// =>
<Button action="rounded" icon="lock" />
```

Also added tooltip support to Button

```js
<Button tootltip="I will be in a tooltip">Click Me</Button>
```

And iconSize
```js
<Button action="rounded" icon="lock" iconSize={32} />
```


philibea (approved):

matthprost:
I think it will be more simple if we just check if button contains only icon then we remove padding ? Cause now I see no difference between action and simple button icon. And for the rounded part it should be a variant

matthprost (changes requested):
See up comment

chambo-e:
> I think it will be more simple if we just check that button contains only icon then we remove padding ? Cause now I see no difference between action and simple button icon. Plus, it create a prop that is meaningless :/
> For the rounded part it should be a variant.

For the rounded prop we can't use a variant because we could have a `rounded secondary` button for example

For the icon only button there will be an issue with existing component which only have an icon and need a padding :)

src/components/Button/index.js
```diff
@@ -289,96 +308,88 @@ function FwdButton({
...
   ]),
   disabled: PropTypes.bool,
-  download: PropTypes.string,
+  download: PropTypes.bool,
   extend: PropTypes.bool,
   href: PropTypes.string,
   /**
    * Name of the icon. All [icons](/?path=/docs/components-icon) are supported.
    */
   icon: PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.func]),
```
matthprost, line 390:
```suggestion
  icon: PropTypes.node,
```

matthprost (changes requested):

matthprost (approved):

PR #454, opened 2021-06-17
feat: add pagination

DorianMaliszewski:
## Type

- Feature

#### What is expected?

- Add pagination component
- Implement List with the new Pagination component

1. Add Pagination

2. Improve List component

3. Add UT for Pagination and List

Page | Before | After
:-   |  :-:   | -:
url | screenshot | screenshot
url | screenshot | screenshot

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,6 +22,7 @@
     // React
     "react/jsx-filename-extension": ["error", { "extensions": [".js", ".mdx"] }],
     "react/jsx-props-no-spreading": "warn",
+    "react/sort-prop-types": "off",
 
     // emotion
     "@emotion/styled-import": "error",
--- a/src/components/List/__tests__/index.js
+++ b/src/components/List/__tests__/index.js
@@ -4,6 +4,7 @@ import React, { useEffect } from 'react'
 import List from '..'
 import shouldMatchEmotionSnapshot from '../../../helpers/shouldMatchEmotionSnapshot'
 import { generateData } from '../../../mocks/list'
+import { getUUID } from '../../../utils'
 
 describe('List', () => {
   test('should render correctly', () =>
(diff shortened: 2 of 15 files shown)


src/components/List/__tests__/index.js
```diff
@@ -1481,4 +1534,126 @@ describe('List', () => {
...
+    jest.spyOn(global.Math, 'random').mockReturnValue(0.4155913669444804)
+    await shouldMatchEmotionSnapshot(
+      <List
+        perPage={5}
+        onLoadPage={({ perPage }) =>
+          new Promise(resolve => {
+            setTimeout(() => {
+              const newData = generateData(perPage, getUUID())
+              resolve(newData)
+            }, 3000)
```
matthprost, line 1548: Can't we decrease this value to reduce time of unit testing ?

src/components/Pagination/__tests__/index.test.js
```diff
@@ -0,0 +1,379 @@
...
+import shouldMatchEmotionSnapshot from '../../../helpers/shouldMatchEmotionSnapshot'
+import getPageNumbers from '../getPageNumbers'
+
+const ExampleChildren = () => {
+  const { pageData } = usePagination()
+
+  return (
+    <ul>
+      {pageData.map(item => (
+        <li key={item}>{item}</li>
```
matthprost, line 14: isn't `item` an array ?

src/components/Pagination/index.js
```diff
@@ -0,0 +1,360 @@
...
+    current
+      ? `
+    text-decoration: underline;
+    &:hover {
+      text-decoration: underline;
+    }
+  `
+      : ''}
+`
+const MAX_PAGE_TAB = 5
```
matthprost, line 39: It can be interesting to make it as property of `Pagination` so users can change it if the want

matthprost (changes requested):

src/components/Pagination/__tests__/index.test.js
DorianMaliszewski: No, because `pageData` is the array of the current page @matthprost

.eslintrc
```diff
@@ -22,6 +22,7 @@
     // React
     "react/jsx-filename-extension": ["error", { "extensions": [".js", ".mdx"] }],
     "react/jsx-props-no-spreading": "warn",
+    "react/sort-prop-types": "off",
```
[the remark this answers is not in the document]
DorianMaliszewski: Conflict with `sort-keys` rule 😅  @philibea

src/components/Pagination/index.js
DorianMaliszewski: If the user want more page he can pass MiddleComponent prop to change the existing one

matthprost: What if he wants same as default but with just more tabs ?

src/components/List/index.js
```diff
@@ -358,6 +473,27 @@ List.propTypes = {
...
+  /**
+   * @param {{page, perPage, sort, order}} params The params to sort the list
+   */
+  onSortClick: PropTypes.func,
+  /**
+   * Initial page (need perPage)
+   */
+  page: PropTypes.number,
+  /**
+   * If you known the page you have (need perPage)
```
matthprost, line 490:
```suggestion
   * If you known the page you have (needs `perPage`)
```

matthprost (changes requested):

src/components/Pagination/__tests__/index.test.js
```diff
@@ -0,0 +1,392 @@
...
+          const backButton = node.getByRole('button', { name: 'Back' })
+          const firstButton = node.getByRole('button', { name: 'First' })
+          const lastButton = node.getByRole('button', { name: 'Last' })
+          userEvent.click(nextButton)
+          userEvent.click(backButton)
+          userEvent.click(lastButton)
+          userEvent.click(firstButton)
+          const page3Button = node.getByRole('button', { name: 'Page 3' })
+          userEvent.click(page3Button)
+          userEvent.click(page3Button)
```
matthprost, line 184: Is it a mistake to make it click twice ?

matthprost (changes requested):

src/components/Pagination/index.js
```diff
@@ -0,0 +1,366 @@
...
+        mr={1}
+        disabled={page === 1 || isLoadingPage}
+        onClick={goToPreviousPage}
+      >
+        Back
+      </Button>
+      {pageNumbersToDisplay.map(pageNumber => (
+        <StyledPageButton
+          aria-label={`Page ${pageNumber}`}
+          key={pageNumber}
```
matthprost, line 86: Maybe add a prefix here, if there is another component with same key it will  create problems

matthprost (changes requested):

.eslintrc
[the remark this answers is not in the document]
philibea: Can we add a comment on this line only and not remove the rules on the project ?

src/components/Pagination/__tests__/index.test.js
DorianMaliszewski: No mistake, to handle when you click on the same page it allow us to don't update the state so no rerender :)

philibea (approved):

matthprost (approved):

PR #467, opened 2021-06-24
refactor: merge icon, menuicon and categoryicon

DorianMaliszewski:
## Type

- Refactor

#### What is expected?

Merge Icon, MenuIcon and CategoryIcon

1. Add MenuIcon and CategoryIcon in Icon


src/components/Icon/index.js
```diff
@@ -237,6 +437,22 @@ const ICONS = {
 
 export const icons = Object.keys(ICONS)
 
+const smallViewBoxIcons = [
```
chambo-e, line 440: Not related to this PR but I wonder if there would be a better way to handle svg than inlining them like this :)

chambo-e (approved):

matthprost (approved):

src/components/Icon/index.js
DorianMaliszewski: IDK, `React.lazy(() => import('./icons/myicon.svg'))` and wrap it in a suspense may be ? I really don't have good idea because there is no many relevant way to do this 😅

PR #473, opened 2021-06-28
fix(merge): merged NotificationBar, IconMessage, Information into Alert

matthprost:
## Type

- Migration

Deleted NotificationBar, IconMessage and Information component. Instead I migrated all in Alert component by adding new properties so it will be easier to migrate.

Tests and storybook have been updated.

Screenshot 2021-07-01 at 12 34 52


src/components/Alert/__stories__/index.stories.mdx
```diff
@@ -4,82 +4,95 @@ import {
   Description,
   Meta,
   Story,
-} from '@storybook/addon-docs'
-import Alert from '..'
+} from '@storybook/addon-docs/blocks'
+import Alert, { AlertVariants } from '..'
```
DorianMaliszewski, line 8: constant should not begin with an uppercase letter no ?

src/components/Alert/index.js
```diff
@@ -1,50 +1,111 @@
...
+        {title && <Title color={color} text={title} />}
+        <Typography variant="bodyA" color={color} ml={2}>
+          {children}
+        </Typography>
+      </StyledBox>
+    </StyledContainer>
+  )
+}
+
+export const notificationVariants = Object.keys(variants)
```
DorianMaliszewski, line 81: We should not have multiple constants for variants

DorianMaliszewski (changes requested):

src/components/Alert/index.js
```diff
@@ -1,50 +1,111 @@
...
+  title,
+  hasBackground,
+  ...props
+}) => {
+  const theme = useTheme()
+  const color = theme.colors[variant]
+
+  return (
+    <StyledContainer variant={variant} hasBackground={hasBackground} {...props}>
+      <Icon color={color} name={icon} size={iconSize} />
```
DorianMaliszewski, line 70: Is not a good idea to add an icon based on variant by default ?

src/components/Alert/index.js
```diff
@@ -1,50 +1,111 @@
...
+  icon,
+  title,
+  hasBackground,
+  ...props
+}) => {
+  const theme = useTheme()
+  const color = theme.colors[variant]
+
+  return (
+    <StyledContainer variant={variant} hasBackground={hasBackground} {...props}>
```
DorianMaliszewski, line 69: hasBackground is too much, we should add a background based on variant otherwise the user should pass `background="transparent"` to drop it no ?

src/components/Alert/index.js
matthprost: Great idea ! I haven't tough of that :D

src/components/Alert/index.js
matthprost: Before it was done with a variant but I think we cannot put every single design change into variant properties. 

So there are two options here:
- We add `background="transparent"` prop
- We add `type` prop, it will define if its `info`, `warning`, etc. and we change `variant` prop to define what design the alert is gonna be, like `outlined`, `filled`, `transparent`, etc. This option is a bit more pain in the ass but it makes a way more consistent component with logical properties.
In this option it will look like:
```js
<Alert type="warning" variant="transparent">...</Alert>
```

@DorianMaliszewski @chambo-e @philibea

matthprost: @DorianMaliszewski done!

src/components/Alert/__stories__/index.stories.mdx
```diff
@@ -4,82 +4,102 @@ import {
...
   Meta,
   Story,
-} from '@storybook/addon-docs'
-import Alert from '..'
+} from '@storybook/addon-docs/blocks'
+import Alert, { alertTypes, alertVariants } from '..'
+import { Box } from '../..'
 
-<Meta title="Components/Feedback/Alert" component={Alert} />
+<Meta title="Components/Alert" component={Alert} />
```
philibea, line 11: title should be follow our new ranking
``
```suggestion
<Meta title="Components/Feedback/Alert" component={Alert} />
```

philibea (changes requested):

chambo-e (approved):

philibea (approved):

PR #508, opened 2021-07-06
fix: tooltip display

DorianMaliszewski:
## Type

- Bug

#### What is expected?

Tooltip should be displayed near the children

1. Cleanup

2. Bug fix


chambo-e (commented):
I like that

But that means we won't support a function as a text ? Maybe need to update the text proptypes :)

chambo-e (approved):

philibea (approved):

PR #532, opened 2021-07-12
docs(storybook): fixed argstable

matthprost:
## Type

- Documentation

#### What is expected?

To be able to show prop of each components in storybook argstable.

1. Fixed all missing props in documentation

### DIFF
--- a/src/components/Boxer/index.js
+++ b/src/components/Boxer/index.js
@@ -3,12 +3,14 @@ import React from 'react'
 import flattenChildren from 'react-flatten-children'
 import Box from '../Box'
 
-const Boxer = ({ children, ...props }) =>
+const SubBoxer = ({ children, ...props }) =>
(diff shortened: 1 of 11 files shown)


DorianMaliszewski (changes requested):
Capture d’écran 2021-07-22 à 11 13 30
Capture d’écran 2021-07-22 à 11 14 00

Still this two components which doesn't work and it's good imo

matthprost:
> Capture d’écran 2021-07-22 à 11 13 30
> 
> Capture d’écran 2021-07-22 à 11 14 00
> 
> Still this two components which doesn't work and it's good imo

@DorianMaliszewski Its because they don't have any props, they are just a styled box without any entry prop

chambo-e (approved):

DorianMaliszewski (approved):

PR #575, opened 2021-07-26
fix: production story order

DorianMaliszewski:
## Type

- Bug

#### What is expected?

Story should be displayed in declaration order

1. Changes in preview.js


matthprost (approved):

DorianMaliszewski:
Can't resolve see https://github.com/storybookjs/storybook/issues/15574

PR #580, opened 2021-07-26
docs: updated all t components documentation

matthprost:
## Type

- Documentation

Updated documentation of:

- TabGroup
- Table
- Tag
- Tags
- TagsPoplist
- TextBox
- TimeInput
- Toaster
- Tooltip
- TooltipIcon
- Touchable
- Typography

### DIFF
--- a/src/components/Touchable/index.js
+++ b/src/components/Touchable/index.js
@@ -22,7 +22,7 @@ const styles = {
   `,
 }
 
-const BaseTouchable = ({
+export const BaseTouchable = ({
   innerRef,
   disabled,
   activeOpacity,
--- a/src/components/Touchable/index.js
+++ b/src/components/Touchable/index.js
@@ -22,7 +22,7 @@ const styles = {
   `,
 }
 
-export const BaseTouchable = ({
+const BaseTouchable = ({
   innerRef,
   disabled,
   activeOpacity,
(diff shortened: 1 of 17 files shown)


src/components/Touchable/index.js
```diff
@@ -22,7 +22,7 @@ const styles = {
   `,
 }
 
-const BaseTouchable = ({
+export const BaseTouchable = ({
```
chambo-e, line 25: Maybe move the propTypes to Touchable instead ?

src/components/Typography/index.js
```diff
@@ -276,8 +276,8 @@ Typography.defaultProps = {
   variant: 'default',
 }
 
-Typography.defaultProps = Text.defaultProps
+OriginalTypography.defaultProps = Text.defaultProps
 
-const MemoizedTypography = memo(Typography)
+const MemoizedTypography = memo(OriginalTypography)
```
chambo-e, line 281: Same here :)

src/components/Touchable/index.js
matthprost: I did a const for propTypes and defaultProps :)

chambo-e (approved):

philibea (approved):

PR #595, opened 2021-08-01
feat: typescript migration of alert and icon

DorianMaliszewski:
## Type

- Migration

#### What is expected?

Migration of Alert and Icon

1. start from my first pr

2. Migrate Alert

3. Migrate Icon


src/types.ts
```diff
@@ -0,0 +1,13 @@
+export type XStyledProps = {
+  as?: string | unknown
```
matthprost, line 2: Why unknown ? Shouldn't it be string only ?

matthprost (changes requested):

src/components/ActionBar/index.tsx
```diff
@@ -26,18 +26,18 @@ const StyledBox = styled(Box)<{ rank: number }>`
 `
 
 type ActionBarProps = {
-  children: React.ReactNode
+  children: ReactNode
```
adriengibrat, line 29:
```suggestion
```
`FunctionComponent` (or it's alias `FC`) implies `children`: no need to specify it in Props type definition

src/components/Avatar/index.tsx
```diff
@@ -1,11 +1,12 @@
 import styled from '@emotion/styled'
 import PropTypes from 'prop-types'
```
adriengibrat, line 2: migrate to Ts & keep proptypes ?

Maybe https://openbase.com/js/babel-plugin-typescript-to-proptypes could help DRY ;)

src/components/ActionBar/index.tsx
chambo-e: There is a bug in react types which currently force us to specify children in props

I submitted a PR to correct this https://github.com/DefinitelyTyped/DefinitelyTyped/pull/54955 :)

src/components/Avatar/index.tsx
chambo-e: You're right, we've decided to keep them both while migrating but an automated way to convert types into proptypes would be of great value when completed :)

src/components/ActionBar/index.tsx
chambo-e: I'm not gonna have it ready anytime soon, I think we'll have to move forward without it for now 😢

matthprost (approved):

philibea (approved):

PR #618, opened 2021-08-09
chore:  remove Box component typing hacks

adriengibrat:
## Summary
Improve Box component typings

& use snapshots classname generation based on style content to reduce false positive snaphosts test failure.

## Type
- Enhancement

### Summarise concisely:

A visual testing was done locally with [loki](https://loki.js.org/getting-started.html) to ensure there was no regression.

#### What is expected?

No change at all, only component typings refactor. 

Review c1ba7c0, the other commit only contains snapshot updates.


src/components/Box/index.tsx
```diff
@@ -1,52 +1,50 @@
...
 
-type BoxType = typeof forwardType & {
-  withComponent?: (
-    element: string | ElementType<unknown>,
-  ) => FunctionComponent<Record<string, unknown>>
-}
+type AllBoxProps = BoxProps &
+  XStyledProps &
+  HTMLAttributes<Element> &
+  ClassAttributes<Element>
```
chambo-e, line 34: RefAttributes instead :)

src/components/Box/index.tsx
```diff
@@ -1,52 +1,50 @@
...
 
-const Box: BoxType = forwardRef<Element, Props>(
-  ({ width, height, bordered = false, ...props }, ref) => (
-    // @ts-expect-error As we won't know the Element kind we can't assume that Ref will be a Element
-    <StyledBox ref={ref} w={width} h={height} bordered={bordered} {...props} />
+const Box = forwardRef(
+  // map props `width` to w & `height` to h, see https://xstyled.dev/docs/upgrade-guide/#width-becomes-w-height-becomes-h
+  ({ width, height, ...props }: BoxProps, ref) => (
+    // @ts-expect-error As we won't know the Element kind we can't assume what Ref will be
+    <StyledBox ref={ref} w={width} h={height} {...props} /> // eslint-disable-line react/jsx-props-no-spreading
```
chambo-e, line 40: We tend to not disable known eslint warnings as we want them to be explicit. This allows us to track them and keep in mind that we have to correct them :)

adriengibrat: done, ammended c1ba7c0

src/components/Box/index.tsx
adriengibrat: done, amended c1ba7c0

matthprost (approved):

src/components/Box/index.tsx
```diff
@@ -1,54 +1,50 @@
...
-  React.HTMLAttributes<HTMLElement>
+}
 
-const forwardType = forwardRef<Element, BoxProps>(() => null)
+const StyledBox = styled(x.div, {
+  shouldForwardProp: prop => prop !== 'bordered',
+})<BoxProps>`
+  ${({ bordered, theme }) =>
+    bordered
+      ? css`
```
DorianMaliszewski, line 23: Do we need to put styled and css ? This can be useful when there are some conditional styles but here, is the import worse ??? 😅

adriengibrat:
No significant improvement in replacing a hack by another, but the emotion snapshot config is intersting.
Closing this PR in favor of https://github.com/scaleway/scaleway-ui/pull/641

PR #625, opened 2021-08-10
fix(Typography): correct missing event in reakit event handlers

chambo-e:
## Summary

Also removed duplicate emotion.d.ts and some unused variables

## Type

- Bug

### DIFF
--- a/src/components/Typography/index.tsx
+++ b/src/components/Typography/index.tsx
@@ -265,6 +265,8 @@ const Text = forwardRef(
   ),
 )
 
+type EventHandler<T> = (ev: T) => void
+
 const TextWithTooltip = ({ children, ...props }: TypographyProps) => {
   const isTruncated = useCallback((target = {}) => {
     // If the text is really truncated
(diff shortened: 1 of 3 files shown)


src/components/Typography/index.tsx
```diff
@@ -280,17 +282,24 @@ const TextWithTooltip = ({ children, ...props }: TypographyProps) => {
...
   return (
     <Tooltip text={finalStringChildren}>
-      {({ onMouseEnter = () => {}, onFocus = () => {}, ...tooltipProps }) => (
+      {({
+        onMouseEnter = () => {},
+        onFocus = () => {},
+        ...tooltipProps
+      }: {
+        onMouseEnter: EventHandler<MouseEvent>
+        onFocus: EventHandler<FocusEvent>
```
matthprost, line 291: can't we use `MouseEvent<HTMLButtonElement>` for `onMouseEnter` instead ?
and `FocusEvent<HTMLInputElement>` for `onFocus` ?

matthprost (changes requested):

src/components/Typography/index.tsx
chambo-e: You're absolutely right, done :)

matthprost (approved):

vincentaudebert (approved):

PR #626, opened 2021-08-10
feat: migrate Slider, Sphere, StealthCopiable, Switch and SwitchButton to typescript

chambo-e:
## Type

- Refactor


src/components/Slider/index.tsx
```diff
@@ -76,36 +82,41 @@ const StyledBorderWrapper = styled(Box)`
...
   const handleScrollLeft = () => {
-    if (scrollRef.current) {
-      if (scrollRef.current.scrollTo) {
-        intervalLeft = setInterval(() => {
-          scrollRef.current.scrollTo(scrollRef.current.scrollLeft + 25, 0)
-        }, 30)
-      }
+    if (scrollRef.current?.scrollTo) {
+      intervalLeft = setInterval(() => {
+        scrollRef.current?.scrollTo(scrollRef.current?.scrollLeft + 25, 0)
```
matthprost, line 113:
```suggestion
    if (scrollRef.current?.scrollTo) {
      intervalLeft = setInterval(() => {
        scrollRef.current.scrollTo(scrollRef.current?.scrollLeft + 25, 0)
```

We already check if the value exists

src/components/Sphere/index.tsx
```diff
@@ -1,12 +1,23 @@
...
 import Box from '../Box'
 
-const bordersStyles = ({ size, bgColors, theme }) => {
+const bordersStyles = ({
+  size,
+  bgColors,
+  theme,
+}: {
+  size: number
+  bgColors: string[]
```
matthprost, line 14:
```suggestion
  bgColors: (string | Color)[]
```

src/components/Switch/index.tsx
```diff
@@ -245,40 +306,18 @@ Switch.propTypes = {
...
   ]),
   name: PropTypes.string.isRequired,
-  /**
-   * Text or node shown in switch button when its state if off
-   */
   offLabel: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
   onChange: PropTypes.func.isRequired,
-  /**
-   * Text or node shown in switch button when its state if on
-   */
```
matthprost, line 263: Those text are supposed to be shown in Storybook, can you keep them and move them into type definitions ?

matthprost (changes requested):

src/components/Sphere/index.tsx
chambo-e: This would break the propTypes below, we have no way to specify that a prop need to be a Color for now :)

src/components/Slider/index.tsx
```diff
@@ -76,36 +82,41 @@ const StyledBorderWrapper = styled(Box)`
...
-  const scrollRef = useRef(null)
-  let intervalLeft
-  let intervalRight
+export const Item: FunctionComponent<{ as?: string | ElementType<unknown> }> =
+  ({ as, ...props }) => (
+    <StyledBorderWrapper as={as} {...props} draggable="true" />
+  )
+
+Item.propTypes = {
+  as: PropTypes.string,
```
matthprost, line 91: Seen with you, to be replaced by `ElementType`

src/components/Switch/index.tsx
```diff
@@ -250,7 +311,7 @@ Switch.propTypes = {
    */
   labeled: PropTypes.oneOfType([
     PropTypes.bool,
-    PropTypes.oneOf(['inside', 'left', 'right']),
+    PropTypes.oneOf(labelPositions),
   ]),
   name: PropTypes.string.isRequired,
   /**
```
matthprost, line 317: Can we also move comments to type definition ? So storybook display them

matthprost (changes requested):

src/components/Slider/index.tsx
chambo-e: Seen with you, we keep as is until we remove the propTypes 😄

matthprost (approved):

philibea (approved):

PR #641, opened 2021-08-16
chore: ensure consistents emotion css classNames

adriengibrat:
## Summary

Use snapshots classname generation based on style content to reduce false positive snaphosts test failure.

## Type
- Refactor

#### What is expected?

No change at all, only tests snapshot update.


chambo-e (approved):

philibea (approved):

adriengibrat:
update DateInput test snapshot with locale set to french -> CI test seems to be run with french locale.

& add an explicit `test:update` script

PR #649, opened 2021-08-18
feat: migrate some components to typescript again

DorianMaliszewski:
## Type
- Migration

#### What is expected?

Migration of PasswordStrengthMeter, PasswordChecker, GlobalStyle, Description, CreationProgress, Counter

1. Migrate to TS components

2. Migrate to TS tests


src/components/Counter/index.tsx
```diff
@@ -16,8 +16,22 @@ const styles = {
...
+   * Value to which counter will go. Can be negative or positive number.
+   */
+  end: number
+  /**
+   * Triggered function when counter animation is finished.
+   */
+  onEnd?: () => unknown
+}
+
+const Counter: FunctionComponent<CounterProps> = ({
```
vincentaudebert, line 30: Can it be a VoidFunctionComponent as there are no children?

src/components/Counter/__tests__/index.test.tsx
```diff
@@ -9,6 +9,6 @@ describe('Counter', () => {
     ${100}
     ${10000}
   `('renders correctly end=$end', ({ end }) =>
-    shouldMatchEmotionSnapshot(<Counter end={end} />),
+    shouldMatchEmotionSnapshot(<Counter end={end as number} />),
```
adriengibrat, line 12: `Counter` end prop only accept number, we may test like:
```
  test.each([0,100,1000])('renders correctly end=%p', (end) =>
    shouldMatchEmotionSnapshot(<Counter end={end} />),
  )
```

src/components/CreationProgress/index.tsx
```diff
@@ -43,7 +50,9 @@ const sizes = {
     stepTextMargin: 15,
     text: 10,
   },
-}
+} as const
```
adriengibrat, line 53: Thanks, I've learn something
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions
but I think it's not usefull here (we don't need the sizes values in types) ?

```suggestion
} 
```

src/components/Description/index.tsx
```diff
@@ -0,0 +1,104 @@
+import { css } from '@emotion/react'
+import styled from '@emotion/styled'
+import PropTypes from 'prop-types'
+import React, { FunctionComponent } from 'react'
+import Box, { BoxProps } from '../Box'
+
+const StyledDt = styled(Box)``
+const StyledDd = styled(Box)``
```
adriengibrat, line 8: 🤯 Why not something like:
```suggestion
const StyledDt = styled(Box.withComponent('dt'))`
    font-weight: 500;
    color: ${({ theme }) => theme.colors.gray950};
    &:after {
      content: ':';
    }
  `
const StyledDd = styled(Box.withComponent('dd'))`
    color: ${({ theme }) => theme.colors.gray700};
    margin: 0;
    margin-top: ${({ theme }) => theme.space['1']};
`
```

adriengibrat (approved):

src/components/Counter/index.tsx
```diff
@@ -16,8 +16,22 @@ const styles = {
...
-  const start = useRef(0)
+type CounterProps = {
+  /**
+   * Value to which counter will go. Can be negative or positive number.
+   */
+  end: number
+  /**
+   * Triggered function when counter animation is finished.
+   */
+  onEnd?: () => unknown
```
chambo-e, line 27:
```suggestion
  onEnd?: () => void
```

We won't care about the return no ?

src/components/Description/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+        &::selection {
+          color: ${theme.colors.gray200};
+          background: ${theme.colors.primary};
+        }
+      }
+    `}
+`
+
+export type DescriptionProps = {
```
chambo-e, line 65: Why export them ?

src/components/PasswordStrengthMeter/index.tsx
```diff
@@ -28,25 +34,56 @@ const StyledMeter = styled.div`
...
-  const [width, setWidth] = useState(0)
+  const [score, setScore] = useState<number>(0)
+  const [backgroundColor, setBackgroundColor] = useState<string>(
+    strength[0].color,
+  )
+  const [width, setWidth] = useState<number | string>(0)
 
   const getScore = useCallback(
-    passwordToTest => estimate(passwordToTest || '', userInputs).score || 0,
+    passwordToTest => estimate?.(passwordToTest || '', userInputs).score || 0,
```
chambo-e, line 82: estimate will never be undefined no ?

src/components/CreationProgress/index.tsx
```diff
@@ -161,15 +176,27 @@ const StyledContainer = styled(Box)`
   }
 `
 
-const CreationProgress = ({
+type CreationProgressProps = {
+  animated?: boolean
+  isStepsNumber?: boolean
+  selected?: number
+  size?: Size
+  children: ReactNodeArray
```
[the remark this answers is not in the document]
DorianMaliszewski: Clearly its NodeArray because the user need to put more than one child to make the component working @chambo-e

chambo-e (approved):

PR #654, opened 2021-08-20
feat: correct es module tree shaking by preserving modules on output

chambo-e:
## Summary

Turns out webpack does not tree shake bundled esm modules (https://github.com/webpack/webpack/issues/9337#issuecomment-524966201)

## Type

- Enhancement

 Before | After
  :-:   | -:
 Screenshot 2021-08-21 at 00 12 54 | Screenshot 2021-08-21 at 00 11 12

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,10 +65,11 @@
   },
   "size-limit": [
     {
-      "path": "dist/index.js",
-      "import": "{ Box }",
+      "path": "dist/**/*.js",
+      "limit": "220 kB",
       "webpack": false,
-      "limit": "200 kB"
+      "brotli": true,
+      "running": false
     }
   ],
   "jest": {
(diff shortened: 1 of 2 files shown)


DorianMaliszewski (commented):
Check the CI, it seems size-limit doesn't work anymore and rollup still output in dist folders instead of dist/src/

package.json
```diff
@@ -25,9 +25,9 @@
     "linux"
   ],
   "sideEffects": false,
-  "main": "dist/index.js",
-  "module": "dist/index.js",
-  "jsnext:main": "dist/index.js",
+  "main": "dist/src/index.js",
```
philibea, line 28: don't forget to change this path into size-limit 
`"size-limit": [
    {
"path": "dist/index.js",`

philibea (approved):

DorianMaliszewski (approved):

PR #662, opened 2021-08-24
fix(typings): correct generated types + some tests

chambo-e:
## Summary

Global types were not being injected in built declaration file


DorianMaliszewski:
.ts is better than .d.ts for type declaration ? @chambo-e

DorianMaliszewski (approved):

philibea (approved):

PR #674, opened 2021-08-27
refactor(tags): migration typescript

philibea:
## Type

- Refactor Tags Component to Typescript


src/components/Tags/__tests__/index.tsx
```diff
@@ -114,8 +116,10 @@ describe('Tags', () => {
       {
         transform: async () => {
           const input = document.getElementById('test')
```
DorianMaliszewski, line 118: We need to drop all document.X and use jest selectors. Our tests mustn't be conditonnals @philibea

src/components/Tags/index.tsx
```diff
@@ -0,0 +1,282 @@
...
+      setStatus({ [tagIndex]: STATUS.IDLE })
+    } catch (e) {
+      onChangeError?.(e)
+      setTags(tagsState)
+    }
+  }
+
+  const handleInputKeydown = e => {
+    // 32 = Space | 13 = Enter | 8 = backspace
+    if (e.keyCode === 32 || e.keyCode === 13) {
```
DorianMaliszewski, line 189: keyCode is deprecated, is it possible to change it in this PR ?

src/components/Tags/index.tsx
```diff
@@ -0,0 +1,282 @@
...
+      addTag()
+      e.preventDefault()
+    }
+    if (
+      e.keyCode === 8 &&
```
DorianMaliszewski, line 194: Same

DorianMaliszewski (changes requested):

src/components/Tags/index.tsx
philibea: i made something for this component, but i believe we should probably use https://github.com/timoxley/keycode

src/components/Tags/__tests__/index.tsx
philibea: Solved

DorianMaliszewski (approved):

src/components/Tags/index.tsx
```diff
@@ -0,0 +1,296 @@
...
+      : tagsState
+    setInput('')
+    setTags(newTags)
+    if (newTags.length !== tagsState.length) {
+      setStatus({ [newTags[newTags.length - 1].index]: STATUS.LOADING })
+    }
+    try {
+      dispatchOnChange(newTags)
+      setStatus({ [newTags[newTags.length - 1].index]: STATUS.IDLE })
+    } catch (error) {
```
chambo-e, line 174: ERROR

chambo-e (approved):

PR #675, opened 2021-08-27
feat: navigationstepper component

DorianMaliszewski:
## Type

- Feature

#### What is expected?

New Navigation Stepper (or Progress) component

1. The component

2. Stories

3. Tests


chambo-e (approved):

src/components/NavigationStepper/__tests__/index.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+    shouldMatchEmotionSnapshot(
+      <NavigationStepper condensed step={2}>
+        <NavigationStepper.Step isLoading>First</NavigationStepper.Step>
+        <NavigationStepper.Step>Second</NavigationStepper.Step>
+        <NavigationStepper.Step>Third</NavigationStepper.Step>
+      </NavigationStepper>,
+    ))
+
+  test('should render correctly with default condensed', async () => {
+    Object.defineProperty(window, 'matchMedia', {
```
chambo-e, line 57: Already done in beforeAll ?

src/utils/responsive/useMedia.ts
```diff
@@ -0,0 +1,43 @@
+import { useCallback, useEffect, useState } from 'react'
```
chambo-e, line 1: https://github.com/streamich/use-media

https://github.com/jepser/use-match-media

https://github.com/streamich/react-use/blob/master/docs/useMedia.md


Plenty of options ?

src/components/NavigationStepper/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+  step = 1,
+}) => {
+  const theme = useTheme()
+  const isMobile = useMedia(
+    [`(max-width: ${theme.screens.small}px)`],
+    [true],
+    false,
+  )
+
+  console.log('IsMobile', isMobile)
```
chambo-e, line 161: 💩

src/components/NavigationStepper/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+            <StyledStepContent aria-hidden={!isCurrent && computedCondensed}>
+              {child}
+            </StyledStepContent>
+          </StyledStep>
+        )
+      }
+
+      return null
+    },
+  ) as ReactNode[]
```
chambo-e, line 206: If you filter out null values would you still need to cast as ReactNode[] ?

chambo-e (changes requested):

src/components/NavigationStepper/__tests__/index.test.tsx
DorianMaliszewski: But I switch from matches from false to true to emulate the default condensed at true for small screens

src/utils/responsive/useMedia.ts
DorianMaliszewski: I took the one from react-hooks, do you prefer to add a dependency ?

src/components/NavigationStepper/index.tsx
DorianMaliszewski: No it is `steps.length` the problem cause React.Children.map can return undefined

src/utils/responsive/useMedia.ts
chambo-e: yes, less maintenance for us :)

src/components/NavigationStepper/index.tsx
chambo-e: Alright :)

chambo-e (approved):

philibea (approved):

PR #704, opened 2021-09-06
fix: avoid to declare emotion Theme typings

adriengibrat:
## Type

- Enhancement

#### What is expected?

Allows projects to define custom Theme typings

1. removed extended emotion Theme typings from exported typings
2. forces to localy extends emotion Theme
3. exports SCWUITheme for convenience

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,6 +15,8 @@
     "**/*.tsx",
     "**/*.js",
     "**/*.jsx",
+    "emotion.d.ts",
+    "global.d.ts"
   ],
   "exclude": [
     "node_modules",
--- a/README.md
+++ b/README.md
@@ -35,6 +35,33 @@ const App = () => (
 );
 ```
 
+N.B. To allow typescript theme typings with `@emotion/styled` components,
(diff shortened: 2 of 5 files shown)


global.d.ts
```diff
@@ -1,4 +1,5 @@
 /// <reference types="@emotion/react/types/css-prop" />
+import type { SCWUITheme } from './src/theme'
```
chambo-e, line 2:
```suggestion
import type { SCWUITheme } from './src'
```

tsconfig.json
```diff
@@ -15,6 +15,8 @@
     "**/*.tsx",
     "**/*.js",
     "**/*.jsx",
+    "emotion.d.ts",
```
chambo-e, line 18: These should already be included with `"**/*.ts"` 🤔

adriengibrat: I know. For those specific types, I just prefer to be "extra" explicit.

DorianMaliszewski (approved):

chambo-e (approved):

matthprost (approved):

PR #705, opened 2021-09-06
fix(modal): onBeforeClose & ExtendReminder CustomLink

matthprost:
## Type

- Refactor

#### What is expected?

**Modals**: added `onBeforeClose` prop that will be triggered before the modal is closed. This is useful when using disclosure and not onClose prop.

**ExtendedReminder**: added `CustomLink` prop that can take a node value. This is useful when wanting to put a modal with disclosure inside of ExtendedReminder without using default link and onClick prop.

### DIFF
--- a/src/components/ExtendedReminder/index.tsx
+++ b/src/components/ExtendedReminder/index.tsx
@@ -1,7 +1,7 @@
 import { Theme, useTheme } from '@emotion/react'
 import styled from '@emotion/styled'
 import PropTypes from 'prop-types'
-import React, { FunctionComponent, MouseEventHandler } from 'react'
+import React, { FunctionComponent, MouseEventHandler, ReactNode } from 'react'
 import Badge from '../Badge'
 import Box from '../Box'
 import Button from '../Button'
@@ -94,6 +94,10 @@ type Props = {
    */
   to?: string
   variant?: Variants
+  /**
+   * Custom link component
+   */
+  CustomLink?: ReactNode
 }
 
 const ExtendedReminder: FunctionComponent<Props> = ({
@@ -105,6 +109,7 @@ const ExtendedReminder: FunctionComponent<Props> = ({
   title,
   to,
   variant = 'info',
+  CustomLink,
   ...props
 }) => {
   const theme = useTheme()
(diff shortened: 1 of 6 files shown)


src/components/Modal/index.tsx
```diff
@@ -303,13 +305,25 @@ const Modal: FunctionComponent<ModalProps> = ({
           hideOnEsc={hideOnEsc}
           preventBodyScroll={preventBodyScroll}
           {...dialog}
-          hide={onClose || dialog.toggle}
+          hide={
+            onClose ||
+            (() => {
```
chambo-e, line 310: might be worth to useCallback here has we have the same function just below :)

src/components/Modal/__tests__/index.tsx
```diff
@@ -46,6 +46,19 @@ describe('Modal', () => {
...
       </Modal>,
     ))
+
+  test(`renders with disclosure and onBeforeClose`, () =>
+    shouldMatchEmotionSnapshotWithPortal(
+      <Modal
+        ariaLabel="modal-test"
+        baseId="modal-test"
+        disclosure={() => <button type="button">Test</button>}
+        onBeforeClose={() => console.log('onBeforeClose triggered')}
```
DorianMaliszewski, line 56: mock a jest function and count how many time the function is involved ?

DorianMaliszewski (changes requested):

src/components/ExtendedReminder/index.tsx
```diff
@@ -129,23 +134,25 @@ const ExtendedReminder: FunctionComponent<Props> = ({
...
+          <StyledButtonLink
+            variant="link"
+            to={to}
+            onClick={onClick}
+            size="xsmall"
+            icon="east"
+          >
+            {linkText}
+          </StyledButtonLink>
+        ))}
```
DorianMaliszewski, line 148:
```suggestion
        (linkText ? (
          <StyledButtonLink
            variant="link"
            to={to}
            onClick={onClick}
            size="xsmall"
            icon="east"
          >
            {linkText}
          </StyledButtonLink>
        ) : null)}
```
https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx/ ?

src/components/Modal/__tests__/index.tsx
matthprost: Good idea 🔝

src/components/ExtendedReminder/index.tsx
matthprost: @DorianMaliszewski `linkText` is undefined if not set, so in case its value is undefined nothing will render I believe

DorianMaliszewski: As you want but I think its a good practices to use ternary instead of && to avoid some mistakes 🤔

DorianMaliszewski (approved):

chambo-e (approved):

PR #709, opened 2021-09-08
refactor: convert RichSelect to typescript

adriengibrat:
## Type

- Migration

#### What is expected?

Migrate RichSelect component to typescript.

1. Migrate component WIP

2. Migrate test TODO


react-select.d.ts
```diff
@@ -0,0 +1,15 @@
+import { ForwardedRef, HTMLAttributes } from "react"
+import { Props } from "react-select"
+
+declare module "react-select" {
```
chambo-e, line 4: Could we contribute on the @types/react-select instead ?

src/components/RichSelect/index.tsx
```diff
@@ -380,14 +421,14 @@ const ValueContainer = ({
...
-  selectProps: {},
-}
+// ValueContainer.defaultProps = {
+//   children: null,
+//   hasValue: false,
+//   isDisabled: false,
+//   isMulti: false,
+//   noTopLabel: false,
+//   selectProps: {},
+// }
```
matthprost, line 431: You can set default values to `ValueContainer` function declaration and remove `defaultProps` here

src/components/RichSelect/index.tsx
```diff
@@ -433,17 +472,16 @@ Input.defaultProps = {
   value: '',
 }
 
-const Option = ({ selectProps, value, label, ...props }) => (
+const Option = ({ selectProps, value, label, ...props }: OptionProps<SelectOption, boolean> & SelectOption) => (
   <div
-    data-testid={`option-${selectProps.name}-${
+    data-testid={`option-${selectProps.name || ''}-${
```
matthprost, line 477: we have:
```js
selectProps: {
      name = ''
}
```

So name by default value is already `''`
```suggestion
    data-testid={`option-${selectProps.name}-${
```

matthprost (changes requested):

vincentaudebert:
Looks good to me after manu and matthias feedbacks. Please ping me when it's fixed I will final check.

src/components/RichSelect/index.tsx
adriengibrat: react-select `OptionProps` typing declares `name` as optional, and I prefer to use an empty string as fallback here rather than trying to override types.

react-select.d.ts
adriengibrat: major v5 is coming with a [large typescript rewrite](https://github.com/JedWatson/react-select/issues/4559)...
image
I think the patch will not be necessary and some parts of the typings will need to be adapted.

adriengibrat:
@chambo-e I still have 2 main issues to resolve:
- inconsistencies between typescript & propTypes about some properties type in `selectProps` & `innerProps` -> I'm clueless at this point
- the `-0.08%` in codecov due to an attempt to type `RichSelect.Option` with a placeholder function (without coverage)

vincentaudebert (approved):

src/components/RichSelect/index.tsx
```diff
@@ -275,21 +363,13 @@ const SelectContainer = props => {
...
 SelectContainer.propTypes = {
   children: PropTypes.node.isRequired,
   className: PropTypes.string,
-  getStyles: PropTypes.func,
-  innerProps: PropTypes.shape({}),
-  isDisabled: PropTypes.bool,
+  getStyles: PropTypes.func.isRequired,
+  innerProps: PropTypes.shape({}).isRequired,
+  isDisabled: PropTypes.bool.isRequired,
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
matthprost, line 372: No way to remove this eslint rule ? Can't it be typed as `unknown` ?

src/components/RichSelect/index.tsx
```diff
@@ -380,96 +469,89 @@ const ValueContainer = ({
...
+    }
   />
 )
+
 Input.propTypes = {
-  hasValue: PropTypes.bool,
-  isMulti: PropTypes.bool,
-  selectProps: SelectContainer.propTypes.selectProps,
+  hasValue: PropTypes.bool.isRequired,
+  isMulti: PropTypes.bool.isRequired,
```
matthprost, line 518: Why is it required if it has a default value ?

```js
isMulti = false,
hasValue = false,
```

src/components/RichSelect/index.tsx
```diff
@@ -686,10 +753,10 @@ RichSelectWithRef.propTypes = {
    */
   customStyle: PropTypes.func,
   disabled: PropTypes.bool,
-  error: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
+  error: PropTypes.string,
   inputId: PropTypes.string,
   isClearable: PropTypes.bool,
-  isMulti: PropTypes.bool,
+  isMulti: PropTypes.bool.isRequired,
```
matthprost, line 759: Same here

matthprost (changes requested):

matthprost:
Screenshot 2021-10-05 at 16 09 52

Regression on RichSelect `noLabel` prop is not working anymore

matthprost (changes requested):
Regression on RichSelect noLabel prop is not working anymore

Screenshot 2021-10-05 at 16 09 52

Which lead to regression of unit input having a label while it shouldn't have
Screenshot 2021-10-05 at 16 11 41

src/components/RichSelect/index.tsx
adriengibrat: I removed default values, it's required by react-select typing of components.Input `InputProps`.

src/components/RichSelect/index.tsx
adriengibrat: :+1: here I removed the isRequired ;)

src/components/RichSelect/index.tsx
adriengibrat: I wish it could...

adriengibrat: I tried:
- tweaking types
- type casting `as Validator<SelectProps>`
- type casting `as unknown`

nothing worked, typescript always complains about proptypes compatibility...

adriengibrat:
@matthprost , the regression on RichSelect noTopLabel prop is fixed:

react-select v5 no longer pass all select props to sub components... I had to handle `noTopLabel` with custom styles only.

DorianMaliszewski (approved):

src/components/RichSelect/__tests__/index.tsx
```diff
@@ -41,7 +42,7 @@ describe('RichSelect', () => {
         inputId="test"
         labelId="test-label"
         name="uncontrolled"
-        customStyles={customStyles}
+        customStyle={() => customStyles}
```
DorianMaliszewski, line 45: Is this not breaking ?? 🤔

adriengibrat: nope, it was broken.

PR #712, opened 2021-09-09
fix: radioborderedbox add xstyledprops

matthprost:
Just added XStyledProps to radioBorderedBox.

Edit: also changed `label` prop to accept not only string but any other complex children.


chambo-e (approved):

matthprost:
@chambo-e I did a modification, could you please review again ?

chambo-e (approved):

DorianMaliszewski (approved):

PR #726, opened 2021-09-16
feat(textbox): reduce padding & add `inputProps`

QuiiBz:
## Type

- Feature

#### What is expected?

The padding-right has been reduced to `required` TextBox, and this PR adds an `inputProps` prop to TextBox which are forwarded to the input element.

Page | Before | After
:-   |  :-:   | -:
-- | Screenshot 2021-09-16 at 15 14 34 | Screenshot 2021-09-16 at 15 15 00

### DIFF
--- a/src/components/TextBox/index.tsx
+++ b/src/components/TextBox/index.tsx
@@ -480,13 +480,10 @@ const TextBox = forwardRef<
     const hasRightElement =
       valid || required || isPassToggleable || random || unit
 
-    const getRightElementPadding = () => {
-      if (required && hasRightElement) {
-        return 22
-      }
-
-      return undefined
-    }
+    // Set the right padding to 22px when the TextBox is required. 22px allows
+    // keeping the required star icon centered, since it is smaller than valid or
+    // unit icons which have a right padding of 32px.
+    const rightElementPadding = required ? 22 : undefined
 
     const getType = () => {
       if (isPassToggleable) {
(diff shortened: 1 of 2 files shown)


src/components/TextBox/index.tsx
```diff
@@ -471,6 +480,14 @@ const TextBox = forwardRef<
     const hasRightElement =
       valid || required || isPassToggleable || random || unit
 
+    const getRightElementPadding = () => {
+      if (required && hasRightElement) {
+        return 22
```
chambo-e, line 485: What is this number ? 🤔

src/components/TextBox/index.tsx
```diff
@@ -618,6 +637,7 @@ TextBox.propTypes = {
   generated: PropTypes.bool,
   height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
   id: PropTypes.string,
+  inputProps: PropTypes.objectOf(PropTypes.any),
```
chambo-e, line 640: In which case are we going to use this props ?

chambo-e (changes requested):

src/components/TextBox/index.tsx
QuiiBz: It allows to have the required icon centered:

Screenshot 2021-09-16 at 15 15 00

src/components/TextBox/index.tsx
QuiiBz: That was requested in CNS-3912, to reduce the input height.

src/components/TextBox/index.tsx
chambo-e: Possible to put it in a const describing it pls ? :)

chambo-e (approved):

DorianMaliszewski (approved):

PR #746, opened 2021-09-27
fix(flexbox): fixed proptypes spreading to accept xstyled props

matthprost:
## Type

- Bug

`Flexbox` was not applying xstyled props such as `m={}`, `p={}`, etc. like `Box` component does.

This is because `Flexbox` style is a div so I changed it to a `Box`.


chambo-e:
No go, this was on purpose :)
Spacing should not be managed through xstyled props anymore

PR #756, opened 2021-09-29
fix(menu): allow to not render inside menu when it s not necessary

philibea:
## Type

- Enhancement

#### What is expected?

Content inside menu will not be present on the DOM if the menu is not open.

### DIFF
--- a/src/components/Menu/__tests__/index.tsx
+++ b/src/components/Menu/__tests__/index.tsx
@@ -7,7 +7,6 @@ describe('Menu', () => {
   test('renders with visible=false', () =>
     shouldMatchEmotionSnapshotWithPortal(
       <Menu
-        visible={false}
         baseId="menu"
         disclosure={() => <button type="button">Menu is visible</button>}
       >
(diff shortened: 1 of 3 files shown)


chambo-e (approved):

src/components/Menu/__tests__/index.tsx
```diff
@@ -4,9 +4,20 @@ import shouldMatchEmotionSnapshot from '../../../helpers/shouldMatchEmotionSnaps
 import shouldMatchEmotionSnapshotWithPortal from '../../../helpers/shouldMatchEmotionSnapshotWithPortal'
 
 describe('Menu', () => {
+  test('renders with visible=false', () =>
+    shouldMatchEmotionSnapshotWithPortal(
+      <Menu
+        visible={false}
+        baseId="menu"
+        disclosure={() => <button type="button">Menu is visible</button>}
+      >
```
matthprost, line 13:
```suggestion
      <Menu
        baseId="menu"
        disclosure={() => <button type="button">Menu is visible</button>}
      >
```
`false` is its default value so no need to precise it

matthprost (changes requested):

DorianMaliszewski (approved):
@philibea I'm ok after fixing @matthprost feedback.

PR #760, opened 2021-10-01
Ci/a11y tests

matthprost:
## Type

- Bug
- Feature
- Enhancement
- Documentation
- Migration
- Refactor

#### What is expected?

(Describe what you did)

1.

2.

3.

Page | Before | After
:-   |  :-:   | -:
url | screenshot | screenshot
url | screenshot | screenshot


.github/workflows/ci.yml
```diff
@@ -37,6 +37,24 @@ jobs:
...
+        with:
+          fetch-depth: "0"
+      - name: Use Node.js
+        uses: actions/setup-node@v2.4.1
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'yarn'
+      - run: yarn
+      - run: yarn run test:a11y
+      - uses: codecov/codecov-action@v2.1.0
```
DorianMaliszewski, line 57: Is codecov that does the report ?

.storybook/preview.js
```diff
@@ -19,11 +19,11 @@ const ENV_PARAMETERS = {
       },
     },
     controls: {
-      disable: true,
+      disable: false,
```
DorianMaliszewski, line 22: Is this the default value ?

package.json
```diff
@@ -44,7 +44,8 @@
     "test:coverage": "yarn run test --coverage",
     "test:watch": "yarn run test --watch",
     "test:update": "yarn run test --updateSnapshot",
-    "test": "LC_ALL=en_US.UTF-8 jest",
+    "test": "LC_ALL=en_US.UTF-8 jest ./src/components",
```
DorianMaliszewski, line 47: test should test all .{test, spec}.{tsx, ts}  files no ? helpers/ ....

src/__tests__/a11y.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+searchFileFromDir('src/components', '.stories.tsx')
+
+expect.extend(toHaveNoViolations)
+
+describe('A11y', () => {
+  afterEach(() => {
+    cleanup()
+  })
+
+  const array: Promise<{
```
DorianMaliszewski, line 37: array is not explicit 😢

src/components/Button/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,224 @@
...
+    storyDescription:
+      'Add `ActivityIndicator` by setting progress property to `left` or `right`.',
+  },
+}
+Progress.decorators = [
+  () => (
+    <>
+      <Button progress="left" mr={2}>
+        left progress
+      </Button>
```
DorianMaliszewski, line 71:
```suggestion
export const Progress = Template.bind({})
Progress.parameters = {
args:{
progress: 'left'
},
  docs: {
    storyDescription:
      'Add `ActivityIndicator` by setting progress property to `left` or `right`.',
  },
}
```

Not enough ? 🤔

package.json
matthprost: I changed it to test `./src/components ./src/helpers ./src/utils`

src/components/Button/__stories__/index.stories.tsx
matthprost: Using `args` works well when you wanna show only one, once you wanna show multiple of them you need to use decorators

package.json
```diff
@@ -41,10 +41,12 @@
     "lint:fix": "yarn run lint --fix",
     "lint": "eslint --cache --ext js,ts,tsx,mdx .",
     "prebuild": "shx rm -rf dist",
-    "test:coverage": "yarn run test --coverage",
+    "test:unit:coverage": "yarn run test:unit --coverage",
     "test:watch": "yarn run test --watch",
     "test:update": "yarn run test --updateSnapshot",
     "test": "LC_ALL=en_US.UTF-8 jest",
+    "test:unit": "yarn test ./src/components ./src/helpers ./src/utils",
```
chambo-e, line 48: Why specify folders ? 🤔

matthprost: To avoid running `ay11.test.tsx` in unit tests

package.json
chambo-e: You can use [testPathIgnorePatterns](https://jestjs.io/docs/en/configuration#testpathignorepatterns-array-string) to only exclude this file
This would be more future proof :)

chambo-e: And I don't really understand, we can't run the a11y tests with the unit tests ?

matthprost: Thanks I was searching for such option :)

We don't run with unit testing for the simple reason that our accessibility is not good at all yet and will make test fails. To avoid confusion and understand better CI I split a11y and unit testing.

Also a11y tests can only be run on new stories format (using TSX).

.storybook/main.js
```diff
@@ -6,9 +6,12 @@ module.exports = {
   },
   typescript: {
     // also valid 'react-docgen-typescript' | false
-    reactDocgen: 'react-docgen',
+    reactDocgen: 'react-docgen-typescript',
```
vincentaudebert, line 9: MMmmh on est surs de ca? de souvenir c'etait pour les args table

matthprost: We are migrating all files to typescript which makes args table non functional (checkout PR on rich select). It is required to go trough react-docgen-typescript

matthprost:
> looks like accessibility test fails? @matthprost

The goal of this MR is not to fix all accessibility issues but to define an accessibility testing method to then find and fix them all more easily :)

vincentaudebert:
Makes sense!

vincentaudebert (approved):

PR #766, opened 2021-10-04
feat:  Charts components

adriengibrat:
## Type

- Feature

#### What is expected?

(Describe what you did)

1.

2.

3.

LineChart
image


src/components/LineChart/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+type Formater = ValueFormat<DatumValue>
+
+type Shape<K extends string, T> = {
+  [P in K]?: T
+}
+
+type LineChartProps = {
+  height?: string | number
+  margin?: Shape<'bottom' | 'left' | 'right' | 'top', number>
```
DorianMaliszewski, line 18: bottom, left, right, top possible to export it into a Position type ?

src/components/LineChart/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const chartTheme = {
+    axis: {
+      ticks: {
+        line: {
+          stroke: theme.colors.gray300,
+          strokeWidth: 1,
+        },
+      },
+    },
+    fontFamily: 'Asap',
```
DorianMaliszewski, line 59: We gonna change our font Family so is it posisble to get font family front theme instead ?

chambo-e:
I think the end goal was to merge all of our Chart under one main component with the same interface
cc @vincentaudebert because I can't remember exactly :)

```jsx
<LineChart />
<PieChart />
<BarChart />
// =>
<Chart kind="line" />
<Chart kind="pie" />
<Chart kind="bar" />
```

src/components/LineChart/index.tsx
adriengibrat: Changed to Box (imported from `@nivo/core`)

adriengibrat:
> I think the end goal was to merge all of our Chart under one main component with the same interface cc @vincentaudebert because I can't remember exactly :)
> 
> ```js
> <LineChart />
> <PieChart />
> <BarChart />
> // =>
> <Chart kind="line" />
> <Chart kind="pie" />
> <Chart kind="bar" />
> ```

They have distinct data format, so I don't think it's a good idea.

src/components/LineChart/index.tsx
adriengibrat: `theme.fonts.sansSerif`

vincentaudebert:
> I think the end goal was to merge all of our Chart under one main component with the same interface cc @vincentaudebert because I can't remember exactly :)
> 
> ```js
> <LineChart />
> <PieChart />
> <BarChart />
> // =>
> <Chart kind="line" />
> <Chart kind="pie" />
> <Chart kind="bar" />
> ```

@chambo-e This is more a less containers/Chart in shire but here I reckon we should leave it as is. With distinct components. As Adrien said, data are totally different from one to another.

vincentaudebert (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@scaleway/ui",
-  "version": "0.112.4",
```
matthprost, line 3: Is it a mistake ?

matthprost (changes requested):

package.json
adriengibrat: yes,  wrong rebase resolution...

adriengibrat: fixed

PR #877, opened 2021-11-04
refactor: migrate Reminder, RichSelect, ScrollView stories to tsx

QuiiBz:
## Type

- Refactor

Migrate Reminder, RichSelect, ScrollView stories to tsx


src/components/ScrollView/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+        component:
+          'A component that will add a scroll bar on the left when text overpass box height.',
+      },
+    },
+  },
+  title: 'Components/ScrollView',
+} as Meta
+
+const Template: Story<ScrollViewProps> = args => (
+  <ScrollView height="300px" {...args}>
```
chambo-e, line 19: As ScrollView does not accept xstyled props anymore it seems that this example is not working

Maybe ?
```suggestion
  <ScrollView style={{ height: "300px" }} {...args}>
```

chambo-e (changes requested):

src/components/RichSelect/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+${Object.keys(animations)
+  .map(animation => `\`${animation}\``)
+  .join(' ')}`,
+  },
+}
+Animated.decorators = [
+  () => (
+    <Box mx={8} my={2}>
+      <ControlValue value="">
```
chambo-e, line 188:
```suggestion
      <ControlValue<string> value="">
```

To avoid eslint-disable below ?

chambo-e (approved):

src/components/Reminder/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+          'A styled badge with extra feature. In the `text` prop place some text between `[]` to make it bolder',
+      },
+    },
+  },
+  title: 'Components/Feedback/Reminder',
+} as Meta
+
+const Template: Story<ReminderProps> = args => (
+  <Reminder
+    display="flex"
```
vincentaudebert, line 20: Do we really need display flex and mb 1 in the template?

vincentaudebert: Just tried and display flex and mb 1 can be removed from here.

vincentaudebert (approved):

PR #886, opened 2021-11-08
refactor: migrate StealthCopiable, Stepper, Switch stories to tsx

QuiiBz:
## Type

- Refactor

Migrate StealthCopiable, Stepper, Switch stories to tsx


src/components/Switch/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,356 @@
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
```
chambo-e, line 1: Why ? 🤔

QuiiBz: This error appears in each story on `e.target.checked`, not sure why

chambo-e: You need to type your onChange :)

matthprost (approved):

chambo-e (approved):

PR #887, opened 2021-11-08
feat(global-style): remove GlobalStyle component

QuiiBz:
## Type

- Refactor

Remove `GlobalStyle` component and tests


chambo-e:
Also, PR should be a `feat` to bump the package and indicate a change :)

QuiiBz:
> Storybook seems to fail, I have an error on each page

Forgot to remove `<GlobalStyle />` from storybook, added back the css

.storybook/preview.js
```diff
@@ -55,6 +55,99 @@ const adjustedTheme = ancestorTheme => ({
   ),
 })
 
+const globalStyles = theme => css`
```
chambo-e, line 58: Can we cut it down to the bare minimum? Eg, not sure that hiding overflow to reenable it after is optimal :D

QuiiBz: It's currently the same CSS as before, what would you want to remove apart the duplicated `overflow`?

chambo-e: I was hoping everything expect fonts 😄 

If this is a prerequisite for this ui lib to look as intended we actually should remove it ...

matthprost (approved):

chambo-e (approved):

PR #897, opened 2021-11-10
refactor: migrate unitinput universallink verificationcode and volumesize and delete unselectable

vincentaudebert:
## Type

- Refactor

#### What is expected?

Migration of stories to TSX

(Describe what you did)

1. Migrate unitinput, universallink, verification and volumesize

2. Delete unselectable ⚠️ breaking change


src/components/UnitInput/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+    storyDescription: `UnitInput provide data as follows:
+
+\`\`\`js
+{
+    unit: [string],
+    value: [string]
+}
+\`\`\``,
+  },
+}
```
matthprost, line 33: Can't you do it this way instead:

```suggestion
Default.parameters = {
  docs: {
    storyDescription: "UnitInput provide data as follows:

```js
{
    unit: [string],
    value: [string]
}
```",
  },
}
```

?

matthprost (changes requested):

matthprost (changes requested):
Screenshot 2021-11-16 at 10 03 28

Unit input is a bit broken

src/components/UnitInput/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+} as Meta
+
+const Template: Story<UnitInputProps> = args => (
+  <UnitInput name="test" {...args} />
+)
+
+export const Default = Template.bind({})
+Default.parameters = {
+  docs: {
+    storyDescription: `UnitInput provide data as follows:
```
matthprost, line 24: Use story.description instead ?

src/components/UnitInput/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+    value: [string]
+}
+\`\`\``,
+  },
+}
+
+export const CustomOptions = Template.bind({})
+CustomOptions.parameters = {
+  docs: {
+    storyDescription: `You can customize options by passing an array of object. Be careful objects should have following structure:
```
matthprost, line 38: Here too

matthprost (changes requested):

vincentaudebert:
> Screenshot 2021-11-16 at 10 03 28
> 
> Unit input is a bit broken

Screenshot 2021-11-16 at 14 55 21

Already the case in production. It is not linked to this MR but I will investigate if it's an easy fix

src/components/UnitInput/__stories__/index.stories.tsx
vincentaudebert: `"` doesn't work with multiline/markdown

matthprost (approved):

DorianMaliszewski:
Capture d’écran 2021-11-17 à 14 30 50

Can we have props in the first Story because we doens't understand what is supposed to be 😅

DorianMaliszewski (changes requested):
See comment

DorianMaliszewski (approved):

PR #900, opened 2021-11-12
chore(storybook): migration BulletList, SwitchButton, TabGroup, Table, Tag, Tags  mdx to tsx (#895)

philibea:
## Type

- Documentation 

List Components:

- BulletList
- SwitchButton
-  TabGroup
- Table
-  Tag
-  Tags

#### What is expected?

Migration of mdx to Tsx


src/components/Switch/__stories__/index.stories.tsx
```diff
@@ -5,6 +5,53 @@ import { Box, Icon } from '../..'
...
+      control: { type: 'boolean' },
+      defaultValue: false,
+      description: 'labeled',
+      name: 'labeled',
+      table: {
+        defaultValue: { summary: false },
+        type: { summary: 'boolean' },
+      },
+      type: { name: 'boolean' },
+    },
```
matthprost, line 41: aren't controls automatically detected by storybook ?

matthprost (changes requested):

src/components/Switch/__stories__/index.stories.tsx
philibea: it's happen sometimes, on a lot of complexe typed component even. 

The only solution is to add argsType to spec and default args to add controls.

matthprost: That's terrible, it means each time we change proptypes we will need to change argsTypes of storybook too 😢 

There is no other way? I don't like much this solution

src/components/BulletList/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+  },
+  title: 'Components/Data Display/BulletList',
+} as Meta
+
+const Template: Story<BulletListProps> = args => <BulletList {...args} />
+
+export const Default = Template.bind({})
+
+Default.args = {
+  ...Template.args,
```
DorianMaliszewski, line 22: You don't need this because you dont put args in the Template 🤔

src/components/Switch/__stories__/index.stories.tsx
DorianMaliszewski: The only way we can get it is too export a proper type which not extend an imported type (here HTMLAttributes<HTMLElement>

src/components/Tag/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+Template.args = {
+  children: 'default',
+}
+
+export const Default = Template.bind({})
+
+Default.args = {
+  ...Template.args,
+}
```
DorianMaliszewski, line 38:
```suggestion
export default {
args: {
children: 'default'
},
  argTypes: {
    children: {
      defaultValue: 'This is a children',
      name: 'children',
      table: {
        defaultValue: { summary: 'This is a children' },
        type: { required: true, summary: 'string' },
      },
      type: { name: 'string', required: true },
    },
  },
  component: Tag,
  parameters: {
    docs: {
      description: {
        component: 'Compact element to display information inside a box.',
      },
    },
  },
  title: 'Components/Data Display/Tag',
} as Meta

const Template: Story<TagProps> = args => <Tag {...args} />

export const Default = Template.bind({})
```

Actually more cleaner https://storybook.js.org/docs/react/writing-stories/args#component-args

src/components/Tags/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+} as Meta
+
+const Template: Story<TagsProps> = args => <Tags {...args} />
+
+Template.args = {
+  name: 'default',
+  onChange: tags => console.log(tags),
+  tags: ['default'],
+  variant: 'base',
+}
```
DorianMaliszewski, line 84: Same here, put them in component args

src/components/Tags/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,90 @@
+import { Meta, Story } from '@storybook/react'
+import React from 'react'
+import Tags, { TagsProps } from '..'
+
+export default {
+  argTypes: {
```
DorianMaliszewski, line 6: Not using this should work with Tags and to add description in the argstable use TSDoc on the interface members

DorianMaliszewski (changes requested):

src/components/Tags/__stories__/index.stories.tsx
philibea: Without argTypes & args you have nothing.

With args, this will show this
Capture d’écran 2021-11-17 à 17 47 32

With ArgTypes 
Capture d’écran 2021-11-17 à 17 54 17

matthprost (approved):

DorianMaliszewski (approved):

PR #908, opened 2021-11-16
feat(rich-select): description and inline description

matthprost:
## Type

- Feature

#### What is expected?

To be able to add a description and inline description into rich select option without display it into input.

1. Added `description` and `inlineDescription` props into `RichSelect.Option`

Screenshot 2021-11-16 at 17 17 31
Screenshot 2021-11-16 at 17 17 42
Screenshot 2021-11-16 at 17 17 52

### DIFF
--- a/src/components/RichSelect/__stories__/index.stories.tsx
+++ b/src/components/RichSelect/__stories__/index.stories.tsx
@@ -9,7 +9,7 @@ export default {
   component: RichSelect,
   decorators: [
     Chidren => (
-      <Box mb={2}>
+      <Box mb={150}>
         <Chidren />
       </Box>
     ),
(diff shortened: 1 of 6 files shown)


QuiiBz (approved):
CI is failing due to snapshots

src/components/RichSelect/__stories__/index.stories.tsx
```diff
@@ -334,6 +334,31 @@ LoadingExample.decorators = [
...
+  <RichSelect name="basic">
+    <RichSelect.Option
+      value="a"
+      inlineDescription="This is an average solution"
+    >
+      Option A
+    </RichSelect.Option>
+    <RichSelect.Option
+      value="b"
+      description="This is the best solution, friend advice"
```
DorianMaliszewski, line 347: Should it be better to have `description={value: "blablabla" /* Or <div>blabla</div> */, inline: true}` ?
Or
`description="edededede" inlineDescription`

Like this description is a consistent prop which display a description 🤔

DorianMaliszewski (changes requested):

src/components/RichSelect/__stories__/index.stories.tsx
matthprost: @DorianMaliszewski That's a good one but in fact I tough that maybe one day we will have both inline and normal description to display at same time. If such case happen then my structure works better I think

DorianMaliszewski (approved):

QuiiBz (approved):

PR #923, opened 2021-11-19
refactor(a11y): alert

matthprost:
## Type

- Refactor

#### What is expected?

Make alert component more accessible.

1. Added `aria-label={type}` for div wrapping alert component with `role="region"`.

2. Added `aria-hidden="true"` for alert icon

3. Updated a11y testing system in order to be able to receive input parameter. ex: `yarn test:a11y src/components/Alert` will only run Alert a11y tests.

4. Updated readme

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "test:update": "yarn run test --updateSnapshot",
     "test": "LC_ALL=en_US.UTF-8 jest",
     "test:unit": "yarn test --testPathIgnorePatterns='src/__tests__/a11y.test.tsx'",
-    "test:a11y": "yarn test src/__tests__/a11y.test.tsx",
+    "test:a11y": "yarn test src/__tests__/a11y.test.tsx --testPathIgnorePatterns='src/components' ",
     "prepare": "husky install",
     "size": "yarn build && size-limit"
   },
--- a/package.json
+++ b/package.json
@@ -46,8 +46,8 @@
     "test:watch": "yarn run test --watch",
     "test:update": "yarn run test --updateSnapshot",
     "test": "LC_ALL=en_US.UTF-8 jest",
-    "test:unit": "yarn test --testPathIgnorePatterns='src/__tests__/a11y.test.tsx'",
-    "test:a11y": "yarn test src/__tests__/a11y.test.tsx --testPathIgnorePatterns='src/components' ",
+    "test:unit": "yarn test --config .jest/unit.config.ts",
+    "test:a11y": "yarn test --config .jest/a11y.config.ts",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "test:update": "yarn run test --updateSnapshot",
     "test": "LC_ALL=en_US.UTF-8 jest",
     "test:unit": "yarn test --testPathIgnorePatterns='src/__tests__/a11y.test.tsx'",
-    "test:a11y": "yarn test src/__tests__/a11y.test.tsx",
+    "test:a11y": "yarn test src/__tests__/a11y.test.tsx --testPathIgnorePatterns='src/components' ",
```
QuiiBz, line 50: Is `testPathIgnorePatterns` mandatory here?

matthprost: Yes I tried many different solutions and that's the only way I found. Let me explain:

if you do `yarn test:a11y src/components/Alert` without this parameter jest will launch `a11y.test.tsx` but also `src/components/Alert/__tests__/index.test.tsx`.

while we only want to give parameter to precise which a11y test should be run we also trigger normal unit tests.

src/components/Alert/index.tsx
```diff
@@ -126,8 +126,18 @@ const Alert: FunctionComponent<AlertProps> = ({
   type = 'warning',
   ...props
 }) => (
-  <StyledContainer type={type} variant={variant} {...props}>
-    <Icon name={icon || typesDefaultIcons[type]} size={iconSize} />
+  <StyledContainer
+    role="region"
+    aria-label={type}
+    type={type}
```
iManu, line 132: y'aurait moyen de localiser le type ?

iManu (commented):
ça passe bien sous VoiceOver

faudra voir quand ça surgit j'ai pas pu tester sous Storybook, sauf en mode "fullscreen" et ensuite rechargement de la page

package.json
chambo-e: Could we instead have two jest config files ? 

This broke jest test narrowing a while ago :)

```bash
$ yarn run test:unit src/component/Alert # this used to only start Alert tests, now it runs everything
```

iManu (approved):

package.json
matthprost: @chambo-e I didn't saw that issue, I did split configuration file in two :)

QuiiBz (approved):
CI failing at: https://github.com/scaleway/scaleway-ui/runs/4302000882?check_suite_focus=true#step:5:592

DorianMaliszewski (approved):

PR #948, opened 2021-11-30
refactor(avatar): add a11y

QuiiBz:
## Type

- Refactor

#### What is expected?

Make avatar component more accessible

1. Add an `alt` prop to the avatar component which defaults to "Avatar", used when the avatar is an image

2. If the avatar has `text` or `lock` props, the alt is the text or "locked"

3. Add some missing tests

### DIFF
--- a/src/components/Avatar/__tests__/__snapshots__/index.tsx.snap
+++ b/src/components/Avatar/__tests__/__snapshots__/index.tsx.snap
@@ -228,7 +228,7 @@ exports[`Avatar renders correctly with lock 1`] = `
   >
     <div
       aria-label="locked"
-      class="cache-wxe7fo e1frqeh81"
+      class="cache-ha24km e1frqeh81"
     >
       <svg
         class="sc-ui-icon etwatq50 cache-makwam e7ah3bn0"
@@ -424,7 +424,7 @@ exports[`Avatar renders correctly with text and custom style 1`] = `
     class="cache-ua3jbc e7ah3bn0"
   >
     <div
-      class="cache-4iifzb e1frqeh81"
+      class="cache-6jj1vw e1frqeh81"
     >
       HW
     </div>
--- a/src/components/Avatar/__tests__/__snapshots__/index.tsx.snap
+++ b/src/components/Avatar/__tests__/__snapshots__/index.tsx.snap
@@ -227,11 +227,11 @@ exports[`Avatar renders correctly with lock 1`] = `
     class="cache-ua3jbc e7ah3bn0"
   >
     <div
-      aria-label="locked"
       class="cache-ha24km e1frqeh81"
     >
       <svg
         class="sc-ui-icon etwatq50 cache-makwam e7ah3bn0"
+        title="Locked"
         viewBox="0 0 24 24"
       >
         <path
--- a/src/components/Avatar/__tests__/__snapshots__/index.tsx.snap
+++ b/src/components/Avatar/__tests__/__snapshots__/index.tsx.snap
@@ -1,30 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Avatar renders correctly with custom alt 1`] = `
-<DocumentFragment>
-  .cache-ua3jbc.cache-ua3jbc {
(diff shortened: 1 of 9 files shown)


src/components/Avatar/index.tsx
```diff
@@ -73,16 +76,21 @@ const Avatar: FunctionComponent<AvatarProps> = ({
         textBgColor={textBgColor}
         textColor={textColor}
         textSize={textSize}
+        alt={lock ? 'locked' : text}
       >
         {lock ? <Icon name="lock" color="gray550" /> : formatTextToAvatar(text)}
       </StyledDiv>
     ) : (
-      <StyledImg src={image} alt="" />
+      <StyledImg src={image} alt={alt} />
```
chambo-e, line 84: I don't really understand why we need to add an alt here ? 🤔 

Imo this is a decorative image no ? https://www.w3.org/WAI/tutorials/images/decorative/

cc @iManu

matthprost (approved):

src/components/Avatar/index.tsx
```diff
@@ -73,16 +76,21 @@ const Avatar: FunctionComponent<AvatarProps> = ({
         textBgColor={textBgColor}
         textColor={textColor}
         textSize={textSize}
+        alt={lock ? 'locked' : text}
```
vincentaudebert, line 79: I don't get it? `alt` on a `div`? 

Maybe `aria-label` and I'm not even sure?

QuiiBz:
Hi @iManu, could you check the comments above, please? Thanks!

src/components/Avatar/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -33,6 +58,7 @@ exports[`Avatar renders correctly with custom background color 1`] = `
     class="cache-ua3jbc e7ah3bn0"
   >
     <div
+      alt="HW"
```
iManu, line 61: `alt` does not exist on `div` elem, only for `img`

src/components/Avatar/index.tsx
```diff
@@ -21,6 +21,7 @@ const formatTextToAvatar = (text: string): string => {
 }
 
 const StyledDiv = styled.div<{
+  alt: string
```
iManu, line 24: blip

src/components/Avatar/index.tsx
iManu: `alt` is mandatory on image, even if it has no content (empty alt)

https://www.w3.org/WAI/tutorials/images/

src/components/Avatar/index.tsx
```diff
@@ -73,16 +76,21 @@ const Avatar: FunctionComponent<AvatarProps> = ({
         textBgColor={textBgColor}
         textColor={textColor}
         textSize={textSize}
+        alt={lock ? 'locked' : text}
       >
         {lock ? <Icon name="lock" color="gray550" /> : formatTextToAvatar(text)}
```
iManu, line 81: for `lock` state, if you can add a `<title>Locked</title>` tag in the SVG

src/components/Avatar/index.tsx
iManu: if it not possible to add title tag in svg (see my comment below), use `aria-label="locked"` and ONLY if it is lock

QuiiBz:
Looks like we can't have an `aria-label` on a `div`: https://github.com/scaleway/scaleway-ui/runs/4594289371?check_suite_focus=true#step:5:41

Screenshot 2021-12-21 at 12 45 50

src/components/Avatar/index.tsx
chambo-e: @iManu agree, what I meant was why can't we keep it empty ? 

In the link you provided it says 

```
Decorative images: Provide a null text alternative (alt="") when the only purpose of an image is to add visual decoration to the page, rather than to convey information that is important to understanding the page.
```

QuiiBz: Reverted to an empty string as suggested @chambo-e

chambo-e (approved):

PR #958, opened 2021-12-02
feat: add abbr component

QuiiBz:
## Type

- Feature

#### What is expected?

Add a new `<Abbr />` component

1. Add the component with snapshot tests and stories

2. Convert some abbreviations in `<Description />` stories with this new component


chambo-e (commented):
Quite confused by this PR, this is a technical encapsulation in a design system

Would this means we need to encapsulate every of our abbreviation on the console ? 

Can't be just add the <abbr> tags in the concerned stories ?

QuiiBz:
> Quite confused by this PR, this means we need to encapsulate every of our abbreviation on the console ?

@iManu asked for this, but not sure if it's a good idea or not

chambo-e:
@QuiiBz I'm all for having this Abbr component in the console codebase, don't really see the point of having it here, to correct the a11y tests we can directly use the native tag in the concerned stories :)

DorianMaliszewski (approved):

matthprost:
> @QuiiBz I'm all for having this Abbr component in the console codebase, don't really see the point of having it here, to correct the a11y tests we can directly use the native tag in the concerned stories :)

Agree with @chambo-e I think we should directly use native tag or we would have to all tags in Scaleway UI and that's not much the goal 😄

chambo-e:
@QuiiBz yes you can close :)

PR #959, opened 2021-12-02
feat(theme): creating new theme colors

matthprost:
## Type

- Feature

#### What is expected?

New theme implementation

#### New theme implementation and usage

Theme is now welcoming a new file `contrasts.tsx` that include all default contrast shades of the theme.

Shade of contrasts are used to generate colors tokens (`primary.backgroundHover`, `primary.textWeakDisabled`, etc.). 
With the default theme configuration contrast are defined into `contrasts.tsx`, if the user wants to change those contrast and re-generate new tokens definition he can use function `createTheme()` that will return new colors token definitions that can then be provided into theme provider.

In order to avoid error and to keep a complete color palette if the user doesn't define all contrast shade we will take default definition in `contrasts.tsx`. So if per example only `primary.100` is defined all others shade (`primary.200`, `300`...) are not then it will be set with default value declared in `contrasts.tsx`.

In a new project it will look like this:

```js
function App() {
    // We generate custom theme here by providing the shade of colors we want to change
    // all others shade will get default theme value
    const myTheme = createTheme({
        contrasts: {
            primary: {
                100: '#46ff00', // green color
                800: '#46ff00' // green color
            }
        }
    })

    return (
        <ThemeProvider theme={myTheme}>
            <div className="App">
                <Badge variant="primary">Example</Badge>
            </div>
        </ThemeProvider>
    );
}

export default App;
```

Resulting: 
Screenshot 2021-12-06 at 13 59 04

#### Technical aspects

- Old colors have been renamed into `colorsDeprecated`
- An eslint rule has been created to avoid using deprecated colors:
Screenshot 2021-12-06 at 14 15 24
- In order to make migration easier auto-completion should work fully with proposition of all available colors such as:
Screenshot 2021-12-06 at 14 11 56
Screenshot 2021-12-06 at 14 12 07
- As example and testing for future migration `Button` and `Badge` have been migrated.

⚠️  Unfortunately it will probably be impossible not to renew snapshots as neutral color has a specific behavior compared to before. Each components is using a different shade of neutral and has its own definition which made me decide to force one color for neutral. This is a mandatory step and we cannot do differently as each component was using its own gray shade it was previously decided component by component and now we try to tend to generic color.


src/index.ts
```diff
@@ -2,3 +2,4 @@ export * from './components'
 export { default as theme } from './theme'
 export type { SCWUITheme } from './theme'
 export * from './utils'
+export { createTheme } from './theme'
```
chambo-e, line 5: This is an extend theme rather than a create no ?

src/theme/index.ts
```diff
@@ -41,16 +42,47 @@ const fonts = {
...
 }
 
+const createTheme = ({
+  contrasts,
+  space: newSpace,
+  screens: newScreens,
+  radii: newRadii,
+  fonts: newFonts,
+}: {
+  contrasts?: { [key in Color]?: Partial<ContrastType> }
```
chambo-e, line 59:
```suggestion
  contrasts?: Record<Color, Partial<ContrastType>>
```

?

src/theme/colors/index.ts
```diff
@@ -0,0 +1,144 @@
...
+        contrast: deepMergedContrasts[contrast as Color],
+        neutralContrast: deepMergedContrasts.neutral,
+        sentiment: contrast,
+      }),
+    }),
+    {},
+  ) as { [key in Color]: ReturnType<typeof generateTokens> }
+}
+
+const colors: ReturnType<typeof colorsTokens> = colorsTokens(localContrasts)
```
chambo-e, line 142: Already implied no ? 

```suggestion
const colors = colorsTokens(localContrasts)
```

src/components/RadioBorderedBox/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -507,7 +507,7 @@ exports[`RadioBorderedBox renders correctly with label desc and badge 1`] = `
   width: -moz-fit-content;
   width: fit-content;
   background-color: #ef5774;
-  color: #fff;
+  color: #ffffff;
```
chambo-e, line 510: Why ? 🤔

.eslintrc.js
```diff
@@ -0,0 +1,102 @@
+const generateColorsRules = () => ({
```
chambo-e, line 1: No need to move this into a js file, you only have hardcoded rules :)

chambo-e: Also be careful you overwrite the no-restricted-syntax rule https://github.com/scaleway/scaleway-lib/blob/d0e138501ae3c7fb59b8e5924504e8cef434070e/packages/eslint-config-react/shared.js#L25

src/index.ts
matthprost: @chambo-e Kinda yeah, I used createTheme as it create theme from what is given and it seems like MUI are doing same but I'm open if you want to rename it

src/theme/colors/index.ts
matthprost: haha right

src/components/RadioBorderedBox/__tests__/__snapshots__/index.tsx.snap
matthprost: Because:

```js
colorsDeprecated: {
   white: = '#fff'
}

colors: {
  neutral: {
    backgroundWeak: '#ffffff'
  }
}
```

adriengibrat: image

chambo-e: Yes, why not keep #fff ? :)

matthprost: I wanted to have consistency in terms of how to write colors but actually I can put old way and we will change it later

DorianMaliszewski (changes requested):

src/theme/index.ts
```diff
@@ -41,16 +42,47 @@ const fonts = {
...
+  fonts: newFonts,
+}: {
+  contrasts?: Record<Color, Partial<ContrastType>>
+  space?: Spaces
+  screens?: ScreenSize
+  radii?: Record<string, string>
+  fonts?: Record<string, string>
+}) => ({
+  ...theme,
+  ...(contrasts ? { colors: colorsTokens(contrasts) } : undefined),
```
chambo-e, line 66: You can't spread undefined :)

```suggestion
  ...(contrasts ? { colors: colorsTokens(contrasts) } : {}),
```

chambo-e (changes requested):
Missing some tests for createTheme, except that all good :)

src/theme/index.ts
matthprost: Oh right my bad I changed it

matthprost:
> Missing some tests for createTheme, except that all good :)

@chambo-e I added unit test for createTheme() it should be all good now :D

src/theme/__tests__/index.ts
```diff
@@ -0,0 +1,294 @@
...
+        medium: 768,
+        small: 576,
+        xlarge: 1200,
+        xsmall: 0,
+      },
+      space: {
+        '0': '6px',
+        '9': '600px',
+      },
+      /* eslint-enable */
```
chambo-e, line 291:
```suggestion
      /* eslint-enable sort-keys */
```

src/theme/colors/index.ts
```diff
@@ -0,0 +1,144 @@
...
+      borderHover: contrast[800],
+      borderDisabled: contrast[800],
+      borderWeak: contrast[800],
+      borderWeakHover: contrast[800],
+      borderWeakDisabled: contrast[800],
+      borderStrong: contrast[800],
+      borderStrongHover: contrast[800],
+      borderStrongDisabled: contrast[800],
+
+      /* eslint-enable */
```
chambo-e, line 66:
```suggestion
      /* eslint-enable sort-keys */
```

src/theme/colors/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    borderHover: contrast[800],
+    borderDisabled: contrast[800],
+    borderWeak: contrast[800],
+    borderWeakHover: contrast[800],
+    borderWeakDisabled: contrast[800],
+    borderStrong: contrast[800],
+    borderStrongHover: contrast[800],
+    borderStrongDisabled: contrast[800],
+
+    /* eslint-enable */
```
chambo-e, line 106:
```suggestion
    /* eslint-enable sort-keys */
```

Might be a good idea to turn off this rule once and for all 

@philibea do you still find any value in this rule ?

matthprost: This rule was for propTypes and props in general if I remember well

chambo-e (approved):

src/theme/colors/index.ts
chambo-e: There is a dedicated rule for propTypes https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-prop-types.md

DorianMaliszewski (approved):

PR #1030, opened 2021-12-20
fix: a11y for tooltipicon

vincentaudebert:
## Type

- Enhancement

#### What is expected?

Better accessibility for tooltipicon

(Describe what you did)

1. Adds aria-label on icon so the tooltip gets read

### DIFF
--- a/src/components/TooltipIcon/__tests__/__snapshots__/index.tsx.snap
+++ b/src/components/TooltipIcon/__tests__/__snapshots__/index.tsx.snap
@@ -19,6 +19,7 @@ exports[`TooltipIcon renders with default props 1`] = `
     tabindex="0"
   >
     <svg
+      aria-label="help-circle-outline"
       class="sc-ui-icon etwatq50 cache-byq9xb e7ah3bn0"
       viewBox="0 0 24 24"
     >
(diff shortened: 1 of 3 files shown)


matthprost (approved):

src/components/TooltipIcon/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -19,6 +19,7 @@ exports[`TooltipIcon renders with default props 1`] = `
     tabindex="0"
   >
     <svg
+      aria-label="help-circle-outline"
```
adriengibrat, line 22: Il serait plus sympa de passer la chaîne à la traduction, mais je ne sais pas évaluer si ça vaut vraiment le coup...

vincentaudebert: Yes it was an option but it means we need to provide a i18n text just for the SVG label :/ not sure it is really worth it?

DorianMaliszewski (approved):

src/components/TooltipIcon/__tests__/__snapshots__/index.tsx.snap
vincentaudebert: @adriengibrat I will let it go as is for now. We can think about something new a bit later maybe if we feel the need.

PR #1031, opened 2021-12-20
fix: a11y for placeholder

vincentaudebert:
## Type

- Enhancement

#### What is expected?

A11y for placeholder component

(Describe what you did)

1. Add progress cursor

2. Add media query for soft/no animation

3. Add comment into storybook to remind about accessibility for Placeholder container.


src/components/Placeholder/index.tsx
```diff
@@ -94,7 +100,11 @@ Placeholder.propTypes = {
 }
 
 Placeholder.defaultProps = {
+  col: undefined,
```
chambo-e, line 103: Why ? 🤔

src/components/Placeholder/index.tsx
```diff
@@ -69,17 +81,11 @@ const Placeholder: React.VoidFunctionComponent<PlaceholderProps> = ({
   const Component = variants[variant]
 
   return (
-    <Box
-      position="relative"
-      display="block"
-      width="100%"
-      overflow="hidden"
-      {...props}
```
chambo-e, line 77: You sure ? There is a lot of Placeholder that still use xstyled props :)

src/components/Placeholder/index.tsx
vincentaudebert: To satisfy linter on commit :/

vincentaudebert: ie. 

95:3   error    propType "col" is not required, but has no corresponding defaultProps declaration     react/require-default-props
  96:3   error    propType "height" is not required, but has no corresponding defaultProps declaration  react/require-default-props
  97:3   error    propType "length" is not required, but has no corresponding defaultProps declaration  react/require-default-props
  99:3   error    propType "width" is not required, but has no corresponding defaultProps declaration   react/require-default-props

vincentaudebert: Oh wait. I need to remove defaultProps completely 😂

chambo-e (approved):

matthprost (approved):

PR #1032, opened 2021-12-21
refactor(a11y): stepper and a11y tests fix

matthprost:
## Type

- Refactor

#### What is expected?

Make stepper more accessible.

1. Changed center touchable into box as button can't have nested interactions + added aria live and role

2. Fixed abbr inside stepper when added + made it not selectable

3. Fixed a11y tests: it was doing an includes(component) which is not properly working as by adding Stepper in tests it was also testing NavigationStepper. I fixed it to be more strict.

### DIFF
--- a/src/__tests__/a11y.test.tsx
+++ b/src/__tests__/a11y.test.tsx
@@ -19,6 +19,7 @@ const testedComponents = [
   'Reminder',
   'Separator',
   'StealthCopiable',
+  'Stepper',
   'Toaster',
   'TooltipIcon',
   'Tag',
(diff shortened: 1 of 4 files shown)


src/components/Stepper/index.tsx
```diff
@@ -1,5 +1,6 @@
 import { Theme, css } from '@emotion/react'
 import styled from '@emotion/styled'
+import { x } from '@xstyled/emotion'
```
chambo-e, line 3: Nope nope nope, nope nope 😂

src/components/Stepper/index.tsx
```diff
@@ -278,12 +288,10 @@ const Stepper: VoidFunctionComponent<StepperProps> = ({
...
           }}
           value={inputValue.toString()} // A dom element can only have string attributes.
+          aria-label="Input"
         />
-
-        <StyledInput disabled={disabled} as="span">
-          {text}
-        </StyledInput>
-      </StyledCenterTouchable>
+        {text ? <StyledAbbr disabled={disabled}>{text}</StyledAbbr> : null}
```
chambo-e, line 293: How can we be sure that it will be an abbreviation  ? Shouldn't we instead allow passing a node ?

src/components/Stepper/index.tsx
```diff
@@ -80,7 +81,8 @@ const StyledTouchable = styled(Touchable, {
   margin: 0 4px;
 `
 
-const StyledCenterTouchable = styled(Touchable)<{ size: ContainerSizesType }>`
+const StyledCenterBox = styled(Box)<{ size: ContainerSizesType }>`
```
chambo-e, line 84: Why use a Box ?

src/__tests__/a11y.test.tsx
```diff
@@ -8,28 +8,36 @@ import * as process from 'process'
...
-        testedComponents.some(component =>
-          filename.toLowerCase().includes(component.toLowerCase()),
-        )
-      ) {
-        // recursive search in case if directory
-        searchFileFromDir(filename, filter)
+      // recursive search in case if directory
+      searchFileFromDir(filePath, filter)
+    } else if (filePath.indexOf(filter) >= 0) {
+      const isTested = testedComponents.some(
```
chambo-e, line 33: A regex would be easier no ? 

```js
`^${fileName}` ~= filter
```

chambo-e (changes requested):

src/components/Stepper/index.tsx
chambo-e: Not always, eg on Function :)

matthprost: oh right I didn't see this one, I will change it then :)

src/components/Stepper/index.tsx
matthprost: my bad I just copied / paste really fast without thinking

src/components/Stepper/index.tsx
vincentaudebert: I was about to comment the same. Any reason @matthprost ?

vincentaudebert (approved):

chambo-e (approved):

PR #1033, opened 2021-12-21
refactor(tags): add a11y

QuiiBz:
## Type

- Migration

Add `aria-label` to the close button on `Tag` and `Tags`.
I asked @ iManu, no need to add the name of the tag in the `aria-label`.

### DIFF
--- a/src/__tests__/a11y.test.tsx
+++ b/src/__tests__/a11y.test.tsx
@@ -18,6 +18,8 @@ const testedComponents = [
   'StealthCopiable',
   'Toaster',
   'TooltipIcon',
+  'Tag',
+  'Tags',
 ]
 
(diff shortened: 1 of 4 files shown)


chambo-e (approved):

chambo-e (changes requested):
Needs to be renamed to avoid useless package bump, that's not a feat :)

QuiiBz:
> Needs to be renamed to avoid useless package bump, that's not a feat :)

Done, thanks for the reminder!

chambo-e (approved):

vincentaudebert (changes requested):
Add component into a11y.test.tsx

vincentaudebert (approved):

PR #1035, opened 2021-12-21
refactor(status-indicator): add a11y

QuiiBz:
## Type

- Migration

1. Always require the `tooltip` prop as discussed

2. Add `aria-label` with the value of the tooltip to the `Dot`


QuiiBz:
Closing since `aria-label` is not allowed, and requiring the `tooltip` prop isn't a good idea since it's sometimes used only to indicate status with a label next to it.

PR #1037, opened 2021-12-21
fix: a11y for markdown but mostly due to reminder

vincentaudebert:
## Type

- Enhancement

#### What is expected?

a11y for reminder but only impacts markdown component

1. removes span for inline mode to avoid having p tag inside a span tag.


src/components/MarkDown/index.tsx
```diff
@@ -61,11 +61,7 @@ const RootRendererComponent: FunctionComponent<{
   inline?: boolean
   parentProps?: XStyledProps
 }> = ({ inline = false, parentProps = {}, children }) => (
-  <StyledContainer
-    as={inline ? 'span' : 'div'}
-    inline={inline}
-    {...parentProps}
-  >
+  <StyledContainer as="div" inline={inline} {...parentProps}>
```
chambo-e, line 64: So this will always be a div now ? 

This is going to break a lot of Markdown components :)

vincentaudebert: it should not

If inline prop is on, we css p tags to be inlined. It should be enough. We need to check if we have some places with multiple markdowns following each other but that's it?

In any case this is required as p tag inside a span tag isn't correct :(

chambo-e (approved):

src/components/MarkDown/index.tsx
chambo-e: 👌 

Btw the as="div" is useless, Box is by default a div

vincentaudebert: Good point I fix it

matthprost (approved):

PR #1038, opened 2021-12-21
feat: new colors in switch button

DorianMaliszewski:
## Type

- Enhancement

#### What is expected?

Use new colors in switch button

### DIFF
--- a/src/components/SwitchButton/index.tsx
+++ b/src/components/SwitchButton/index.tsx
@@ -32,7 +32,7 @@ const variants = {
 
       &[aria-checked='true'] {
         background-color: ${theme.colors.primary.backgroundStrong};
-        color: ${theme.colors.neutral.textStrong};
+        color: ${theme.colors.primary.textStrong};
       }
 
       &[aria-disabled='true'] {
--- a/src/components/SwitchButton/index.tsx
+++ b/src/components/SwitchButton/index.tsx
@@ -14,37 +14,43 @@ import Box, { BoxProps } from '../Box'
 import Tooltip from '../Tooltip'
 
 const variants = {
-  segment: ({ theme }: { theme: Theme }) => css`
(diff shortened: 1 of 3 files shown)


src/components/SwitchButton/index.tsx
```diff
@@ -14,37 +14,43 @@ import Box, { BoxProps } from '../Box'
...
+      &[aria-disabled='true'] {
+        color: ${theme.colors.neutral.textDisabled};
+
+        &[aria-checked='true'] {
+          color: ${theme.colors.neutral.textDisabled};
+          background-color: ${theme.colors.neutral.backgroundDisabled};
+        }
+      }
+
+      &:not([aria-checked='true']):not([aria-disabled='true']) {
```
QuiiBz, line 48: Can't we do this? It's only not supported in IE https://developer.mozilla.org/fr/docs/Web/CSS/:not#compatibilit%C3%A9_des_navigateurs

```suggestion
      &:not([aria-checked='true'], [aria-disabled='true']) {
```

QuiiBz (approved):

src/components/SwitchButton/index.tsx
```diff
@@ -14,37 +14,43 @@ import Box, { BoxProps } from '../Box'
...
+      font-size: 14;
+      height: 40px;
+      transition: none;
+      border-radius: 4px;
       box-shadow: none;
-      border-color: ${theme.colorsDeprecated.transparent};
-      color: ${theme.colorsDeprecated.gray700};
-    }
+      border-color: transparent;
+      background-color: ${theme.colors.neutral.backgroundWeak};
```
matthprost, line 25: Why adding background while it wasn't there before ?

matthprost (changes requested):

vincentaudebert (approved):

PR #1041, opened 2021-12-21
feat: provide modern normalize function

chambo-e:
## Type

- Feature

### Summarise concisely:

This is to avoid to some magic fuckery we started doing since the removal of GlobalStyle and have a shared ground of css development 

## Relevant logs

| url  | Screenshot 2021-12-21 at 17 33 18 | Screenshot 2021-12-21 at 17 33 13

### DIFF
--- a/src/components/Button/__stories__/index.stories.tsx
+++ b/src/components/Button/__stories__/index.stories.tsx
@@ -141,23 +141,17 @@ Action.parameters = {
 }
 Action.decorators = [
   () => (
-    <>
+    <div style={{ display: 'flex', flexWrap: 'wrap', gap: '16px' }}>
       {icons.map(icon => (
-        <Button action key={icon} icon={icon} mr={2} mb={2} />
+        <Button action key={icon} icon={icon} />
       ))}
       {buttonVariants.map(variant => (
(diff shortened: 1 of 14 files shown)


vincentaudebert (approved):
good idea 👍

vincentaudebert:
Screenshot 2021-12-21 at 18 22 15

Not linked to ticket but it looks like textbox and richselect doesn't have the same disabled colors 😅

src/components/Button/__stories__/index.stories.tsx
```diff
@@ -141,23 +141,17 @@ Action.parameters = {
 }
 Action.decorators = [
   () => (
-    <>
+    <div style={{ display: 'flex', flexWrap: 'wrap', gap: '16px' }}>
```
QuiiBz, line 144: I think this is happening a few times in scaleway-ui, shouldn't we replace all mr/mb with a flex gap on the parent? (in a separate PR)

QuiiBz (approved):

src/components/Button/__stories__/index.stories.tsx
chambo-e: Yes absolutely, we are in the process of removing xstyled props, so whenever you can please do :)

PR #1049, opened 2021-12-23
fix(tags): onChange prop type definition

QuiiBz:
## Type

- Bug

Update the `onChange` prop type definition so each tag is only a `string`, not a `string | null | undefined`

### DIFF
--- a/src/components/Tags/index.tsx
+++ b/src/components/Tags/index.tsx
@@ -115,7 +115,7 @@ type TagsProps = {
   id?: string
   manualInput?: boolean
   name?: string
-  onChange?: (tags: (string | null | undefined)[]) => void
+  onChange?: (tags: string[]) => void
   onChangeError?: (error: Error | string) => void
   placeholder?: string
   tags?: TagsProp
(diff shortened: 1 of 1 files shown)


src/components/Tags/index.tsx
```diff
@@ -145,9 +145,14 @@ const Tags: VoidFunctionComponent<TagsProps> = ({
   const inputRef = useRef<HTMLInputElement>(null)
 
   const dispatchOnChange = (newState: TagsProp) => {
-    const changes = newState.map(tag =>
-      typeof tag === 'object' ? tag?.label : tag,
-    )
+    function isTagDefined(tag: string | null | undefined): tag is string {
```
chambo-e, line 148: I don't understand why this is necessary ? 🤔

QuiiBz: Typescript won't infer the type with a normal filter without specifying the `tag is string` part:

Screenshot 2021-12-23 at 10 14 51

chambo-e: What if you move away from VoidFunctionComponent ?

QuiiBz: Same issue, if I try to manually type the `filter`, typescript effectively tells me to use a type predicate, so I don't think there's another way to solve this:

Screenshot 2021-12-23 at 10 23 01

chambo-e: You also need to set 

```ts
type TagsProp = (
  | string
  | { label: string; index: string }
)[]

```

QuiiBz: Awesome!

chambo-e (approved):

matthprost (approved):

PR #1055, opened 2021-12-24
fix: switch input

DorianMaliszewski:
## Type

- Bug
- Enhancement

#### What is expected?

Input should not be under the ball

1. Ball is behind the input

2. Make some beautiful styles

### DIFF
--- a/src/components/Switch/index.tsx
+++ b/src/components/Switch/index.tsx
@@ -57,8 +57,16 @@ const SwitchVariantsStyles = {
       background-color: ${theme.colors.primary.backgroundStrong};
     }
 
-    &[aria-checked='true'] > ${StyledInsideLabel} {
-      color: ${theme.colors.neutral.textStrong};
+    &[aria-checked='true'] {
+      & > ${StyledInsideLabel} {
+        color: ${theme.colors.neutral.textStrong};
+      }
+
+      &:focus-within,
+      &:focus {
+        box-shadow: 0 0 1px 2px ${theme.colors.neutral.backgroundWeak},
+          0 0 1px 3px ${theme.colors.primary.backgroundStrong};
+      }
     }
   `,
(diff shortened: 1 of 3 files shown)


src/components/Switch/index.tsx
```diff
@@ -239,16 +266,7 @@ const Switch: VoidFunctionComponent<SwitchProps> = ({
           aria-checked={checked}
           data-variant={variant}
           data-size={size}
-          role="checkbox"
```
QuiiBz, line 242: Can we instead use the `switch` role? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/switch_role

src/components/Switch/index.tsx
```diff
@@ -94,12 +112,18 @@ const StyledSwitch = styled('div', {
   border-radius: 34px;
   position: relative;
   transition: all 250ms;
-  cursor: pointer;
```
QuiiBz, line 97: Shouldn't we have a `cursor: pointer` on the whole switch? And `cursor: not-allowed` when disabled

philibea: i'm agree with @QuiiBz, it's weird when you try to click.

https://user-images.githubusercontent.com/14060273/147454947-e4162516-8c63-49e4-a27a-7bf53d26bae4.mov

vincentaudebert: Vu avec Dodo, non necessaire le cursor pointer est deja la.

src/components/Switch/index.tsx
vincentaudebert: Vu avec Dodo non necessaire

vincentaudebert (approved):

chambo-e (approved):

PR #1071, opened 2021-12-29
fix: alert width to allow dismiss button

vincentaudebert:
## Type

- Enhancement

#### What is expected?

Makes alert content full width so it can allow a button on far right end.

Nothing should change visually for existing content.

### DIFF
--- a/src/components/Alert/index.tsx
+++ b/src/components/Alert/index.tsx
@@ -79,6 +79,10 @@ const StyledContainer = styled(Box, {
   ${alertStyles}
 `
 
+const StyledIcon = styled(Icon)`
+  margin-right: ${({ theme }) => theme.space[2]};
+`
+
 const StyledBox = styled(Box)<{ color: string }>`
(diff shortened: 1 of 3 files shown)


src/components/Alert/index.tsx
```diff
@@ -123,10 +124,11 @@ const Alert: FunctionComponent<AlertProps> = ({
       name={icon || typesDefaultIcons[type]}
       size={iconSize}
       aria-hidden="true"
+      mr={2}
```
chambo-e, line 127: Go styled :)

chambo-e (changes requested):
Maybe add an example in storybook ?

vincentaudebert:
> Maybe add an example in storybook ?

Having a button on the far right is just some tweaks with Alert content. Nothing linked to the component itself sadly.

chambo-e (approved):

QuiiBz (approved):

PR #1140, opened 2022-01-28
fix(stealth-copiable): added new prop and fixed line-height

matthprost:
## Type

- Enhancement

#### What is expected?

I have removed line-height and padding so button doesn't make parent div bigger + I changed typing for `tooltip` prop as original tooltip component can receive a ReactNode.

### DIFF
--- a/src/components/StealthCopiable/index.tsx
+++ b/src/components/StealthCopiable/index.tsx
@@ -38,7 +38,6 @@ type StealthCopiableProps = {
   side?: 'left' | 'right'
   copyText?: string
   copiedText?: string
-  hide?: boolean
 }
 
 const StealthCopiable: FunctionComponent<StealthCopiableProps> = ({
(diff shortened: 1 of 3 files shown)


src/components/StealthCopiable/index.tsx
```diff
@@ -38,20 +38,24 @@ type StealthCopiableProps = {
...
   copiedText = 'Copied',
+  hide = false,
 }) => {
   const string = recursivelyGetChildrenString(children)
 
   const [isCopied, setCopied] = useClipboard(string, {
     successDuration: 5000,
   })
 
+  if (hide) return <StyledContainer>{children}</StyledContainer>
```
chambo-e, line 57: Can't this be handled by the parent ? This is not really a Copiable component with hide=true :)

matthprost: @chambo-e You're right, component should stay pure. I will make a custom component that will handle this case directly in the project

chambo-e (approved):

DorianMaliszewski (approved):

PR #1173, opened 2022-02-14
fix(stealth-copiable): fix text overflow

matthprost:
## Type

- Enhancement

1. Fixed text overflow:

Screenshot 2022-02-14 at 12 18 59

### DIFF
--- a/src/components/StealthCopiable/index.tsx
+++ b/src/components/StealthCopiable/index.tsx
@@ -24,13 +24,20 @@ const UnselectableSpan = styled.span`
 `
 
 const StyledContainer = styled.div`
-  display: block;
+  display: flex;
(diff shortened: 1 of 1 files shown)


src/components/StealthCopiable/index.tsx
```diff
@@ -33,6 +33,13 @@ const StyledContainer = styled.div`
   }
 `
 
+const StyledContent = styled.span<{ width?: number | string }>`
```
chambo-e, line 36: width is never used ?

matthprost: oops my bad it's not necessary any more

philibea (approved):

PR #1229, opened 2022-02-28
ci(deploy): add serverless deploy

philibea:
## Type

- CI


.github/workflows/teardown_pull_request.yml
```diff
@@ -8,28 +8,32 @@ jobs:
...
+        id: teardown
         with:
-          docker_host: ${{ env.REMOTE_SSH_SERVER }}
-          context_name: v-${{ env.BRANCH_SLUG }}
-          ssh_cert: ${{ secrets.SSH_CERT }}
-          ssh_key: ${{ secrets.SSH_KEY }}
-      - name: Remove deployment
-        env:
-          VERSION: v-${{ env.BRANCH_SLUG }}
-        run: docker-compose -p ${VERSION} --context ${VERSION} down --rmi all
```
chambo-e, line 32: This means we can remove the docker-compose :)

chambo-e (changes requested):

.github/workflows/teardown_pull_request.yml
philibea: Yes, right, i just need to handle an other case when we want to deploy on our main dns actually it's not what we want.

chambo-e (approved):

PR #1242, opened 2022-03-04
fix(statebar): add new case on stateBar

Lawndlwd:
## Type

- Feature

#### What is expected?

when the stateBar value is between 70 && 90 the progress bar background color will be orange 

(Describe what you did)

1. add new condition on the variant when the value  is between 70 && 90  to display orange color

Capture d’écran 2022-03-04 à 17 33 47


src/components/StateBar/index.tsx
```diff
@@ -1,6 +1,7 @@
 import { css } from '@emotion/react'
 import PropTypes from 'prop-types'
 import React, { ReactNode, useMemo } from 'react'
+import { colors } from '../../theme'
```
chambo-e, line 4: We try to avoid importing directly from the theme file, could you use the injected Theme in emotion instead ?

chambo-e (changes requested):
Small comment and also we shouldn't post internal links publicly, I edited your description :)

matthprost:
Screenshot 2022-03-04 at 14 42 32

You can add an example in the documentation of storybook :)

Lawndlwd:
> Screenshot 2022-03-04 at 14 42 32
> 
> You can add an example in the documentation of storybook :)

 Yes, and we have this issue that the imported color from theme wouldn't appear on storybook, any suggestion @matthprost

src/components/StateBar/index.tsx
```diff
@@ -53,12 +53,15 @@ export const StateBarBar = ({
   value = 0,
   ...props
 }: StateBarBarProps): JSX.Element => {
+  const theme = useTheme()
+
   const variant = useMemo(() => {
     if (unlimited) return 'success'
```
philibea, line 59: we can now using the new colors everywhere ? this will avoid misunderstood where the colors come from ?

philibea:
```js  
if (unlimited) return 'success'
```

will be 
```js
  if (unlimited) return theme.colors.therightcolorstobeused ?
```

Lawndlwd: Good point @philibea, there is an issue that I didn't understand `theme.colors.succes.background` or whatever color from the new colors don't work, don't work I mean the bar show without color on storybook like the photo above

chambo-e: @philibea ProgressBar expect a variant, not a color :)

chambo-e (approved):

src/components/StateBar/index.tsx
Lawndlwd: @chambo-e  do you think that the history of variant need to be changed to accept color as well as variant?

chambo-e: @Lawndlwd nope, a variant is a variant, if we start accepting colors we are going to be lost in no time 😄

philibea (approved):

PR #1253, opened 2022-03-08
feat(design): use new sync colors and audit all components

matthprost:
## Type

- Feature

#### What is expected?

To remove all colors by using new synchronised colors from figma. Add dark theme on storybook and expose it to the library.
Audit all components as matching colors might not be same (especially neutral) and change them to be correct to old verison.

1. Updated all colors for each components and checked one by one that they are correctly displayed + tested on shire

2. Added real dark theme and exposed it

The main goal is to be retro-compatible with previous version awaiting design review of each components one by one.

⚠️ To make you review easier I recommend this filter:

Screenshot 2022-03-09 at 15 38 48

Screenshot 2022-03-09 at 16 36 44


src/components/Alert/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -16,8 +16,8 @@ exports[`Alert renders correctly unknow theme type and fallback to defaults 1`]
   align-items: center;
   border-radius: 4px;
   padding: 12px;
-  background-color: #ffe1e7;
-  color: #ef5774;
+  background-color: #FFE1E7;
```
chambo-e, line 19: Why is there some colors on all caps and other on lower ?

src/index.ts
```diff
@@ -1,4 +1,4 @@
 export * from './components'
-export { default as theme, extendTheme, createTheme } from './theme'
+export { default as theme, extendTheme, darkTheme } from './theme'
```
chambo-e, line 2: Maybe export default theme as `lightTheme` ?

src/theme/deprecated/helper.ts
```diff
@@ -1,147 +0,0 @@
-/**
```
chambo-e, line 1: This will break console no ?

src/theme/index.ts
```diff
@@ -47,35 +50,13 @@ const theme = {
   colors,
   colorsDeprecated,
   fonts,
+  others,
```
chambo-e, line 53: others ???

src/theme/index.ts
```diff
@@ -91,12 +72,11 @@ const extendTheme = ({
 }: {
   baseTheme?: typeof theme
   extendedTheme: RecursivePartial<typeof theme>
-}) => deepmerge(baseTheme, extendedTheme)
+}) => deepmerge(baseTheme, extendedTheme) as typeof theme
```
chambo-e, line 75: deep merge does not accept a generic ?

src/components/Alert/__tests__/__snapshots__/index.test.tsx.snap
matthprost: It comes from figma, I can ask design team to make it uniform

src/theme/index.ts
matthprost: Its all tokens that can't be in normal behavior, like overlay background that is not part of normal tokens and should be separated to "specific" category.

If it sounds strange we can put this overlay into `colors.overlay` right away 😄

src/components/Alert/__tests__/__snapshots__/index.test.tsx.snap
chambo-e: We can force a lowercase with jq ?

src/theme/deprecated/helper.ts
chambo-e: nah never mind, you did not removed colorsDeprecated :)

src/theme/index.ts
chambo-e: yeaaaaah but the theme does not only represents colors, the `other` or `extra` key at root is the beginning of a big mess 😄 

These are still colors ? Maybe colors.others ?

matthprost: From now it is still colors, so I think I can still put in into `colors.overlay` it will make sense

adriengibrat:
image
oops .loki/reference/chrome_laptop_Components_Data_Entry_Switch_Custom_Label_Render.png

adriengibrat:
image
oops .loki/reference/chrome_laptop_Components_Data_Entry_Switch_Custom_Label_Render.png

adriengibrat:
image
.loki/reference/chrome_laptop_Components_Foundation_Link_Variants.png
white variant -> black color

adriengibrat:
Drop of placeholders contrast
image

adriengibrat:
Huge contrast diff on .loki/reference/chrome_laptop_Components_List_Table_Variant.png

image

src/theme/index.ts
matthprost: @chambo-e I did a little fix of script + this change right here: https://github.com/scaleway/scaleway-ui/pull/1259

src/theme/index.ts
matthprost: @chambo-e It does but both arguments doesn't have same types, first argument is of type `typeof theme` and second argument is `RecursivePartial<typeof theme>`.

If I put `typeof theme` as generic it will trow an error on `extendedTheme` parameter, that it's type `RecursivePartial<typeof theme>` and not `typeof theme`.

I could put `RecursivePartial<typeof theme>` as generic, but then `extendTheme` function will return type `RecursivePartial<typeof theme>` and that's not what we want, we want returned type `typeof theme` for our function `extendTheme`, as it is what the function return.

So that's the solution I found to be best working in this case

matthprost: Sorry for long post, here is a potato

potato

scripts/figma-synchronise-tokens.sh
```diff
@@ -22,7 +22,7 @@ function generateTokens {
   reduce (.[$theme].shades | to_entries | .[]) as $sentiment
     ({}; . + {"\($sentiment.key)": (
       reduce($sentiment.value | to_entries | .[]) as $shade
-        ({}; . + { "\($shade.key)": $shade.value.value })) })
+        ({}; . + { "\($shade.key)": ($shade.value.value | ascii_upcase) })) })
```
chambo-e, line 25:
```suggestion
        ({}; . + { "\($shade.key)": ($shade.value.value | ascii_downcase) })) })
```

To reduce amount of changes in snaps ?

matthprost: Sure thing :)

chambo-e (approved):

vincentaudebert:
y a un ptit fail detecté sur .loki/difference/chrome_laptop_Components_Feedback_ActivityIndicator_Trail_Color.png

adriengibrat (approved):

PR #1259, opened 2022-03-10
fix: script and tokens

matthprost:
## Type

- Enhancement

#### What is expected?

Schedule script wasn't much working + I moved others into color

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "test:visual:update": "yarn run build:storybook-visual && yarn loki update",
     "test:visual": "yarn loki test",
     "prepare": "husky install",
-    "size": "yarn run build && size-limit"
+    "size": "yarn run build && size-limit",
+    "tokens:update": "./scripts/figma-synchronise-tokens.sh https://raw.githubusercontent.com/scaleway/design-tokens/main/tokens.json && yarn run format"
   },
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -51,7 +51,8 @@
     "test:visual:update": "yarn run build:storybook-visual && yarn loki update",
     "test:visual": "yarn loki test",
     "prepare": "husky install",
-    "size": "yarn run build && size-limit"
+    "size": "yarn run build && size-limit",
+    "tokens:update": "./scripts/figma-synchronise-tokens.sh https://raw.githubusercontent.com/scaleway/design-tokens/main/tokens.json && yarn run format"
```
chambo-e, line 55: Better to run them outside in the action


```yarn run tokens:update && yarn run format```

philibea (approved):

package.json
matthprost: @chambo-e That's what I tough first but then I tough that anyone who will run `yarn run tokens:update` locally and won't have file correctly formatted will probably struggle to understand. I found it easier to just have one command that does it all, what do you think ?

chambo-e: ok for me :)

chambo-e (approved):

PR #1264, opened 2022-03-11
feat(list): added some animation on list

matthprost:
## Type

- Enhancement

#### What is expected?

As List was already have an animation I tough it would be cool to have them all as we already did for RichSelect. I will use animation for instance when adding new volume and I think there might be others use case where animation will make console look better.

https://user-images.githubusercontent.com/15812968/157975916-d52e35be-4164-4467-8a98-daaac092f104.mov

https://user-images.githubusercontent.com/15812968/157975930-f793ccdb-adef-4a94-9a77-384832e29d42.mov


src/components/List/__stories__/index.stories.tsx
```diff
@@ -518,3 +521,80 @@ PaginationLoadingPageCount.decorators = [
     )
   },
 ]
+
+export const Animated = Template.bind({})
```
adriengibrat, line 525: Excellent demo ;)
An UX review should be as important as the code review for this!

src/components/List/variantProduct.tsx
```diff
@@ -305,13 +299,15 @@ export const Row: FunctionComponent<ListRowProps> = ({
   id,
   children,
   animated,
-  edition,
-  isEditable,
-  isHoverable,
-  locked,
+  animation = 'fadeIn',
```
adriengibrat, line 302: should use type `keyof animations`

src/components/List/__stories__/index.stories.tsx
matthprost: @adriengibrat Probably animations will be more complete once design system process will start and we will have more from design team :)

src/components/List/variantTable.tsx
```diff
@@ -73,13 +64,22 @@ const StyledRow = styled(Box, {
...
+  animationDuration?: number
+  disabled?: boolean
+  highlighted?: boolean
+}>`
+  ${({ animated, animationDuration, animation }) =>
     animated
       ? css`
-          animation: ${fadeInAnimation} 1s linear;
+          animation: ${(
+              animations as Record<string, ReturnType<typeof keyframes>>
```
chambo-e, line 78: why ?

src/components/List/variantTable.tsx
```diff
@@ -73,13 +64,22 @@ const StyledRow = styled(Box, {
...
+            )[animation as string]}
+            ${animationDuration as number}ms linear;
```
chambo-e, line 80: why ?

chambo-e (changes requested):

chambo-e (approved):

philibea (approved):

PR #1282, opened 2022-03-16
fix: adjust disabled colors for button and switchbutton

matthprost:
## Type

- Bug

Some disabled colors where not same as before I changed them in order to make no changes on shire.


adriengibrat:
Contrast on disabled elements is very low, but I guess it's on purpose.
Otherwise :+1:

adriengibrat (approved):

PR #1299, opened 2022-03-22
chore: update package json tokens sync

matthprost:
Fix package json to use directly prettier and prettify only token when updating them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "test:visual": "pnpm loki test",
     "prepare": "husky install",
     "size": "pnpm run build && size-limit",
-    "tokens:update": "./scripts/figma-synchronise-tokens.sh https://raw.githubusercontent.com/scaleway/design-tokens/main/tokens.json && pnpm run format"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -52,7 +52,7 @@
     "test:visual": "pnpm loki test",
     "prepare": "husky install",
     "size": "pnpm run build && size-limit",
-    "tokens:update": "./scripts/figma-synchronise-tokens.sh https://raw.githubusercontent.com/scaleway/design-tokens/main/tokens.json && pnpm run format"
+    "tokens:update": "./scripts/figma-synchronise-tokens.sh https://raw.githubusercontent.com/scaleway/design-tokens/main/tokens.json && pnpm prettier --write src/theme/tokens"
```
chambo-e, line 55: Need to be removed from ci workflow ?

matthprost: @chambo-e what do you mean?

chambo-e: Nothing, just saw that it's not there anymore :)

chambo-e (approved):

vincentaudebert (approved):

PR #1300, opened 2022-03-22
fix: correct theme type

chambo-e:
## Type

- Bug

### DIFF
--- a/src/theme/index.ts
+++ b/src/theme/index.ts
@@ -71,7 +71,7 @@ const extendTheme = ({
 }: {
   baseTheme?: typeof theme
   extendedTheme: RecursivePartial<typeof theme>
-}) => deepmerge(baseTheme, extendedTheme) as typeof theme
+}) => deepmerge(baseTheme, extendedTheme) as SCWUITheme
 
 const darkTheme = extendTheme({
   extendedTheme: dark,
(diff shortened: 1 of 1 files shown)


src/theme/index.ts
```diff
@@ -56,31 +56,33 @@ const theme = {
...
   extendedTheme,
 }: {
-  baseTheme?: typeof theme
-  extendedTheme: RecursivePartial<typeof theme>
-}) => deepmerge(baseTheme, extendedTheme) as typeof theme
+  baseTheme?: SCWUITheme
+  extendedTheme: RecursivePartial<SCWUITheme>
+}) => deepmerge(baseTheme, extendedTheme) as SCWUITheme
+
+const lightTheme: SCWUITheme = theme
```
QuiiBz, line 80: I don't understand why we create this `lightTheme`, can't we export the existing `theme` and cast it as a `SCWUITheme`?

matthprost: I think it is because that way, the exported theme is correctly typed and can be imported without any `theme is of type any` into an outside project

matthprost (approved):

src/theme/index.ts
QuiiBz: Yeah but can't we simply `export theme as SCWUITheme`?

chambo-e: Casting a value will force the type, it's better to set the variable type, not override it

Also SCWUITheme is inferred from theme, meaning we can't do `export theme as SCWUITheme` as we would a circular ref :)

matthprost: @QuiiBz I think the idea is also to make it explicit that we export lightTheme as default theme

QuiiBz: Thanks for the clarification :)

QuiiBz (approved):

PR #1303, opened 2022-03-23
fix(colors): badge neutral to be more dark

matthprost:
## Type

- Enhancement

#### What is expected?

I know it makes no sense, it is only temporary until we passe trough all components.


chambo-e (approved):

vincentaudebert (approved):
Bien cool la diff visual testing quand meme hein 😇

PR #1322, opened 2022-03-31
feat(react): react17 new jsx transform

philibea:
## Type

- Enhancement

#### What is expected?

- New JSX Transformer.
- Improve size bundle
- Using new prerelease of storybook ( can wait a stable version before merge )

1. npx react-codemod update-react-imports

2. nr format


package.json
```diff
@@ -120,17 +120,19 @@
...
+    "@storybook/builder-webpack5": "^6.5.0-alpha.52",
     "@storybook/csf": "canary",
-    "@storybook/manager-webpack5": "6.4.19",
-    "@storybook/react": "6.4.19",
+    "@storybook/manager-webpack5": "^6.5.0-alpha.52",
+    "@storybook/mdx2-csf": "^0.0.1-canary.1.838a6ca.0",
+    "@storybook/react": "6.5.0-alpha.52",
     "@storybook/testing-react": "1.2.4",
-    "@storybook/theming": "6.4.19",
+    "@storybook/theming": "6.5.0-alpha.52",
```
QuiiBz, line 135: I also think we should wait for a stable release

PR #1338, opened 2022-04-05
fix: background overlay and elevated colors

vincentaudebert:
## Type

- Enhancement

#### What is expected?

background elevated color + overlay color on modal

(Describe what you did)

1. Change some bg colors on elements

2. It should not change anything. Just nicer on dark theme mode

### DIFF
--- a/src/components/Modal/index.tsx
+++ b/src/components/Modal/index.tsx
@@ -222,7 +222,7 @@ const StyledDialog = styled(Dialog, {
   ${({ placement }) => MODAL_PLACEMENT[placement]}
   width: ${({ width }) => MODAL_WIDTH[width]}px;
   min-height: ${({ height }) => height};
-  box-shadow: 0 0 12px 18px #151a2d5c;
+  box-shadow: 0 8px 40px 10px #151A2D5C;
   opacity: 1;
   &::before {
     content: '';
--- a/src/components/Modal/index.tsx
+++ b/src/components/Modal/index.tsx
@@ -1,6 +1,5 @@
 import { css } from '@emotion/react'
 import styled from '@emotion/styled'
-import { transparentize } from 'polished'
 import PropTypes from 'prop-types'
 import React, {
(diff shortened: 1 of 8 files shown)


src/components/Modal/index.tsx
```diff
@@ -215,16 +213,16 @@ const StyledDialog = styled(Dialog, {
...
   position: relative;
   border-radius: ${({ bordered }) => (bordered ? 4 : 0)}px;
   border: 0;
   padding: 32px;
   ${({ placement }) => MODAL_PLACEMENT[placement]}
   width: ${({ width }) => MODAL_WIDTH[width]}px;
   min-height: ${({ height }) => height};
-  box-shadow: 0 0 12px 18px
-    ${({ theme }) => transparentize(0.8, theme.colors.neutral.backgroundStrong)};
+  box-shadow: 0 0 12px 18px #151a2d5c;
```
matthprost, line 225: From what I see in shadows tokens it should be more like:

```suggestion
  box-shadow: 0 8px 40px 10px #151A2D5C;
```

It should make a more beautiful shadow in light theme

matthprost (changes requested):

.github/workflows/ci.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         cache: 'pnpm'
     - run: pnpm install
     - run: pnpm run test:unit:coverage
-    - uses: codecov/codecov-action@v2.1.0
```
matthprost, line 52: Is due to broken rebase that we see that ?

matthprost (changes requested):

matthprost (approved):

.github/workflows/ci.yml
chambo-e: yes :)

chambo-e (approved):

philibea (approved):

PR #1369, opened 2022-04-11
feat(theme): typography synchronized with figma

matthprost:
## Type

- Feature

#### What is expected?

To synchronise font tokens with figma and being able to use theme trough theme.

1. I updated script to first git units, then lineHeight, fontSize to finally generate typography with all those parameters.


chambo-e (approved):
Huge fan GG ❤️

src/theme/tokens/dark.ts
```diff
@@ -231,5 +231,107 @@ export default {
     popover: '0px 4px 32px 8px #0000008C',
     tooltip: '0px 4px 32px 8px #00000066',
   },
+  typography: {
+    Body: {
```
chambo-e, line 235: Just a small comment, could we have the variant names as camelCase?

QuiiBz (approved):

PR #1374, opened 2022-04-12
feat: new bullet component

Exayy:
## Type

- Feature

### Summarise concisely:

Add a new component : Bullet


matthprost (changes requested):

src/components/Bullet/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+  tooltipBaseId,
+}) => (
+  <Tooltip baseId={tooltipBaseId} text={tooltip}>
+    <StyledContainer variant={variant} size={size} className={className}>
+      {icon ? <Icon name={icon} size={iconSizes[size]} /> : text}
+    </StyledContainer>
+  </Tooltip>
+)
+
+const customXorPropType = (
```
chambo-e, line 105: What if we put icon and text as optional ?

src/components/Bullet/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+  variant = 'default',
+  size = 'medium',
+  icon,
+  text,
+  tooltip,
+  tooltipBaseId,
+}) => (
+  <Tooltip baseId={tooltipBaseId} text={tooltip}>
+    <StyledContainer variant={variant} size={size} className={className}>
+      {icon ? <Icon name={icon} size={iconSizes[size]} /> : text}
```
chambo-e, line 100: Can we use percentage as Icon size ?

src/components/Bullet/index.tsx
Exayy: it works fine but you have no Proptypes warning if both aren't specified

src/components/Bullet/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,99 @@
...
+    storyDescription: 'Add a `tooltip` using tooltip property',
+  },
+}
+Tooltip.decorators = [
+  () => (
+    <div style={{ display: 'flex' }}>
+      <Bullet text="1" tooltip="tooltip text" />
+    </div>
+  ),
+]
```
matthprost, line 99:
```suggestion
export const Tooltip = Template.bind({})
Tooltip.parameters = {
  docs: {
    storyDescription: 'Add a `tooltip` using tooltip property',
  },
}
Tooltip.args = {
  text: '1',
  tooltip: 'tooltip text'
}
```

matthprost, line 99:
```suggestion
export const Tooltip = Template.bind({})
Tooltip.parameters = {
  docs: {
    storyDescription: 'Add a `tooltip` using tooltip property',
  },
}
Tooltip.args = {
  text: '1',
  tooltip: 'tooltip text'
}
```

matthprost (changes requested):

src/components/Bullet/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    width: 32px;
+    height: 32px;
+    font-size: 16px;
+  `,
+  small: css`
+    width: 24px;
+    height: 24px;
+    font-size: 14px;
+  `,
+} as const
```
matthprost, line 54:
```suggestion
const sizes = {
  medium: `
    width: 32px;
    height: 32px;
    font-size: 16px;
  `,
  small: `
    width: 24px;
    height: 24px;
    font-size: 14px;
  `,
} as const
```

doesn't it work this way ?

matthprost (changes requested):

src/components/Bullet/__stories__/index.stories.tsx
Exayy: without a Flexbox around tooltip, the tooltip position is bad (Tooltip component issue not related with Bullet)

src/components/Bullet/index.tsx
chambo-e: Yeah, don't care about proptypes :D
Less logic we add the better

matthprost (approved):

matthprost:
All good to me, well done 🎉

vincentaudebert (approved):

PR #1386, opened 2022-04-14
fix: stepper changes

vincentaudebert:
## Type

- Enhancement

#### What is expected?

Optimises a bit Stepper

### DIFF
--- a/src/components/Stepper/index.tsx
+++ b/src/components/Stepper/index.tsx
@@ -8,7 +8,6 @@ import React, {
   KeyboardEventHandler,
   MutableRefObject,
   ReactNode,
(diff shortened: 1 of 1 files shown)


vincentaudebert:
Needs to be tested on shire to check form.change() issue

vincentaudebert:
@matthprost @chambo-e couldn't test on console but it may have fixed form.change issue. If not it's not changing anything to existing so we can check/fix later when we use this version into codebase. (reminder: pnpm/yalc not working)

matthprost (approved):

chambo-e (approved):

PR #1397, opened 2022-04-19
fix: removed and fixed notice proptypes

matthprost:
## Type

- Bug

#### What is expected?

Removed proptypes and fixed warnings.

### DIFF
--- a/src/components/Notice/index.tsx
+++ b/src/components/Notice/index.tsx
@@ -1,5 +1,4 @@
 import styled from '@emotion/styled'
-import PropTypes from 'prop-types'
 import React, { ReactNode } from 'react'
 import Box, { BoxProps } from '../Box'
 import Icon from '../Icon'
@@ -11,8 +10,9 @@ const Container = styled(Box)`
   display: flex;
   align-items: center;
 `
+
 type NoticeProps = {
-  children: ReactNode
+  children: ReactNode | string
 } & BoxProps
 
 const Notice = ({ children, ...props }: NoticeProps) => (
@@ -23,8 +23,5 @@ const Notice = ({ children, ...props }: NoticeProps) => (
     ) : null}
   </Container>
 )
-Notice.propTypes = {
-  children: PropTypes.node.isRequired,
-}
 
 export default Notice
--- a/src/components/Notice/index.tsx
+++ b/src/components/Notice/index.tsx
@@ -12,11 +12,12 @@ const Container = styled(Box)`
 `
 
 type NoticeProps = {
-  children: ReactNode | string
+  children: ReactNode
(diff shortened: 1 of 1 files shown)


src/components/Notice/index.tsx
```diff
@@ -11,8 +10,9 @@ const Container = styled(Box)`
   display: flex;
   align-items: center;
 `
+
 type NoticeProps = {
-  children: ReactNode
+  children: ReactNode | string
```
chambo-e, line 15: ?

src/components/Notice/index.tsx
```diff
@@ -11,8 +10,9 @@ const Container = styled(Box)`
   display: flex;
   align-items: center;
 `
+
 type NoticeProps = {
```
chambo-e, line 14: Could you add className also pls ? :)

chambo-e (changes requested):

src/components/Notice/index.tsx
matthprost: my IDE who's displaying wrong warnings sorry I'm removing it

matthprost: my IDE who's displaying wrong warnings sorry I'm removing it

src/components/Notice/index.tsx
```diff
@@ -11,8 +10,10 @@ const Container = styled(Box)`
...
   children: ReactNode
+  className: string
 } & BoxProps
 
 const Notice = ({ children, ...props }: NoticeProps) => (
```
chambo-e, line 19: className need to be passed down here too :)

chambo-e (approved):

src/components/Notice/index.tsx
```diff
@@ -11,20 +10,19 @@ const Container = styled(Box)`
...
+
 type NoticeProps = {
   children: ReactNode
+  className?: string
 } & BoxProps
 
-const Notice = ({ children, ...props }: NoticeProps) => (
-  <Container {...props}>
+const Notice = ({ children, className, ...props }: NoticeProps) => (
+  <Container className={className} {...props}>
```
Exayy, line 20: any reason using for spread everything except className ?

Change not required it was just a question

matthprost: @chambo-e I let you answer haha

matthprost: @chambo-e I let you answer haha

chambo-e: emotion require a type with className to be able to style them, the rest is xstyled props which we will remove soon :)

vincentaudebert (approved):

PR #1399, opened 2022-04-20
feat: bullet list becomes step title

vincentaudebert:
## Type

- Feature

#### What is expected?

BulletList component is now StepTitle

| url  | Screenshot 2022-04-20 at 15 47 11 | Screenshot 2022-04-20 at 15 47 25 |


src/components/StepTitle/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    disabled={disabled}
+  >
+    <Bullet
+      text={bulletText}
+      size={size}
+      variant={disabled ? 'disabled' : 'default'}
+    />
+    <StyledDiv size={size}>{children}</StyledDiv>
+  </Step>
+)
```
matthprost, line 54: I'm pretty sure we will need to customize bullet icon and variant too. For example to do a successful stepTitle.

Maybe we should add it into props too, what do you think ?

matthprost (changes requested):

src/components/StepTitle/index.tsx
vincentaudebert: Done & RTR @matthprost

src/components/StepTitle/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  children: ReactNode
+} & ContentProps
+
+const StepTitle = ({
+  bulletText,
+  bulletIcon,
+  bulletVariant,
+  children,
+  size = 'medium',
+  disabled = false,
```
matthprost, line 45: I still think we could call prop `variant` instead of `bulletVariant` and pass it to Bullet component like we do for size prop. I believe that variant will always influence Bullet and Title

matthprost (changes requested):

src/components/StepTitle/index.tsx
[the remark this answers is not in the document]
vincentaudebert: Done 👌

chambo-e (changes requested):
Where did the list go ? 😄

vincentaudebert:
@chambo-e it has to be done by hand with a loop around StepTitle component. 

If we set a variant, size etc. on a list wrapper inside this component, we will have all the children using the same variant/size/diabled etc... when we need to be able to customize each step as we want.

src/components/StepTitle/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  disabled?: boolean
+  children: ReactNode
+} & ContentProps
+
+const StepTitle = ({
+  bulletText,
+  bulletIcon,
+  variant,
+  children,
+  size = 'medium',
```
matthprost, line 44: Size is never undefined as when it is, you define it so for me it should be more like this:

```suggestion
export type Sizes = 'small' | 'medium'

type ContentProps =
  | { bulletIcon: ComponentProps<typeof Bullet>['icon']; bulletText?: never }
  | { bulletIcon?: never; bulletText: string }

type StepTitleProps = {
  variant?: ComponentProps<typeof Bullet>['variant']
  size?: Sizes
  disabled?: boolean
  children: ReactNode
} & ContentProps

const StepTitle = ({
  bulletText,
  bulletIcon,
  variant,
  children,
  size = 'medium',
```
```suggestion
export type Sizes = 'small' | 'medium' | undefined

type ContentProps =
  | { bulletIcon: ComponentProps<typeof Bullet>['icon']; bulletText?: never }
  | { bulletIcon?: never; bulletText: string }

type StepTitleProps = {
  variant?: ComponentProps<typeof Bullet>['variant']
  size?: Sizes
  disabled?: boolean
  children: ReactNode
} & ContentProps

const StepTitle = ({
  bulletText,
  bulletIcon,
  variant,
  children,
  size = 'medium',
```

matthprost (changes requested):

chambo-e:
> @chambo-e it has to be done by hand with a loop around StepTitle component. (see List example in storybook)
> 
> If we set a variant, size etc. on a list wrapper inside this component, we will have all the children using the same variant/size/disabled etc... when we need to be able to customize each step as we want.

Alright so I think there is a bigger issue on this component. This does not replace BulletList, as this is neither a list nor a bullet :)

To replace current BulletList we would still need  a dedicated component which would be close to our current BulletList

StepTitle, at best, replace 
```tsx
        <Chip>{`${index + 1}`}</Chip>
        <StyledDiv>{child}</StyledDiv>
``` 
in BulletList

QuiiBz:
The disabled state doesn't really look like disabled, maybe with a grayer color?

Screenshot 2022-05-11 at 09 35 23

src/components/StepList/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  padding-left: ${({ theme }) => theme.space['2']};
+  margin-top: ${({ theme }) => theme.space['3']};
+  text-align: left;
+  font-size: 16px;
+`
+
+export const Step = styled('li', {
+  shouldForwardProp: prop => !['disabled'].includes(prop.toString()),
+})<{ disabled: boolean }>`
+  margin-bottom: 24px;
```
QuiiBz, line 17: Margin can be from theme

QuiiBz (approved):

src/components/StepList/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+Item.propTypes = {
+  bulletIcon: PropTypes.string,
+  bulletText: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  disabled: PropTypes.bool,
+  size: PropTypes.string,
+  variant: PropTypes.string,
+}
+
```
matthprost, line 82: Can be removed as we go full typescript

matthprost (changes requested):

src/components/StepList/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  justify-content: center;
+  color: ${({ theme, disabled }) =>
+    disabled ? theme.colors.neutral.textWeak : theme.colors.neutral.textStrong};
+`
+
+const StyledDiv = styled('div', {
+  shouldForwardProp: prop => !['size'].includes(prop.toString()),
+})<{ size: Sizes }>`
+  flex: 1;
+  margin-left: ${({ theme, size }) =>
```
chambo-e, line 29: Could we have a margin relative to font-size ?

vincentaudebert: @chambo-e not sure to understand, that's more or less the case? 

```
margin-left: ${({ theme, size }) =>
    size === 'medium' ? theme.space['2'] : theme.space['1']};
  font-size: ${({ size }) => (size === 'medium' ? '24px' : '16px')};
```

vincentaudebert:
@QuiiBz done

src/components/StepList/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    disabled
+      ? theme.colors.neutral.textDisabled
+      : theme.colors.neutral.textStrong};
+`
+
+const StyledDiv = styled('div', {
+  shouldForwardProp: prop => !['size'].includes(prop.toString()),
+})<{ size: Sizes }>`
+  flex: 1;
+  margin-left: 1rem;
```
matthprost, line 29: Are we using rem now ?

matthprost (approved):

src/components/StepList/index.tsx
vincentaudebert: @matthprost Linked to @chambo-e feedback above as font-size is modified depending on "size" props it avoids having twice the same piece of code

PR #1429, opened 2022-05-02
fix(badge): default size if wrong value is given

matthprost:
## Type

- Enhancement

#### What is expected?

If prop given doesn't exists we still set as medium as default size to avoid strange behavior.

### DIFF
--- a/src/components/Badge/index.tsx
+++ b/src/components/Badge/index.tsx
@@ -131,7 +131,7 @@ const Badge: FunctionComponent<BadgeProps> = ({
     [prominence, theme],
   )
 
-  const sizeValue = SIZES[size]
+  const sizeValue = SIZES[size] || SIZES.medium
 
   return (
     <StyledBox


src/components/Badge/index.tsx
```diff
@@ -131,7 +131,7 @@ const Badge: FunctionComponent<BadgeProps> = ({
     [prominence, theme],
   )
 
-  const sizeValue = SIZES[size]
+  const sizeValue = SIZES[size] || SIZES.medium
```
chambo-e, line 134:
```suggestion
  const sizeValue = SIZES?.[size] ?? SIZES.medium
```

?

chambo-e: Can we also ad a test ?

matthprost: more like `SIZES[size] ?? SIZES.medium` as `SIZES` always exists, I will add testing :)

matthprost: more like `SIZES[size] ?? SIZES.medium` as `SIZES` always exists, I will add testing :)

PR #1431, opened 2022-05-02
feat: 🎸 status component

Exayy:
## Summary

Add a "Status" component which will replace Status Indicator

- It will replace `StatusIndicator` but if we let them both, we simplify our migration
- `babel-plugin-named-exports-order` allow exported story to be in the same order than in file, avoiding strange behavior with `Playground` story


babel.config.json
```diff
@@ -14,5 +14,6 @@
         "labelFormat": "[filename]--[local]"
       }
     ]
-  ]
+  ],
+  "plugins": ["babel-plugin-named-exports-order"]
```
chambo-e, line 18: Can we put this only in storybook config ?

src/components/Status/__tests__/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import Status, { statusVariants } from '..'
+import { shouldMatchEmotionSnapshot } from '../../../helpers/jestHelpers'
+
+describe('Status', () => {
+  statusVariants.forEach(variant => {
+    test(`render ${variant}`, () =>
```
chambo-e, line 7: test.each maybe ?

src/components/Status/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+    `,
+  neutral: ({ theme }: { theme: Theme }) => `
+      background-color: ${theme.colors.neutral.backgroundStronger};
+    `,
+  success: ({ theme }: { theme: Theme }) => `
+      background-color: ${theme.colors.success.backgroundStrong};
+    `,
+  warning: ({ theme }: { theme: Theme }) => `
+      background-color: ${theme.colors.warning.backgroundStrong};
+      color: ${theme.colors.primary.text};
```
chambo-e, line 22: color ?

chambo-e (commented):
Shouldn't we delete StatusIndicator too?

src/components/Status/index.tsx
Exayy: done

src/components/Status/__tests__/index.tsx
Exayy: done 👌

babel.config.json
Exayy: done

Exayy:
> Shouldn't we delete StatusIndicator too?

We should probably let this component until migration isn't done

package.json
```diff
@@ -185,6 +185,7 @@
     "@types/intl-tel-input": "17.0.4",
     "@types/react-datepicker": "4.4.1",
     "@xstyled/emotion": "2.5.0",
+    "babel-plugin-named-exports-order": "^0.0.2",
```
chambo-e, line 188: devDeps maybe ?

QuiiBz: We can also move the above `@types` packages to devdeps

QuiiBz (approved):

package.json
chambo-e: Nope, @types/intl-tel-input and @types/react-datepicker are exported in the definition bundle :)

Exayy: Done

src/components/Status/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  className,
+  tooltip,
+  variant,
+}: StatusProps): JSX.Element => (
+  <Tooltip text={tooltip}>
+    <StyledCircle
+      className={className}
+      css={animated && cssAnimation}
+      variant={variant}
+    />
```
matthprost, line 60: Can't we pass animated prop directly to `StyledCircle` and condition with ternary `animation: ${flash} linear 1s infinite;` there directly ?

matthprost (changes requested):

src/components/Status/index.tsx
Exayy: done

matthprost (approved):

PR #1446, opened 2022-05-06
feat(checkbox): refactoring

matthprost:
## Type

- Bug
- Feature
- Enhancement
- Documentation
- Migration
- Refactor

#### What is expected?

(Describe what you did)

1.

2.

3.


DorianMaliszewski (approved):

src/components/Checkbox/index.tsx
```diff
@@ -139,62 +185,33 @@ const Checkbox = ({
...
+        {children}
       </StyledCheckBoxContainer>
-      <Expandable height={56} overflow="hidden" opened={!!error}>
-        <StyledError>{error}</StyledError>
-      </Expandable>
-    </Box>
+      {error ? (
+        <StyledError
+          id={`${computedName}-id`}
+          as="p"
```
chambo-e, line 206: bodyB is already a p

src/components/Checkbox/index.tsx
```diff
@@ -12,123 +12,169 @@ import {
...
-  const color = useMemo(() => {
-    if (disabled) return 'gray100'
-    if (valid === false || !!error) return 'warning'
-    if (valid === true) return 'success'
-    if (checkbox.state) return 'primary'
 
-    return 'gray300'
-  }, [disabled, valid, checkbox.state, error])
+  const computedName = useMemo(() => {
+    if (!name) return getUUID('checkbox')
```
chambo-e, line 128: useUUID

src/components/Checkbox/index.tsx
```diff
@@ -12,123 +12,169 @@ import {
...
+  shouldForwardProp: prop => !['size'].includes(prop.toString()),
+})`
   opacity: 0.01;
   width: ${({ size }) => size}px;
   height: ${({ size }) => size}px;
   position: absolute;
   cursor: pointer;
-  margin-right: ${({ hasChildren }) => (hasChildren ? '10px' : 0)};
-  padding: 2px;
   pointer-events: auto;
```
chambo-e, line 38: useless I think ?

chambo-e (changes requested):
The violet checkboxes on List is intended ? 

Screenshot 2022-05-11 at 09 55 51

src/components/Checkbox/index.tsx
matthprost: @chambo-e I want to be able to set a full name into my checkbox input without having generated ID next to it if name exists. Only when name prop is empty I generate a name with an ID. 

`useUUID` can't be used conditionally as it uses a `useMemo` into it. In order to be able to conditionally set this variable I had to use `getUUID`.

matthprost:
> The violet checkboxes on List is intended ?
> 
> Screenshot 2022-05-11 at 09 55 51

That's in fact a mistake, I forgot to re-generate visual testing. Those checkboxes at not supposed to be violet.

src/components/Checkbox/index.tsx
matthprost: Yes you're right :)

vincentaudebert (approved):

PR #1472, opened 2022-05-11
feat: explore loki interactions

vincentaudebert:
## Type

- Enhancement

#### What is expected?

Interactions before loki screenshots

chrome_laptop_Components_Navigation_Menu_Advanced_WIP


vincentaudebert:
@matthprost @Exayy your thoughts? :/

matthprost:
@vincentaudebert I think its a good solution, seems to work just fine. Do you have any doubts about it ?

matthprost (approved):

vincentaudebert:
Yes @matthprost few problems spotted:
- I can't seem to be able to have a proper "hover" event. I have to click on the element. It is not ideal. (especially for buttons or links etc.)
- I can't screenshot all the elements at once, for instance on Tooltip when we have multiple tooltips on the same story I can't hover all of them and therefore screenshot would be incomplete? For these cases, maybe we should just screenshot the "default" story... better than nothing :/

matthprost:
@vincentaudebert lets go with this we will improve in another iteration :)

PR #1482, opened 2022-05-17
docs: deprecated perameter and migration documentation

matthprost:
## Type

- Documentation

#### What is expected?

To be able to mark a component deprecated trough storybook and trough JSDoc.

1. Added deprecated parameter into storybook layout .

2. Added documentation about migration and how to do it.

Screenshot 2022-05-17 at 16 23 10
Screenshot 2022-05-17 at 16 54 26


.storybook/components/DocsContainer.tsx
```diff
@@ -2,30 +2,46 @@ import React from 'react'
 import { DocsContainer as BaseContainer } from '@storybook/addon-docs/blocks'
```
chambo-e, line 2: Could we fix this ts error at the same time ?

chambo-e (changes requested):
Could we mark Box, FlexBox and Grid as deprecated ?

src/__stories__/Migration.stories.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+<Meta title="Migration"/>
+
+# Migration
+
+<Description>
+  How to make a component deprecated and what is the process of migration? That's what we will see in this documentation.
+</Description>
+
+## I want to migrate my component with new name
```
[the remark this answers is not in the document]
matthprost: @chambo-e that way of migration is only for component that are changing name + changing some props/style. For component that keeps same name I will add way of doing it later on.

Maybe `My component is changing name plus others properties and we can't have a 1:1 easy migration path` ?

src/components/Button/__stories__/index.stories.tsx
```diff
@@ -5,6 +5,11 @@ import { icons } from '../../Icon'
 
 export default {
   component: Button,
+  parameters: {
+    deprecated: true,
+    deprecatedReason:
+      'This component is deprecated please do not use it any more.',
+  },
   title: 'Components/Button/Button',
```
matthprost, line 13: > **Warning** 
> To remove ASAP

src/__stories__/Migration.stories.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+  ```typescript
+    /**
+     * @deprecated Has been replaced by component BetterButton, please this one instead.
+     */
+    const Button = props => {
+        <div>
+            This is supposed to be a button
+        </div>
+    }
```
QuiiBz, line 47:
```suggestion
    const Button = props => (
        <div>
            This is supposed to be a button
        </div>
    )
```

QuiiBz (approved):

matthprost:
@chambo-e RTR

vincentaudebert (approved):

PR #1483, opened 2022-05-17
fix: allow classname on steplist

vincentaudebert:
## Type

- Bug

#### What is expected?

Pass through classname

### DIFF
--- a/src/components/StepList/index.tsx
+++ b/src/components/StepList/index.tsx
@@ -74,7 +74,7 @@ const Item = ({
 
 type StepListProps = {
   children: ReactNode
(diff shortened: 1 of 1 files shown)


src/components/StepList/index.tsx
```diff
@@ -72,9 +74,12 @@ const Item = ({
 
 type StepListProps = {
   children: ReactNode
+  className: string
```
matthprost, line 77:
```suggestion
  className?: string
```

matthprost (approved):

philibea (approved):

PR #1485, opened 2022-05-18
fix(modal): disclosure type

philibea:
## Type

- Enhancement

### Summarise concisely:

Allow user of modal to not type Disclosure by themself.


matthprost (approved):

chambo-e (commented):
The user can extract prop type with ComponentProps no ?

PR #1504, opened 2022-05-24
fix: update Tag component

Lawndlwd:
## Type

- Refactor

#### What is expected?

Code refactor of component Tag .

Screenshot 2022-05-24 at 17 33 57


src/components/TagsPoplist/index.tsx
```diff
@@ -60,6 +60,15 @@ type TagsPoplistProps = {
...
 
+const TagContainer = styled.div<{ index: number; visibleTagsCount: number }>`
+  margin-right: ${({ index, visibleTagsCount }) =>
+    index + 1 !== visibleTagsCount ? 1 : 0};
+`
+
+const StyledTag = styled(Tag)`
+  margin-right: ${({ theme }) => theme.space['1']};
+`
+
```
chambo-e, line 71: Instead of always playing with margin I think we could use a gap somewhere :)

src/components/Tag/index.tsx
```diff
@@ -1,140 +1,147 @@
...
+  /**
+   * Badge should display an aria-label if the status is not neutral or primary
+   */
+  const ariaLabel = useMemo(
+    () =>
+      ['neutral', 'primary'].some(baseVariant => baseVariant === variant)
+        ? undefined
+        : variant,
+    [variant],
+  )
```
chambo-e, line 112: Why ?

src/components/Tag/index.tsx
```diff
@@ -1,140 +1,147 @@
...
-  variant: TagVariant
-} & BoxProps
-
-const StyledContainer = styled(Box, {
-  shouldForwardProp: props =>
-    !['disabled', 'variant'].includes(props.toString()),
-})<StyledContainerProps>`
-  ${({ disabled }) => disabled && disabledStyles}
-  border-radius: 4px;
+const StyledBox = styled('div', {
```
chambo-e, line 10: Could we use another name than Box, for the sake of the current migration ? :)

src/components/Tag/index.tsx
```diff
@@ -1,140 +1,147 @@
...
   disabled = false,
-  variant = 'base',
-  ...props
-}: TagProps) => (
-  <StyledContainer {...props} disabled={disabled} variant={variant}>
-    <StyledText aria-disabled={disabled} css={textStyle}>
-      {children}
-    </StyledText>
+  variant = 'neutral',
+  className,
```
chambo-e, line 99: no issues to dropping xsytled props ?

src/components/Tags/index.tsx
```diff
@@ -232,7 +232,7 @@ const Tags = ({
     >
       {tagsState.map(tag => (
         <Tag
-          variant={variant === 'no-border' ? 'base' : variant}
```
chambo-e, line 235: So we loose variant on Tags ?

chambo-e (changes requested):

src/components/Tag/index.tsx
Lawndlwd: I tested it with yalc and there is some `tag` in shire effected, so I created an MR to migrate

src/components/Tags/index.tsx
Lawndlwd: Probably we will have component review for `Tags`,  I saw this with @matthprost, and It's fine to have one variant for tag.
checking on shire, we don't use any variant for tags just `no-border` in a couple of them

src/components/Tag/index.tsx
Lawndlwd: true

src/components/Tag/index.tsx
Lawndlwd: for accessibility  reasons

src/components/TagsPoplist/index.tsx
Lawndlwd: done

src/components/Tag/index.tsx
chambo-e: Alright, so I see some issues, this is not a role status, so I don't think it needs a label, also a label with just the name of the variant is kinda useless a11y wise :)

Lawndlwd: yeah, I'm not expert in it to be honest, so do we remove the aria-label ?

src/components/Tag/index.tsx
Lawndlwd: done

matthprost:
Not to forget @scaleway/design here too :)

chambo-e:
Screenshot 2022-05-30 at 15 03 30

I feel that there is a margin missing here :)

src/components/TagsPoplist/index.tsx
```diff
@@ -60,6 +61,11 @@ type TagsPoplistProps = {
   multiline?: boolean
 }
 
+const TagContainer = styled.div<{ index: number; visibleTagsCount: number }>`
+  margin-right: ${({ index, visibleTagsCount }) =>
+    index + 1 !== visibleTagsCount ? 1 : 0};
+`
```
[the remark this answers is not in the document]
Lawndlwd: nope, because we don't have the parameters `index`, `visibleTagsCount` in the parent what we could do is add div and the style on it which i think its  so over, what do you think?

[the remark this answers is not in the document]
chambo-e: Well if you put a gap on the parent you won't need to have index and visibleTagsCount no ? 

You already have StyledTagContainer as parent

src/components/Tag/index.tsx
```diff
@@ -1,140 +1,133 @@
...
+}: TagProps) => {
+  const theme = useTheme()
+
+  const generatedStyles = useMemo(() => generateStyles({ theme }), [theme])
```
chambo-e, line 103: Why not generate them in the class ? This would avoid passing theme around

chambo-e (changes requested):

NicolasGuyFG (changes requested):
1 - On all variant except the neutral one, add a border of the same color of the bg. (because they don't have the same size)
2 - On the disabled version, also disable the hover comportment on the close icon.

matthprost (changes requested):
https://user-images.githubusercontent.com/15812968/171373818-61e2108f-c795-480e-a68e-89fe13a5c815.mov

There is a hover on close button

src/components/Tag/index.tsx
```diff
@@ -1,115 +1,103 @@
...
-type TagVariant = keyof typeof variantsContainer
-
-const variantStyles = ({
-  variant,
-  ...props
-}: {
-  variant: TagVariant
-  theme: Theme
-}) => variantsContainer[variant]?.(props)
+const StyledContainer = styled('span', {
```
matthprost, line 9: One last thing, is there a reason to put a `span` instead of a `div`? 

From what I see we were using a `div` before so this change might be breaking at some places 😛

matthprost (changes requested):

src/components/TagsPoplist/index.tsx
```diff
@@ -35,10 +35,8 @@ const StyledTagContainer = styled.div<{ multiline?: boolean }>`
   display: flex;
   align-items: center;
   color: ${({ theme }) => theme.colors.neutral.text};
-  ${({ multiline, theme }) =>
-    multiline &&
-    `flex-wrap: wrap;
-  > * { margin-bottom: ${theme.space['1']}}`}
+  gap: ${({ theme }) => theme.space['1']};
+  ${({ multiline }) => multiline && `flex-wrap: wrap;`};
```
matthprost, line 39:
```suggestion
  ${({ multiline }) => multiline && 'flex-wrap: wrap'};
```

src/components/Tag/__stories__/index.stories.tsx
```diff
@@ -1,33 +1,109 @@
...
-const Template: Story<ComponentProps<typeof Tag>> = args => <Tag {...args} />
+export const Loading = Template.bind({})
+Loading.parameters = {
+  docs: {
+    description: {
+      story:
+        'Loading for async operation or other use cases, You can define it by passing `isLoading` property to true.',
+    },
+  },
+}
```
matthprost, line 93: For Icons, onClose and Loading story you don't need to display the property for each variant and so you don't need to use `Icons.decorators`. Instead use `Icons.args` it will make story shorter and more clear to understand & read.

matthprost (changes requested):

src/components/Tag/__stories__/index.stories.tsx
```diff
@@ -1,33 +1,83 @@
...
+        '`onClose` property to pass function trigger on click on close button.',
     },
   },
-  title: 'Components/Data Display/Tag',
-} as Meta
+}
+onClose.args = {
+  icon: 'check',
+  onClose: () => {},
+}
```
matthprost, line 63:
```suggestion
onClose.args = {
  onClose: () => {},
}
```

Lets just show the prop linked to title

src/components/Tag/__stories__/index.stories.tsx
```diff
@@ -1,33 +1,83 @@
...
+    description: {
+      story:
+        'Loading for async operation or other use cases, You can define it by passing `isLoading` property to true.',
+    },
+  },
+}
+Loading.args = {
+  icon: 'check',
+  isLoading: true,
+}
```
matthprost, line 78:
```suggestion
Loading.args = {
  isLoading: true,
}
```

src/components/Tag/index.tsx
Lawndlwd: this done

src/components/Tag/index.tsx
Lawndlwd: i passed them directly to the styled  so done

Lawndlwd:
> Screenshot 2022-06-01 at 15 16 58
> 
> Design: Screenshot 2022-06-01 at 15 16 54
> 
> Isn't there is a border missing in disabled ?
> 
> That's my last review, I validate in advanced, great job 🚀



yup, I see, but it strange to use `borderWeak` in place of `borderDisabled` as the component in disabled state

NicolasGuyFG:
> 

There wasn't ! There is now !

NicolasGuyFG:
I updated the Disabled Tag here : https://www.figma.com/file/XHoO4V9LOFEFkAZsocYJ83/?node-id=2431%3A111981

src/components/Tag/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -2,129 +2,205 @@
...
+  margin-top: 0;
+  font-size: 12px;
+  font-weight: 500;
+  color: inherit;
+}
+
+<span
+    class="cache-1q42p0o-StyledContainer e1vt8sjg2"
+  >
+    <div
```
chambo-e, line 198: div in span not good :)

src/components/Tag/index.tsx
```diff
@@ -1,115 +1,103 @@
...
-  height: 32px;
-`}
 `
 
 type TagProps = {
-  children: ReactNode
-  disabled?: boolean
   isLoading?: boolean
   onClose?: MouseEventHandler<HTMLButtonElement>
   textStyle?: JSX.IntrinsicAttributes['css']
```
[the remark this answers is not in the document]
Lawndlwd: what you mean by complex children ?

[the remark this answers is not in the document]
Lawndlwd: we can't pass anything else than text here, which is styled depending on variant.

for `TextStyle` It's something we use in shire to add `maxwidth` and `overflow hidden` with tooltip to avoid very big text content.

what we could do is use Tooltip and max width in the tag it self i will see with design team

src/components/Tag/__tests__/__snapshots__/index.tsx.snap
Lawndlwd: done

src/components/Tag/index.tsx
```diff
@@ -1,115 +1,103 @@
...
+  >
+    {icon ? <Icon name={icon} size={16} /> : null}
+    <StyledTypography aria-disabled={disabled} css={textStyle}>
       {children}
-    </StyledText>
+    </StyledTypography>
 
-    {onClose && (
+    {onClose || isLoading ? (
       <StyledTouchable
```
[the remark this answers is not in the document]
Lawndlwd: done

chambo-e (approved):

PR #1545, opened 2022-06-04
fix: radioborderedbox spacing

matthprost:
## Type

- Bug

#### What is expected?

Error component is always displayed even when there is no error which leads to a supplement useless spacing.

| RadioBorderedBox  | Screenshot 2022-06-04 at 09 31 04 |  |


src/components/RadioBorderedBox/__stories__/index.stories.tsx
```diff
@@ -132,7 +132,7 @@ Badge.decorators = [
   () => (
     <ControlValue value="badge-1">
       {({ value, onChange }) => (
-        <div style={{ display: 'flex', flexDirection: 'column', gap: 8 }}>
+        <div style={{ display: 'flex', flexDirection: 'column', gap: 16 }}>
```
vincentaudebert, line 135: should we use theme here?

chambo-e (approved):

src/components/RadioBorderedBox/__stories__/index.stories.tsx
matthprost: @vincentaudebert doesn't really apply for stories yet

PR #1550, opened 2022-06-06
fix: accessibility compatibility button & checkbox

Lawndlwd:
## Type

- Feature
- Refactor

#### What is expected?

In order to continue accessibility compatibility, we should re-check Button and Checkbox to make them accessible.

I've run a11y tests on those components and error went out of it, then solve them.

To test run ` pnpm test:a11y `

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -357,6 +357,7 @@ type ButtonProps = Omit<StyledButtonProps, 'variant' | 'size' | 'download'> & {
   innerRef: Ref<Element>
   size?: ButtonSize
   download?: boolean | string
+  ariaLabel: string
 }
 
 const FwdButton = ({
@@ -374,6 +375,7 @@ const FwdButton = ({
   to,
   tooltip,
   tooltipBaseId,
+  ariaLabel,
   type: elementType = 'button',
   variant = 'primary',
   ...props
@@ -397,6 +399,8 @@ const FwdButton = ({
         {...props}
         href={href}
         to={to}
+        aria-label={ariaLabel}
+        role="button"
         download={download}
         ref={innerRef}
         as={as}
--- a/src/__tests__/a11y.test.tsx
+++ b/src/__tests__/a11y.test.tsx
@@ -14,6 +14,8 @@ const testedComponents = [
   'Badge',
   'Bullet',
   'Counter',
+  'Button',
+  'Checkbox',
   'MarkDown',
   'Placeholder',
   'Reminder',
@@ -53,7 +55,6 @@ const searchFileFromDir = (startPath: string, filter: string) => {
     }
   }
 }
-
 // Check if a path was given as input argument, if not we check all stories
 if (process.argv[4]) {
   searchFileFromDir(process.argv[4], '.stories.tsx')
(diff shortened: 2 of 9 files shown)


src/components/Button/index.tsx
```diff
@@ -397,6 +399,8 @@ const FwdButton = ({
         {...props}
         href={href}
         to={to}
+        aria-label={ariaLabel}
+        role="button"
```
chambo-e, line 403: A button role button ?

src/components/Button/index.tsx
```diff
@@ -374,6 +375,7 @@ const FwdButton = ({
   to,
   tooltip,
   tooltipBaseId,
+  ariaLabel,
```
chambo-e, line 378: aria-label is only required when there is no explicit text content, ie only a few of our buttons (icon buttons)

chambo-e (changes requested):

src/components/Button/index.tsx
Lawndlwd: It's more clear, thanks for feedback

src/components/Button/index.tsx
```diff
@@ -397,9 +404,11 @@ const FwdButton = ({
         {...props}
         href={href}
         to={to}
+        aria-label={ariaLabel}
         download={download}
         ref={innerRef}
         as={as}
+        role={as === 'a' ? 'button' : undefined}
```
chambo-e, line 411: Well, if it's a link it's not a button 😄 

```suggestion
        role={as === 'a' ? 'link' : undefined}
```

src/components/Button/index.tsx
```diff
@@ -386,6 +386,13 @@ const FwdButton = ({
     return 'button'
   }, [disabled, to, href, download])
 
+  const ariaLabel = useMemo(() => {
+    if (props['aria-label']) return props['aria-label']
+    if (typeof children === 'string') return children
+
+    return ''
```
chambo-e, line 393: No need to specify an aria label if not needed, meaning it depends on the usage

chambo-e (changes requested):

src/components/Button/index.tsx
Lawndlwd: why we do those tickets ? your point is correct, but this mean we need to force use of aria label if text is not specified or so

src/components/Button/index.tsx
```diff
@@ -400,6 +400,7 @@ const FwdButton = ({
         download={download}
         ref={innerRef}
         as={as}
+        role={as === 'a' ? 'button' : undefined}
```
chambo-e, line 403: Well if it's a link it's not a button 😄 

you can remove the role altogether I think

chambo-e (changes requested):

src/components/Button/index.tsx
Lawndlwd: A11y throws an error if the button is a link with href.
the `a` element cant have aria-label

Lawndlwd: so it doesn't need for aria label if its link ?

matthprost (approved):

matthprost:
@Lawndlwd please change your commit and PR title by using `test:` instead of `fix:`

QuiiBz (approved):

PR #1566, opened 2022-06-09
feat(tooltip): new tooltip homemade

matthprost:
## Type

- Feature

### Summarise concisely:

logo-fait-maison-e1506430293518

#### What is expected?

A new home made tooltip to have full control on behavior and not to use dependency.
Using this new tooltip allow us to remove z-index from modals as we use portals only.

1. Made tooltip portal only that is rendered into DOM only when children is hovered avoid pollution in DOM.

2. Removed z-index of modal as we use dynamic portal for tooltip z-index for modal is no more required.

3. Add tiny animation on tooltip playing on opacity.

4. Simplified props but kept old usage to avoid regression and to be retro-compatible.

https://user-images.githubusercontent.com/15812968/173094758-fa4ccec7-8eac-43ce-aa65-fe49614a48b8.mov


src/components/Tooltip/index.tsx
```diff
@@ -1,147 +1,262 @@
...
+  ReactNode,
+  RefObject,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react'
+import { createPortal } from 'react-dom'
+import { getUUID } from '../../utils'
```
philibea, line 12: can we use useId from react now ?

QuiiBz: We are not yet using react 18 here

matthprost: It will be replaced when we will be using react 18 :)

src/components/Tooltip/index.tsx
```diff
@@ -1,147 +1,262 @@
...
-  const { setVisible } = tooltip
-  useEffect(() => setVisible(visible), [setVisible, visible])
+  /**
+   * Once tooltip is visible in the dom we can compute positions and then set it visible on screen.
+   */
+  useEffect(() => {
+    getPositions()
+
+    window.addEventListener('resize', getPositions)
+    window.addEventListener('scroll', getPositions)
```
QuiiBz, line 202: We are leaking memory by not removing those events in a clear callback

matthprost: @QuiiBz damn I forgot you're right!

QuiiBz (approved):

vincentaudebert:
Les tests plantent mais sinon 💯

NicolasGuyFG (changes requested):
Hello !

Is it possible to animate tooltip like the exemple on the video :
https://user-images.githubusercontent.com/105646727/174066113-d6c987de-9bfd-411c-8171-5222004c24d8.mov

For right tooltip, it's coming from the left, for left tooltip, from the right, etc... The tooltip come from the content.

matthprost:
@NicolasGuyFG It should be better now :)

NicolasGuyFG (approved):
Thx for the edit. Perfect now.

PR #1570, opened 2022-06-09
feat: reduce bundle size

matthprost:
## Type

- Optimisation

1. Removed 2 `.png` we have in phone number input to keep the best quality one. I also converted it into `.webp` which reduced size even more.

2. Removed `avatar.svg` in `Avatar` component, the image should be imported into the component and not be present in this library. ⚠️ This is might be breaking change as there is no more image as default value

| size  | Screenshot 2022-06-09 at 14 42 44 | Screenshot 2022-06-16 at 11 10 48 |


QuiiBz:
Flags are a bit buggy:

Screenshot 2022-06-10 at 10 13 16

matthprost:
> Flags are a bit buggy:
> 
> Screenshot 2022-06-10 at 10 13 16

Can you tell me your web browser ? On chrome it works fine for me 
Screenshot 2022-06-10 at 17 16 49

QuiiBz:
> 
> Can you tell me your web browser ? On chrome it works fine for me Screenshot 2022-06-10 at 17 16 49

Using Brave, Chrome, Firefox, and Safari, they all have the issue 🤔

Screenshot 2022-06-10 at 21 33 40

src/utils/xor.ts
```diff
@@ -0,0 +1,10 @@
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never }
```
chambo-e, line 1: into global.d.ts :)

chambo-e (changes requested):

vincentaudebert (approved):

QuiiBz (approved):

PR #1575, opened 2022-06-10
refactor: drop dot component

DorianMaliszewski:
## Type

- Refactor

#### What is expected?

Drop Dot component

1. Remove the component

2. Replace usage in StatusIndicator and DotStep


DorianMaliszewski:
StatusIndicator should be deleted so ignore coverage testing

src/components/StatusIndicator/index.tsx
```diff
@@ -1,11 +1,25 @@
...
-import { ColorDeprecated as Color } from '../../theme/deprecated/colors'
+import { Color } from '../../theme'
+import { ColorDeprecated } from '../../theme/deprecated/colors'
 import { flash } from '../../utils/animations'
-import { XStyledProps } from '../Box'
-import Dot from '../Dot'
+import Box, { XStyledProps } from '../Box'
 import Tooltip from '../Tooltip'
 
+const Dot = styled(Box, {
```
chambo-e, line 10: Styled div ?

DorianMaliszewski: No we are passing multiple props in our console like margin and padding, because of using it with a prop spreading i think it can have lot of impact on our console @chambo-e :/

chambo-e (approved):

matthprost (approved):

PR #1580, opened 2022-06-13
feat(Switch, SwitchButton): fix a11y issues

QuiiBz:
## Type

- Feature

Fix a11y issues with `Switch` and `SwitchButton`.

- Rename `aria-checked` to `data-checked`
- Not forward not needed props
- Replace `<label />` in `Switch` by a `<div />` and add `aria-label` prop to the `<input />`, since the switch don't have a visible label

### DIFF
--- a/src/components/Switch/index.tsx
+++ b/src/components/Switch/index.tsx
@@ -288,7 +288,7 @@ const Switch = ({
           </SwitchBallContainer>
           <StyledCheckbox
             id={id || name}
-            aria-label={`${name}-label`}
+            aria-label={name}
             checked={checked}
             disabled={disabled}
             name={name}
--- a/src/components/Switch/index.tsx
+++ b/src/components/Switch/index.tsx
@@ -48,12 +48,12 @@ type StyledSwitchProps = VariantProps & {
 
 const SwitchVariantsStyles = {
   primary: (theme: Theme) => css`
-    &[aria-checked='true'] {
+    &[data-checked='true'] {
       color: ${theme.colors.neutral.textStrong};
       background-color: ${theme.colors.primary.backgroundStrong};
     }
 
-    &[aria-checked='true'] {
+    &[data-checked='true'] {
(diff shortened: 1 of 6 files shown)


src/components/Switch/index.tsx
```diff
@@ -287,6 +288,7 @@ const Switch = ({
           </SwitchBallContainer>
           <StyledCheckbox
             id={id || name}
+            aria-label={`${name}-label`}
```
chambo-e, line 291: accessibility wise this is worst than having nothing, we are ticking a11y checks with placeholder data

chambo-e: Could we force either aria-label or label instead ?

chambo-e (changes requested):

src/components/Switch/index.tsx
QuiiBz: Label needs real content, we don't always have something to write as a label. Maybe we should add a new `label` prop and set this `aria-label={label}`? cc @matthprost

matthprost: @QuiiBz Is there an issue of using name and put it into `aria-label={name}` ? If there is then yes I agree you can create a prop `label` and set `aria-label={label}`.

src/components/Switch/index.tsx
```diff
@@ -273,7 +274,7 @@ const Switch = ({
         ) : null}
         <StyledSwitch
           size={size}
-          aria-checked={checked}
+          data-checked={checked}
```
matthprost, line 277: Can't we use `checked={}` directly ?

matthprost (changes requested):

src/components/Switch/index.tsx
QuiiBz: `StyledSwitch` is not an input but a div

matthprost (approved):

PR #1586, opened 2022-06-15
docs(theme): updated theme documentation

matthprost:
## Type

- Documentation

#### What is expected?

Updated theme documentation


src/__stories__/Documentation.stories.mdx
```diff
@@ -1,7 +1,6 @@
 import {
```
chambo-e, line 1: @philibea I seem to remember that we can migrate these file to tsx since the last updates ?

matthprost: @chambo-e I think we said we keep pure documentation in MDX as it is easier to read & write and only stories of components in tsx

chambo-e: Afaik MDX prevents us from fully using react 17 automated runtime, maybe it's supported now

chambo-e: Alright so apparently storybook is now able to have the new react runtime in mdx so I'll close this :)

chambo-e (approved):

QuiiBz (approved):

PR #1589, opened 2022-06-16
test: radio and TextBox accessibility

Lawndlwd:
## Type

- Bug
- Refactor

#### What is expected?

accessibility compatibility on Radio and TextBox

### DIFF
--- a/src/components/TextBox/index.tsx
+++ b/src/components/TextBox/index.tsx
@@ -560,8 +560,8 @@ const TextBox = forwardRef<
         <StyledRelativeDiv>
           <StyledInput
             aria-controls={ariaControls}
-            aria-label={label && noTopLabel ? label : undefined}
-            aria-labelledby={ariaControls}
+            aria-label={label && noTopLabel ? label : name}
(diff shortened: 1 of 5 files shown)


src/components/TextBox/index.tsx
```diff
@@ -560,8 +560,8 @@ const TextBox = forwardRef<
         <StyledRelativeDiv>
           <StyledInput
             aria-controls={ariaControls}
-            aria-label={label && noTopLabel ? label : undefined}
-            aria-labelledby={ariaControls}
+            aria-label={label && noTopLabel ? label : name}
```
chambo-e, line 563: An input name is not a label, a11y-wise this is kinda useless :)

chambo-e (changes requested):

src/components/TextBox/index.tsx
Lawndlwd: @chambo-e all good now

matthprost (approved):
All good for me 🚀

chambo-e (approved):

PR #1596, opened 2022-06-16
fix: popper component working with a function disclosure

vincentaudebert:
## Type

- Bug

#### What is expected?

Menu working with a function disclosure

(Describe what you did)

1. Fallback on previous solution

2. Seems to work + regen tests

Screenshot 2022-06-16 at 17 44 08


src/components/Menu/index.tsx
```diff
@@ -180,18 +187,15 @@ const Menu = ({
     visible,
   })
 
+  // if you need dialog inside your component, use function, otherwise component is fine
+  const target = isValidElement(disclosure) ? disclosure : disclosure(popover)
+  const innerRef = useRef(target) as unknown as RefObject<HTMLButtonElement>
```
chambo-e, line 192:
```suggestion
  const innerRef = useRef<HTMLButtonElement>(target) 
```

Does not work ?

QuiiBz (approved):

src/components/Menu/index.tsx
DorianMaliszewski: May be `HTMLButtonElement | null` ?

chambo-e (approved):

PR #1623, opened 2022-06-24
fix: fixed tooltip and button

matthprost:
## Type

- Bug

#### What is expected?

Fixed tooltip behavior with PieChart and fixed Button colors.

1. Fixed tooltip on PieChart component + position of the tooltip for animation

2. Fixed hover colors of Button component

3. Fixed story of avatar

| url  | Screenshot 2022-06-24 at 20 26 06 | Screenshot 2022-06-24 at 20 27 28 |

Screenshot 2022-06-24 at 20 27 52


philibea (approved):

src/components/Avatar/__stories__/index.stories.tsx
```diff
@@ -14,9 +14,6 @@ const Template: Story<ComponentProps<typeof Avatar>> = args => (
 )
 
 export const Default = Template.bind({})
-Default.args = {
```
chambo-e, line 17: Possible to have a proper exemple ?

chambo-e (changes requested):

src/components/Avatar/__stories__/index.stories.tsx
matthprost: @chambo-e You're right story makes no sense, I rolled back to show avatar image into default story

chambo-e (approved):

PR #1634, opened 2022-06-27
feat: new Button V2

fabienhebert:
## Type

- Feature
- Enhancement

### Summarise concisely:

Adding the new Button as **ButtonV2**
Mark Button (Button v1) as deprecated

#### What is expected?

New ButtonV2 should work following new specifications (figma/jira)

Check behaviors on deployed storybook 🙏


src/components/ButtonV2/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+      ${
+        extend
+          ? `
+        transition: max-width 450ms ease;
+        max-width: 48px;
+        [data-text] { display: none; }
+
+        &:hover, &:focus {
+          [data-text] { display: inherit }
+          max-width: 275px;
```
chambo-e, line 154: What happens with long text in French or Deutsch ?

matthprost (changes requested):
Screenshot 2022-06-28 at 09 46 59

There is a big box shadow around button on hover on design examples, maybe we should add it?

src/components/ButtonV2/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+    </Tooltip>
+  )
+}
+
+const ButtonV2 = forwardRef<HTMLButtonElement, Omit<ButtonProps, 'innerRef'>>(
+  // eslint-disable-next-line react/jsx-props-no-spreading
+  (props, ref) => <FwdButton {...props} innerRef={ref} />,
+)
+
+ButtonV2.displayName = 'fwd(ButtonV2)'
```
matthprost, line 236: Args table of button is empty I think it might be because of that 🤔 

Screenshot 2022-06-28 at 09 49 07

matthprost (changes requested):

matthprost (changes requested):
https://user-images.githubusercontent.com/15812968/176125610-aad25133-45c3-4b7c-9a3e-9505f47d944d.mov

Animation of extended is a bit broken, the + goes hidden then appear in the animation while it should stay and only the text should appear

src/components/ButtonV2/index.tsx
Exayy: Même souci que Button actuel, mais pas sure qu'il y est une meilleure solution, ou sinon en se servant d'un code JS pour connaitre la taille du contenu et ajuster (si tenté que ce soit jouable). Je vais regarder si je trouve un truc plus flexible mais avoir une animation avec une width auto ca me parrait mort en pur CSS

src/components/ButtonV2/index.tsx
Exayy: Solve with a workaround ;)

src/components/ButtonV2/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+      'target',
+    ].includes(prop),
+})<StyledButtonProps>`
+  display: inline-flex;
+  align-items: center;
+  gap: ${({ theme }) => theme.space[1]};
+
+  font-weight: 500;
+  font-size: 16px;
+  border-radius: 4px;
```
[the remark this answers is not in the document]
Exayy: done

src/components/ButtonV2/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+  icon,
+  iconPosition = 'left',
+  innerRef,
+  isLoading = false,
+  onClick,
+  prominence = 'strong',
+  size = 'large',
+  text,
+  tooltip,
+  tooltipBaseId,
```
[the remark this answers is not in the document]
Exayy: I think we could/should have a tooltip almost every time a button is disabled to explain why it is disabled. If we wan't to change ButtonV2 API, let's wait for @matthprost point of view about that (for me it's not a bad thing, it can enforce design system)

src/components/ButtonV2/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+  className,
+  extend = false,
+  icon,
+  iconPosition = 'left',
+  innerRef,
+  isLoading = false,
+  onClick,
+  prominence = 'strong',
+  size = 'large',
+  text,
```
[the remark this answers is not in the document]
Exayy: Idea there was also to solidify DesignSystem by limiting button content with only an Icon or/and a Text

PR #1644, opened 2022-06-29
fix(Charts): correct color generation

chambo-e:
## Type

- Bug

### Summarise concisely:

Chart would only support 3 colors and the remaining indexes would fallback on the last of the list

#### What is expected?

Now Chart will circle through colors, repeating chart color only every 6 items. Neutral color has been excluded has it's too transparent


vincentaudebert (approved):

vincentaudebert:
A tout hasard pas moyen de bump le code cov?

chambo-e:
> A tout hasard pas moyen de bump le code cov?

Il gueule pour rien imo, y'a eu 0% de changement de coverage

PR #1653, opened 2022-07-08
fix(text): remove margin set by headers and paragraphs

matthprost:
## Type

- Enhancement

#### What is expected?

When setting `p` or `h1`, etc tags you have a native margin which we want to get rid of.


src/components/Text/index.tsx
```diff
@@ -47,6 +47,7 @@ const generateStyles = ({
   }` as keyof typeof themeColor
 
   return `
+    margin: 0;
```
chambo-e, line 50: Where does the margin comes from ? Is this a browser stylesheet ? If this is a reset should it be put in normalize instead ?

chambo-e (approved):

chambo-e (approved):

Exayy (approved):

PR #1680, opened 2022-07-19
fix(PhoneInput): render flags correctly when dezoomed

QuiiBz:
## Type

- Bug

Fix PhoneInput component that rendered flags incorrectly when page in dezoomed.
Check if it works and if it still works when zooming.

| ?path=/docs/components-data-entry-phoneinput--default  | Screenshot 2022-07-19 at 12 24 10 | Screenshot 2022-07-19 at 12 23 19 |


src/components/PhoneInput/index.tsx
```diff
@@ -80,10 +79,6 @@ const StyledLabel = styled.label<PhoneInputLabelProps>`
     .iti__selected-dial-code {
       padding-left: 5px;
     }
-
-    .iti__flag {
-      background-image: url(${flags});
```
chambo-e, line 85: The issue is that it won't work on project that use scw-ui as flags provided by intl-tel-input are relative paths, we need to explicitly tell it where to find the flags

https://github.com/jackocnr/intl-tel-input#getting-started-using-a-bundler-eg-webpack

chambo-e (changes requested):

src/components/PhoneInput/index.tsx
QuiiBz: Fixed by using the `flags.png` image instead of `flags@2x` that was used as `flags.png`

chambo-e: So we loose support for high dpi devices ? 😄

QuiiBz: Pushed a commit that adds back the `flags@2x` image, so we're back to the state before.

QuiiBz:
Closing since we will remove this component soon from `scaleway-ui`

PR #1686, opened 2022-07-21
fix(button): progress with icon

QuiiBz:
## Type

- Bug

Fix button that was displaying two loading icons if you set `progress` and `icon` instead of the progress icon and the specified icon. Also add a snapshot test and a story.

| ?path=/docs/components-button-button--progress-and-icon | Screenshot 2022-07-21 at 16 08 29 | Screenshot 2022-07-21 at 16 07 42 |


baptisteArno (approved):

src/components/Button/index.tsx
```diff
@@ -420,7 +420,7 @@ const FwdButton = ({
             margin={iconMargin}
             position={children ? 'left' : undefined}
           >
-            {progress ? (
+            {progress === 'left' ? (
```
chambo-e, line 423: Progress can be a boolean too no ?

chambo-e (approved):

PR #1690, opened 2022-07-22
feat: drop to prop on link component

DorianMaliszewski:
## Type

- Enhancement

#### What is expected?

Drop to prop on Link


chambo-e (approved):

src/components/Button/index.tsx
```diff
@@ -393,14 +395,16 @@ const FwdButton = ({
   tooltipBaseId,
   type: elementType = 'button',
   variant = 'primary',
+  as: asProp,
   ...props
 }: ButtonProps) => {
   const as = useMemo(() => {
     if (disabled) return 'button'
+    if (asProp) return asProp
     if (to || href || download) return Link
```
philibea, line 404: Link doesn't accept `to` props anymore so we need to remove `to` from here and all our usage into Button ?

baptisteArno (approved):

src/components/Button/index.tsx
DorianMaliszewski: I put something to keep the old usage of Button.

philibea: But this will break no ? 

If we use a props `to` and we used our Link ?

philibea (approved):

PR #1697, opened 2022-07-27
fix: remove all errors and enable typecheck

matthprost:
## Type

- Enhancement

#### What is expected?

To remove all erros and enable typecheck in to CI in order to avoid mistakes.


chambo-e (approved):
So much love

philibea (approved):

PR #1698, opened 2022-07-27
fix: display link with inline flex

DorianMaliszewski:
## Type

- Bug

#### What is expected?

Display inline flex is a better behaviour than flex


matthprost (approved):

chambo-e (commented):
Might need to correct default Link story :)

chambo-e (approved):

PR #1704, opened 2022-07-28
fix(icons): removed product icons

matthprost:
## Type

- Enhancement

#### What is expected?

Remove products icons as we don't need them anymore since last refacto.


chambo-e (approved):

chambo-e (commented):
Needs to be removed in the customViewBoxes list too :)

philibea (approved):

PR #1719, opened 2022-08-01
fix(link): add on click for react router

matthprost:
## Type

- Bug

#### What is expected?

Added `onClick` on link component as it is required by react router to be able to work correctly.

### DIFF
--- a/src/components/Link/index.tsx
+++ b/src/components/Link/index.tsx
@@ -3,6 +3,7 @@ import React, {
   AnchorHTMLAttributes,
   ForwardedRef,
   HTMLAttributeAnchorTarget,
+  MouseEventHandler,
   ReactNode,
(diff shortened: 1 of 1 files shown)


src/components/Link/index.tsx
```diff
@@ -21,6 +21,7 @@ type LinkProps = {
   rel?: AnchorHTMLAttributes<HTMLAnchorElement>['rel']
   className?: string
   href: string
+  onClick?: () => void
```
chambo-e, line 24: Can we add a comment indicating that this is mainly for react router and should not be explcitiely used ?

matthprost: @chambo-e Yes, it's added!

chambo-e (approved):

PR #1739, opened 2022-08-08
feat(menu): add default z-index

baptisteArno:
## Summary

Add `z-index: 1` to the menu list. This helps having consistent z positioning when the list is over other positioned elements


baptisteArno:
`z-index` forbidden to avoid conflicts with `Modal` and other components

PR #1744, opened 2022-08-09
fix(PieChart): tooltip alignement

matthprost:
## Type

- Bug

#### What is expected?

To align text on left into PieChart tooltip.

| url  | Screenshot 2022-08-09 at 11 48 19 | Screenshot 2022-08-09 at 11 46 42 |

### DIFF
--- a/src/components/PieChart/Tooltip.tsx
+++ b/src/components/PieChart/Tooltip.tsx
@@ -12,13 +12,8 @@ const StyledItem = styled.li`
   display: flex;
   margin-top: 6px;
   width: 100%;
-`
-
-const Space = styled.span`
-  position: relative;
-  display: flex;
-  flex: 1;
-  min-width: 5px;
+  justify-content: space-between;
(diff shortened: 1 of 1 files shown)


src/components/PieChart/Tooltip.tsx
```diff
@@ -38,25 +42,25 @@ const Tooltip = ({ data }: TooltipProps) => (
   <div tabIndex={-1} role="tooltip">
     <StyledList>
       <StyledItem>
-        <Text as="p" variant="body" prominence="stronger">
+        <StyledText as="p" variant="body" prominence="stronger">
```
chambo-e, line 45: Could we instead have a flex space-between on the StyledItem ? 

We could remove Space also

chambo-e (changes requested):

QuiiBz (approved):

chambo-e (approved):

PR #1746, opened 2022-08-09
fix: labelDescription parm

matthprost:
## Type

- Bug

#### What is expected?

I change a prop by mistake.

### DIFF
--- a/src/components/RadioBorderedBox/index.tsx
+++ b/src/components/RadioBorderedBox/index.tsx
@@ -119,7 +119,10 @@ const RadioBorderedBox = ({
           >
             {label}
           </Radio>
-          {labelDescription ? (
+          {labelDescription && typeof labelDescription === 'function'
+            ? labelDescription
+            : null}
+          {labelDescription && typeof labelDescription !== 'function' ? (
             <Text as="span" variant="body" disabled={disabled}>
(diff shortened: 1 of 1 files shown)


philibea (approved):

src/components/RadioBorderedBox/index.tsx
```diff
@@ -79,7 +79,7 @@ type RadioBorderedBoxProps = RadioProps & {
   badgeVariant?: ComponentProps<typeof Badge>['variant']
   children?: ReactNode
   label: string
-  labelDescription?: string
+  labelDescription?: ReactNode
```
chambo-e, line 82: A ReactNode in a Text ?

src/components/RadioBorderedBox/index.tsx
```diff
@@ -119,7 +119,10 @@ const RadioBorderedBox = ({
           >
             {label}
           </Radio>
-          {labelDescription ? (
+          {labelDescription && typeof labelDescription === 'function'
+            ? labelDescription
+            : null}
+          {labelDescription && typeof labelDescription !== 'function' ? (
```
chambo-e, line 125: These are exact opposite ? Why not one ternary ?

PR #1751, opened 2022-08-10
feat: port Description from console

chambo-e:


src/components/Description/index.tsx
```diff
@@ -1,56 +1,70 @@
...
+        overflow: hidden;
+        text-overflow: ellipsis;
+        width: ${width};
+      }
+    `}
 
-  ${({ theme, selectable }) =>
-    selectable &&
+  ${({ userSelect, theme }) =>
+    userSelect &&
```
matthprost, line 65: Use ternary instead ?

```ts
ellipsis ? ... : null
```

src/components/Description/index.tsx
```diff
@@ -1,56 +1,70 @@
...
+  ${({ inline }) =>
+    inline &&
+    css`
+      display: flex;
+      justify-content: flex-start;
+      align-items: center;
+      ${Desc} {
+        margin: 0 0 0 8px;
+      }
+    `};
```
matthprost, line 51: Here too

src/components/Description/index.tsx
```diff
@@ -1,56 +1,70 @@
...
+
+const Desc = styled('dd', {
+  shouldForwardProp: prop => !['capitalize'].includes(prop),
+})<{ capitalize?: boolean }>`
+  display: flex;
+  align-items: center;
   color: ${({ theme }) => theme.colors.neutral.text};
-  margin: 0;
-  margin-top: ${({ theme }) => theme.space['1']};
+  margin: 8px 0 0 0;
```
matthprost, line 19: Can we use theme spacing ?

src/components/Description/index.tsx
```diff
@@ -1,56 +1,70 @@
...
+  userSelect?: boolean
+  width?: string
+}
+
+const BareDescription = styled('dl', {
+  shouldForwardProp: prop =>
+    !['inline', 'ellipsis', 'userSelect'].includes(prop),
+})<DescriptionProps>`
   font-size: 16px;
   line-height: 16px;
```
matthprost, line 39: Possible to use text / theme.typography properties instead of hard coded values ?

matthprost (changes requested):

src/components/Description/index.tsx
```diff
@@ -1,56 +1,70 @@
 import { css } from '@emotion/react'
 import styled from '@emotion/styled'
-import PropTypes from 'prop-types'
-import { WeakValidationMap } from 'react'
-import Box, { BoxProps } from '../Box'
 
-const Term = styled(Box.withComponent('dt'))`
+const Term = styled.dt`
   font-weight: 500;
```
matthprost, line 5: Same here we could use values from a Text or use `theme.typography`

matthprost (changes requested):

src/components/Description/index.tsx
chambo-e: @matthprost Why ? In template literal it does not makes a lot of sense 🤔

src/components/Description/index.tsx
chambo-e: There is no typography with a line-height 16

matthprost: Maybe we should see with design what kind of typography we can use in this case?

src/components/Description/index.tsx
matthprost: @chambo-e why not?

chambo-e: Sure, it's just more verbose with no safety addition :)

src/components/Description/index.tsx
chambo-e: @matthprost so process design system, can we leave it as is for this PR and refactor it when the component as been spec-ed ?

matthprost: @chambo-e yes go ahead :)

matthprost (approved):

philibea (approved):

Issue #548, opened 2021-07-19
refactor: xstyled removal


chambo-e:
We'd like to move on from xstyled to have a more conventional usage of our components. Meaning no props based styled modifiers, spacing should be handled by the parent component

This issue might be transformed into a milestone at some point because it's a huge task

chambo-e:
@baptisteArno wut ?

baptisteArno:
@matthprost wut ?

PR #1753, opened 2022-08-10
feat(Notice): migrate from Box to div

chambo-e:
Related to #548

### DIFF
--- a/src/components/TextBox/index.tsx
+++ b/src/components/TextBox/index.tsx
@@ -623,7 +623,7 @@ const TextBox = forwardRef<
         <Expandable height={56} overflow="hidden" opened={!!error}>
           <StyledError>{error}</StyledError>
         </Expandable>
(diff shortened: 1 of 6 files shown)


src/components/TextBox/index.tsx
```diff
@@ -620,7 +623,7 @@ const TextBox = forwardRef<
         <Expandable height={56} overflow="hidden" opened={!!error}>
           <StyledError>{error}</StyledError>
         </Expandable>
-        {notice && <Notice mt={1}>{notice}</Notice>}
+        {notice && <StyledNotice>{notice}</StyledNotice>}
```
matthprost, line 626:
```suggestion
        {notice ? <StyledNotice>{notice}</StyledNotice> : null}
```

matthprost (changes requested):
It will also mean a migration on projects using Notice :)

matthprost (approved):

baptisteArno (approved):

PR #1773, opened 2022-08-16
fix(SelectableCard): avoid hidding all children svg

matthprost:
## Type

- Bug

#### What is expected?

Avoid removing all svg children, only the one of the input.

| url  | Screenshot 2022-08-16 at 11 45 48 | Screenshot 2022-08-16 at 11 45 16 |

### DIFF
--- a/src/components/SelectableCard/index.tsx
+++ b/src/components/SelectableCard/index.tsx
@@ -55,7 +55,7 @@ const StyledElement = (element: typeof Radio) => styled(element, {
     }}
   }
 
-  svg {
+  input + svg {
     ${({ showTick }) => (!showTick ? `display: none;` : null)}
   }
 `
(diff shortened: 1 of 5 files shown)


src/components/SelectableCard/index.tsx
```diff
@@ -55,7 +55,7 @@ const StyledElement = (element: typeof Radio) => styled(element, {
     }}
   }
 
-  svg {
+  & + svg {
```
chambo-e, line 58: Can we target a styled instead ?

chambo-e: Or handle this case in the Radio and Checkbox ?

matthprost: @chambo-e we cannot target styled as it is different depending if it is a radio or a checkbox, also I prefer not creating a prop into those components to display or not tick.

Also you make me notice selector is wrong it should be: 

```js
input + svg {...}
```

This way only the svg after input is removed and no others. As svg is always after input it should work in any scenario.

QuiiBz (approved):

PR #1787, opened 2022-08-17
fix: remove phone input component

Lawndlwd:
## Type

- Migration

#### What is expected?

Remove Phone input from Scw-ui


matthprost (changes requested):
There is all visual testing from PhoneInput to remove too :)

chambo-e (approved):

matthprost (approved):

PR #1792, opened 2022-08-18
feat(Toggle): refactor of component

### LINKED ISSUE #1791: refactor: Toggle

matthprost:
## Summary

Closes #1791 

## Type

- Enhancement

#### What is expected?

Refactor of toggle component with animation.

https://user-images.githubusercontent.com/15812968/185442012-4cc6f34b-d64f-4c11-99ff-c25e14c3137f.mov


src/components/Toggle/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  'data-disabled': boolean
+  'data-size': 'small' | 'large'
+}>`
+  box-sizing: content-box;
+  outline: none;
+  overflow: hidden;
+  display: flex;
+  align-items: center;
+  border: none;
+  border-radius: 34px;
```
DorianMaliszewski, line 35: Should not be hardcoded

src/components/Toggle/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+  &:after {
+    content: '';
+    position: absolute;
+    top: ${({ 'data-size': dataSize }) =>
+      SIZES[dataSize].height / 2 - SIZES[dataSize].ball / 2}px;
+    left: 5px;
+    width: ${({ 'data-size': dataSize }) => SIZES[dataSize].ball}px;
+    height: ${({ 'data-size': dataSize }) => SIZES[dataSize].ball}px;
+    border-radius: 100px;
```
DorianMaliszewski, line 50: Here too

src/components/Toggle/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+    ball: 12,
+    height: 20,
+    width: 40,
+  },
+} as const
+
+const StyledToggle = styled.div<{
+  'data-checked': boolean
+  'data-disabled': boolean
+  'data-size': 'small' | 'large'
```
DorianMaliszewski, line 27: Dont need to use `data-size` because the css doesn't depend of its value. Use `size` prop and don't forget to use shouldForwardProp 😄

src/components/Toggle/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  height: 100%;
+  cursor: pointer;
+
+  &[disabled] {
+    cursor: not-allowed;
+  }
+`
+
+const StyledLabel = styled.label<{
+  'data-size': 'small' | 'large'
```
DorianMaliszewski, line 103: Here too dont need a data attribute

src/components/Toggle/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+            name={name}
+            onChange={onLocalChange}
+            type="checkbox"
+          />
+        </StyledToggle>
+        {label && labelPosition === 'right' ? (
+          <LabelContent label={label} labelPosition={labelPosition} />
+        ) : null}
+      </StyledLabel>
+    </Tooltip>
```
DorianMaliszewski, line 206: There is no more label inside the toggle ? We use it in our console we will need to replace them

DorianMaliszewski (changes requested):

src/components/Toggle/index.tsx
matthprost: @DorianMaliszewski there is no more, as seen with design the toggle is explicit enough to make the user understand that option is on / off. All toggle with text inside on console will be soon replaced with no text inside and a label on left or right side 😄

src/components/Toggle/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+      },
+    },
+  },
+  title: 'Components/Data Entry/Toggle',
+} as Meta
+
+const Template: Story<ComponentProps<typeof Toggle>> = ({
+  checked,
+  ...args
+}) => <Toggle {...args} name="Toggle" onChange={() => {}} />
```
baptisteArno, line 22:
```suggestion
}) => <Toggle {...args} name="Toggle" />
```

baptisteArno (changes requested):

src/components/Toggle/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+}: Required<Pick<ToggleProps, 'label' | 'labelPosition'>>) => {
+  if (typeof label === 'string') {
+    return (
+      <StyledLabelContent labelPosition={labelPosition}>
+        {label}
+      </StyledLabelContent>
+    )
+  }
+
+  return <>{label}</>
```
QuiiBz, line 142: Can't we return `label` directly?

```suggestion
  return label
```

src/components/Toggle/__stories__/index.stories.tsx
baptisteArno: Otherwise, the code examples don't look "clean"
CleanShot 2022-08-24 at 17 45 32@2x

QuiiBz (approved):

baptisteArno (approved):

PR #1802, opened 2022-08-22
feat(SwitchButton): full refacto

matthprost:
## Type

- Refactor

#### What is expected?

To refactor SwitchButton, in fact this component is make out of two `SelectableCard` and only two options are available.
I also removed variant that should not be used and added a BorderedBox around.

Screenshot 2022-08-22 at 16 18 58

https://user-images.githubusercontent.com/15812968/186159881-a8981a14-92db-4f5e-a945-eecb57077a37.mov


DorianMaliszewski:
It is weird to have a `selected` prop which can be left or right, it should be the value that match left or right value ? @matthprost  🤔

QuiiBz:
When left is selected, the shadow from right is above left:

https://user-images.githubusercontent.com/43268759/186124692-8a2099e9-023d-4a7a-a1b0-98ac11d0e445.mov

src/components/SwitchButton/index.tsx
```diff
@@ -1,204 +1,141 @@
...
-        onFocus={onFocus}
-        type="radio"
-        value={value}
-      />
-    </StyledSwitch>
-  </Tooltip>
-)
+}: SwitchButtonProps) => {
+  const [localValue, setLocalValue] = useState(
+    value === 'left' ? leftValue : rightValue,
```
QuiiBz, line 100: Shouldn't we check `if (value === leftValue)` instead?

matthprost: @QuiiBz You speak truth

QuiiBz (approved):

src/components/SwitchButton/index.tsx
```diff
@@ -1,204 +1,141 @@
...
   tooltip?: string
-  value: string
-  checked?: boolean
-  disabled?: boolean
-  variant?: Variants
+  value?: string | number
+  leftValue: string | number
+  rightValue: string | number
+  leftText: string
+  rightText: string
```
baptisteArno, line 84: I think it would be a better DX to have label-value pairs. But it's up to you!

```suggestion
  leftButton: {
    label: string,
    value: string
  },
  rightButton: {
    label: string,
    value: string
  }
```

baptisteArno (approved):

src/components/SwitchButton/index.tsx
matthprost: @baptisteArno I like your idea

PR #1807, opened 2022-08-25
feat: new tabs components

DorianMaliszewski:
## Type

- Feature

#### What is expected?

New tabs component

Capture d’écran 2022-08-25 à 16 14 52


src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,176 @@
...
+  border-radius: 50%;
+  font-size: 12px;
+  background-color: ${({ theme }) => theme.colors.neutral.backgroundStrong};
+  border: 1px solid ${({ theme }) => theme.colors.neutral.borderStrong};
+  margin-left: ${({ theme }) => theme.space['1']};
+  height: 24px;
+  width: 24px;
+  text-overflow: ellipsis;
+  line-height: 22px;
+`
```
matthprost, line 26: It couldn't work to use component `Text` ? So we can remove all props linked to text

src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,176 @@
...
+  text-decoration: none;
+  user-select: none;
+  touch-action: manipulation;
+  transition: color 0.2s;
+  border: none;
+  background: none;
+  border-bottom-width: 2px;
+  border-bottom-style: solid;
+  border-bottom-color: ${({ theme }) => theme.colors.neutral.borderWeak};
+  outline: none;
```
matthprost, line 56: Here also we have a lot of props linked to text maybe we should use a component text inside of this component?

src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,176 @@
...
+  &[aria-selected='true'] {
+    color: ${({ theme }) => theme.colors.primary.text};
+    border-bottom-color: ${({ theme }) => theme.colors.primary.borderWeak};
+    ${StyledCounter} {
+      background-color: ${({ theme }) => theme.colors.primary.background};
+      border-color: ${({ theme }) => theme.colors.primary.background};
+    }
+  }
+
+  &:not([aria-disabled='true']):not(:disabled) {
```
matthprost, line 78:
```suggestion
  &[aria-disabled='false']:not(:disabled) {
```

?

src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,176 @@
...
+    badge,
+    children,
+    className,
+    counter,
+    disabled = false,
+    value,
+    onClick,
+    onKeyDown,
+    tooltip,
+    ...props
```
matthprost, line 133: Do we still need props spreading ?

matthprost (changes requested):

src/components/Tabs/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  margin-block-start: 0;
+  margin-block-end: 0;
+  padding-inline-start: 0;
+  overflow-x: scroll;
+  position: relative;
+  z-index: 0;
+  list-style-type: none;
+
+  &::after {
+    z-index: -1;
```
matthprost, line 20: Why using z-index ? Maybe put it into `::before` ?

src/components/Tabs/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const TabsContainer = styled.ul`
+  display: flex;
+  flex-wrap: nowrap;
+  margin-block-start: 0;
+  margin-block-end: 0;
+  padding-inline-start: 0;
+  overflow-x: scroll;
+  position: relative;
+  z-index: 0;
```
matthprost, line 16: Please try to avoid z-index

matthprost (changes requested):

src/components/Tabs/Tab.tsx
DorianMaliszewski: Yes because of as prop

src/components/Tabs/index.tsx
DorianMaliszewski: Doesn't work too need z-index 😞

matthprost (changes requested):
Screenshot 2022-08-26 at 10 23 59
Screenshot 2022-08-26 at 10 24 03

Seems like horizontal scroll bar is all the time displayed and not compatible with dark theme 🤔

BlandinePelat:
- More tabs shouldn't take the width left 
- Do we have the behavior when we select a choice inside the more option?

src/components/Menu/index.tsx
```diff
@@ -190,7 +196,11 @@ const Menu = ({
 
   // if you need dialog inside your component, use function, otherwise component is fine
   const target = isValidElement(disclosure) ? disclosure : disclosure(popover)
-  const innerRef = useRef(target) as unknown as RefObject<HTMLButtonElement>
+  const innerRef = useRef(
+    target,
+  ) as unknown as MutableRefObject<HTMLButtonElement | null>
```
QuiiBz, line 201: Why do we need to cast the whole `useRef`? I would prefer to cast only `target`

src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,183 @@
...
+        as={computedAs}
+        aria-label={value ? `${value}` : undefined}
+        aria-selected={isSelected}
+        aria-disabled={disabled}
+        disabled={computedAs === 'button' ? disabled : undefined}
+        onClick={event => {
+          if (value) {
+            onChange(value)
+          }
+          ;(onClick as MouseEventHandler<HTMLElement>)?.(event)
```
QuiiBz, line 163: `onClick` already has a type of `MouseEventHandler<HTMLElement> | undefined`, why do we need to cast?

src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,183 @@
...
+        }}
+        onKeyDown={event => {
+          ;(onKeyDown as KeyboardEventHandler<HTMLElement>)?.(event)
```
QuiiBz, line 166: Same as `onClick`

src/components/Tabs/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+import TabMenu from './TabMenu'
+import TabMenuItem from './TabMenuItem'
+import TabsContext from './TabsContext'
+
+const MenuContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+
+  ${StyledTabButton} {
+    font-size: 14px;
```
QuiiBz, line 20: Font size can be from theme

src/components/Tabs/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+        tabsRef.current.scrollWidth - tabsRef.current.clientWidth
+          ? 'none'
+          : ''
+    }
+    if (element && displayMore) {
+      element.addEventListener('scroll', handler)
+    }
+
+    return () => {
+      if (element) element.removeEventListener('scroll', handler)
```
QuiiBz, line 124: I think we should also check for `displayMore` as above in this `if`

QuiiBz (approved):

matthprost (approved):

PR #1814, opened 2022-08-29
fix: button icon color when disabled

DorianMaliszewski:
## Type

- Bug

#### What is expected?

Icon has textDisabled color when button is disabled (fill property)

1. Fix CSS


matthprost (approved):

src/components/Button/index.tsx
```diff
@@ -334,6 +334,9 @@ const StyledButton = styled(Box, {
     cursor: default;
     pointer-events: none;
     color: ${theme.colors.neutral.textDisabled};`}
+    svg {
+    fill: ${({ theme }) => theme.colors.neutral.textDisabled};
+  }
```
QuiiBz, line 339: Bad indentation:

```suggestion
      fill: ${({ theme }) => theme.colors.neutral.textDisabled};
    }
```

QuiiBz (approved):

PR #1826, opened 2022-08-30
fix: use `useId` instead of `useUUID`

### LINKED ISSUE #1790: fix: useId to generate id instead of getUUID
Since react 18 has a hook for generating IDs we can directly use it instead of using `getUUID` helper. Helper should also be removed.

QuiiBz:
## Summary

Closes #1790

## Type

- Refactor

#### What is expected?

Use React 18 `useId` hook instead of `useUUID`.
Remove the exports from index of `useUUID` and `getUUID`.
I could not delete `getUUID` because it's still used in places where we can't use `useId`

### DIFF
--- a/src/components/SelectableCard/__tests__/__snapshots__/index.tsx.snap
+++ b/src/components/SelectableCard/__tests__/__snapshots__/index.tsx.snap
@@ -966,7 +966,7 @@ exports[`SelectableCard renders correctly with checkbox type and tooltip prop 1`
 }
 
 <div
-    aria-describedby="tooltip-6c756c756c756c756c756c756c756c75"
+    aria-describedby=":rn:"
   >
     <label
       aria-disabled="false"
@@ -2111,7 +2111,7 @@ exports[`SelectableCard renders correctly with radio type and tooltip prop 1`] =
 }
 
 <div
-    aria-describedby="tooltip-6c756c756c756c756c756c756c756c75"
+    aria-describedby=":rl:"
   >
     <label
       aria-disabled="false"
--- a/src/utils/ids.ts
+++ b/src/utils/ids.ts
@@ -1,9 +1,5 @@
-import { useMemo } from 'react'
-
 const chr4 = () => Math.random().toString(16).slice(-4)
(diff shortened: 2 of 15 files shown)


src/utils/ids.ts
```diff
@@ -1,9 +1,5 @@
-import { useMemo } from 'react'
-
 const chr4 = () => Math.random().toString(16).slice(-4)
 const uuid = () => Array.from({ length: 8 }, chr4).join('')
 
 export const getUUID = (prefix = ''): string =>
   `${prefix ? `${prefix}-` : ''}${uuid()}`
```
matthprost, line 5: To remove ?

src/utils/ids.ts
```diff
@@ -1,9 +1,5 @@
-import { useMemo } from 'react'
-
 const chr4 = () => Math.random().toString(16).slice(-4)
 const uuid = () => Array.from({ length: 8 }, chr4).join('')
```
matthprost, line 2: Maybe all the file actually

matthprost (changes requested):

src/components/SelectableCard/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -2111,7 +2111,7 @@ exports[`SelectableCard renders correctly with radio type and tooltip prop 1`] =
 }
 
 <div
-    aria-describedby="tooltip-6c756c756c756c756c756c756c756c75"
+    aria-describedby=":rl:"
```
matthprost, line 2114: is that the normal format of id ??

matthprost (changes requested):

src/utils/ids.ts
QuiiBz: From the PR description:

> I could not delete `getUUID` because it's still used in places where we can't use `useId`

src/components/SelectableCard/__tests__/__snapshots__/index.tsx.snap
QuiiBz: Yes

matthprost (approved):

PR #1828, opened 2022-08-30
feat: change loader by placeholder in List

Lawndlwd:
## Type

- Feature

Replace classic loader with the list placeholder


Lawndlwd:
TODO: add default perPage

src/components/List/index.tsx
```diff
@@ -87,20 +80,13 @@ function Body<DataType extends Record<string, unknown>>({
     customLoader,
     isLoading,
     emptyListComponent,
-    perPage,
+    perPage = 2,
```
matthprost, line 83: Why?

baptisteArno (changes requested):

matthprost (approved):

baptisteArno:
That still doesn't feel right IMO

CleanShot 2022-09-01 at 15 34 03@2x

Here I would expect 3 lines per row (none for the checkbox column) and each line would have the width of the row

Lawndlwd:
> Here I would expect 3 lines per row (none for the checkbox column) and each line would have the width of the row

the 4th line not for the checkbox, it is for unlabelled col that contain action,
yeah, I tried to get style from the list it so complicated, i prefer to go with this for now @baptisteArno

philibea (approved):

baptisteArno (approved):

PR #1835, opened 2022-09-06
fix(SelectableCard): encapsulate children into div and add label

matthprost:
## Type

- Bug

SelectabledCard was problematic when having an input element as children. This PR solve this issue.

Screenshot 2022-09-07 at 10 00 39


baptisteArno (approved):

src/components/SelectableCard/index.tsx
```diff
@@ -83,57 +86,66 @@ type SelectableCardProps = {
...
+  tooltip,
+  id,
+  label,
+}: SelectableCardProps) => {
+  const Element = useMemo(
+    () => StyledElement((type === 'radio' ? Radio : Checkbox) as typeof Radio),
+    [type],
+  )
+
+  const ref = useRef<HTMLInputElement>({} as HTMLInputElement)
```
QuiiBz, line 120:
```suggestion
  const ref = useRef<HTMLInputElement>(null)
```

QuiiBz (approved):

PR #1837, opened 2022-09-07
fix(richselect): multi value padding without top label

baptisteArno:
## Summary

Fix top padding when rich select has multi values and `noTopLabel` enabeld

## Type

- Bug

| Before | After |
| :-: | :-: |
|  CleanShot 2022-09-07 at 10 41 37@2x | CleanShot 2022-09-07 at 10 41 02@2x |


src/components/RichSelect/index.tsx
```diff
@@ -199,7 +199,7 @@ const getSelectStyles = ({
     fontWeight: 500,
     height: '24px',
     justifyContent: 'center',
-    marginTop: '11px',
+    marginTop: noTopLabel ? '4px' : '11px',
```
matthprost, line 202: Could we use `theme.space['0.5']` ?

matthprost (changes requested):

baptisteArno:
Also improved spacing when top label is present IMO:

| Before | After |
| :-: | :-: |
| CleanShot 2022-09-08 at 10 06 15@2x | CleanShot 2022-09-08 at 10 06 27@2x |

src/components/RichSelect/index.tsx
```diff
@@ -199,7 +199,7 @@ const getSelectStyles = ({
     fontWeight: 500,
     height: '24px',
     justifyContent: 'center',
-    marginTop: '11px',
+    marginTop: noTopLabel ? theme.space['0.5'] : '16px',
```
QuiiBz, line 202: We can use `theme.space['2']` instead of hardcoding `16px`

QuiiBz (approved):

philibea (approved):

matthprost (approved):

PR #1839, opened 2022-09-07
fix(PlaceHolder): make it compatible dark theme

matthprost:
## Type

- Bug

#### What is expected?

Make placeholder compatible dark theme

### DIFF
--- a/src/components/Placeholder/index.tsx
+++ b/src/components/Placeholder/index.tsx
@@ -35,11 +35,11 @@ const StyledDiv = styled.div`
   opacity: 0.8;
   background: linear-gradient(
     90deg,
-    rgba(255, 255, 255, 0),
-    rgba(255, 255, 255, 0.3),
-    rgba(255, 255, 255, 0.4),
-    rgba(255, 255, 255, 0.3),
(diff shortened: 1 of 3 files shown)


src/components/Placeholder/index.tsx
```diff
@@ -35,11 +35,11 @@ const StyledDiv = styled.div`
   opacity: 0.8;
   background: linear-gradient(
     90deg,
-    rgba(255, 255, 255, 0),
-    rgba(255, 255, 255, 0.3),
-    rgba(255, 255, 255, 0.4),
-    rgba(255, 255, 255, 0.3),
-    rgba(255, 255, 255, 0)
+    ${({ theme }) => theme.colors.neutral.backgroundWeak}00,
```
philibea, line 38: what ???

matthprost: @philibea we wanted tokens from theme + opacity, in hexa you can simply add values at the end it to change opacity which we did here

ref: https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4

baptisteArno (approved):

QuiiBz (approved):

PR #1843, opened 2022-09-08
ci: improve tolerance config

baptisteArno:
## Summary

Improve tolerance config

## Type

- CI


baptisteArno:
It seems that the config doesn't accept argument below `1`, so we're on the minimum already.

PR #1844, opened 2022-09-09
feat(switchbutton): add animation

baptisteArno:
## Summary

Add `SwitchButton` toggle animation

## Type

- Enhancement

### Before

https://user-images.githubusercontent.com/16015833/189287985-8ffb5b5b-0178-46cd-96dc-1f0fe0456725.mp4

### After

https://user-images.githubusercontent.com/16015833/189288017-585dcd2c-258c-4217-88b3-167b26c667d6.mp4


src/components/SwitchButton/index.tsx
```diff
@@ -72,6 +109,12 @@ const SwitchButton = ({
   rightButton,
   tooltip,
 }: SwitchButtonProps) => {
+  const leftButtonRef = useRef<HTMLLabelElement | null>(null)
+  const rightButtonRef = useRef<HTMLLabelElement | null>(null)
```
QuiiBz, line 113: Not sure but I think we can remove the `| null`
```suggestion
  const leftButtonRef = useRef<HTMLLabelElement>(null)
  const rightButtonRef = useRef<HTMLLabelElement>(null)
```

src/components/SwitchButton/index.tsx
```diff
@@ -80,11 +123,31 @@ const SwitchButton = ({
...
       <div style={{ display: 'inline-flex' }}>
-        <StyledBorderedBox>
+        <StyledBorderedBox
+          onMouseDown={setMouseDown(true)}
+          onMouseUp={setMouseDown(false)}
+          onMouseLeave={setMouseDown(false)}
+        >
+          <FocusOverlay
+            leftCardWidth={leftCardWidth}
+            rightCardWidth={rightCardWidth}
```
QuiiBz, line 145: Can we use data attributes instead?

baptisteArno: data attributes would only work for discrete values, so can't do this for `width` and `transform` props. But I now compute them in the inline `style` (to avoid changing the class name on every change). I used data attribute for focus position.

Let me know what you think. Can we do better?

matthprost (changes requested):
When you load page or when you change theme there is a transition on width of div used as background. I don't really know how this could be solved, its initial size should be size of the width of the switch button initially selected

https://user-images.githubusercontent.com/15812968/189354306-4331d4d2-89f8-4d73-b9b6-4425f97bbda4.mov

baptisteArno:
> When you load page or when you change theme there is a transition on width of div used as background. I don't really know how this could be solved, its initial size should be size of the width of the switch button initially selected
> 
>  Screen.Recording.2022-09-09.at.14.52.44.mov

That's not a bug, but a feature 🚀 🚀

Haha, joke aside. I'll see what I can do 👍

src/components/SwitchButton/index.tsx
```diff
@@ -80,11 +102,47 @@ const SwitchButton = ({
...
+          onMouseUp={setMouseDown(false)}
+          onMouseLeave={setMouseDown(false)}
+        >
+          <FocusOverlay
+            data-focusposition={
+              localValue === leftButton.value ? 'left' : 'right'
+            }
+            style={{
+              transform: `translate3d(${
+                localValue === leftButton.value ? 0 : leftCardWidth + 8
```
QuiiBz, line 137: Where does the `8` come from?  I would prefer to have it in a top-level const with a name to understand what it is.

src/components/SwitchButton/index.tsx
```diff
@@ -80,11 +102,47 @@ const SwitchButton = ({
...
+    setLeftCardWidth(leftButtonRef.current.getBoundingClientRect().width)
+    setRightCardWidth(rightButtonRef.current.getBoundingClientRect().width)
+  }, [leftButton.value, leftButtonRef, localValue, rightButtonRef])
+
+  const setMouseDown = (isMouseDown: boolean) => () =>
+    setHasMouseDown(isMouseDown)
+
+  const getScaleValue = () => {
+    if (!hasMouseDown) return 1
+    const offsetScaleRatio = 6
```
QuiiBz, line 116: Can we extract to a top-level const?

QuiiBz (approved):
Two comments but looks good otherwise!

QuiiBz (approved):
CI is failing

baptisteArno:
RTR @matthprost 🤟

matthprost (approved):

matthprost:
@baptisteArno sorry I forgot about it, you can merge any time you want 🎉

PR #1857, opened 2022-09-12
feat(tooltip): dymanic position

matthprost:
## Type

- Enhancement

#### What is expected?

To allow placement `auto` which will positionnante tooltip depending on its size and window size. 

The goal is to always display tooltip and avoid overflow. In the worst case scenario if the tooltip is too big for any position, it will be displayed on top.

Placement is computed again when window size changes or when user scroll. 

I also made tiny change in CSS: to use `transform3d` for using GPU while rendering and increase performance.

https://user-images.githubusercontent.com/15812968/189650226-e769b335-7726-41f4-ae48-35500a86dbd6.mov

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8474,7 +8474,7 @@ packages:
       mississippi: 3.0.0
       mkdirp: 0.5.5
       move-concurrently: 1.0.1
-      promise-inflight: 1.0.1_bluebird@3.7.2
+      promise-inflight: 1.0.1
       rimraf: 2.7.1
       ssri: 6.0.2
       unique-filename: 1.1.1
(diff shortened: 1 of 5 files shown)


QuiiBz:
I don't think it's due to this PR, but the animation for max width should be left-right then right-left, and not left-right then top-bottom

https://user-images.githubusercontent.com/43268759/189685210-0296f490-d12d-4957-b191-70131100febc.mov

matthprost:
> 

@QuiiBz How to do reproduce this issue? When I try the tooltip goes on the top

QuiiBz (approved):

baptisteArno (approved):

PR #1863, opened 2022-09-13
fix(linechart): select chart even when provided asynchronously

baptisteArno:
## Summary

When data was first undefined and then defined, no graph were selected.

I've added a `useEffect` that populates the `selected` state when the data is initialized

## Type

- Enhancement

### DIFF
--- a/src/components/LineChart/index.tsx
+++ b/src/components/LineChart/index.tsx
@@ -3,7 +3,7 @@ import { DatumValue, Box as NivoBox, ValueFormat } from '@nivo/core'
 import { LineSvgProps, Point, ResponsiveLine, Serie } from '@nivo/line'
 import type { ScaleSpec } from '@nivo/scales'
 import PropTypes from 'prop-types'
-import { ComponentProps, Validator, useState } from 'react'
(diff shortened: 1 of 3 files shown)


src/components/LineChart/index.tsx
```diff
@@ -68,20 +70,25 @@ const LineChart = ({
...
   )
 
-  const finalData = dataset.datasets.filter(
-    ({ id }, index) => selected.indexOf(`${id}${index}`) > -1,
+  const finalData = dataset.datasets?.filter(({ id }, index) =>
+    selected ? selected.indexOf(`${id}${index}`) > -1 : false,
   )
 
+  useEffect(() => {
+    if (selected !== undefined) return
```
philibea, line 82: Can you add bracket ? more clear

baptisteArno: I definitely prefer this compact formatting, but if that's part of the coding style guideline, no problem! It can also be forced by an eslint rule btw!

matthprost (approved):

vincentaudebert (approved):

PR #1883, opened 2022-09-20
fix(menu): show args and valid code snippets in menu doc

baptisteArno:
## Summary

Better `Menu` documentation. Could serve as an example to refactor other storybook docs. It's open to suggestions. 😀

## Type

- Enhancement

1. Make sure args are displaying correctly below the Default component
2. Display valid code snippets
3. Better folder organization.
4. Better "Function Disclosure" doc. Now showing popover state on disclosure text.

### Before

https://user-images.githubusercontent.com/16015833/191274640-2707ae25-d163-4ad5-aafe-e4826b0c1432.mp4

### After

https://user-images.githubusercontent.com/16015833/191275100-eb508985-2e36-4b22-a0d6-d1e5eb40c26b.mp4


src/components/Menu/__stories__/ChildrenProps.tsx
```diff
@@ -0,0 +1,79 @@
...
+    </Menu.Item>
+    <Menu.Item
+      href="/?path=/docs/components-navigation-menu--children-props"
+      onClick={toggle}
+    >
+      Menu Item Link
+    </Menu.Item>
+  </>
+)}
+</Menu>`
```
baptisteArno, line 31: As you can see, we need to manually overwrite the code snippet for custom cases like this one.

(In that case that's because passing a function to `children` param isn't working as expected)

src/components/Menu/__stories__/ChildrenProps.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+ChildrenProps.parameters = {
+  docs: {
+    source: {
+      code: customCodeSnippet,
+    },
```
matthprost, line 37: Why do you need to do that in that case ?

QuiiBz (approved):

src/components/Menu/__stories__/Default.tsx
```diff
@@ -0,0 +1,24 @@
+import { fireEvent, screen } from '@storybook/testing-library'
+import Menu from '..'
+import { Template } from './Template'
+
+export const Default = Template.bind({})
```
Exayy, line 5: Optional: We could call it "Playground"

Exayy:
Sounds really nice,  especially the stories splitting 👏

Exayy (approved):

src/components/Menu/__stories__/Template.tsx
```diff
@@ -0,0 +1,15 @@
...
+import Icon from '../../Icon'
+import Touchable from '../../Touchable'
+
+const DefaultDisclosure = (
+  <Touchable title="menu" name="menu">
+    <Icon name="dots-horizontal" color="neutral" size={24} />
+  </Touchable>
+)
+
+export const Template: ComponentStory<typeof Menu> = ({
```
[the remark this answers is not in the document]
baptisteArno: Wouldn't it create a dependency cycle?

src/components/Menu/__stories__/Default.tsx
baptisteArno: Naming it `Playground` definitely makes more sense 👌

QuiiBz: Strongly agree!

matthprost: Sounds perfect !

matthprost (approved):

PR #1888, opened 2022-09-21
fix(status): classname on container

matthprost:
ClassName of status should be on container and not on one of the childrens.


QuiiBz (approved):

chambo-e (approved):

QuiiBz:
We can also add a snapshot test with `className` to check if it's applied on the right element.

PR #1895, opened 2022-09-22
feat(list): better loading placeholder

baptisteArno:
## Summary

Improve loading placeholder to avoid content shifting

## Type

- Enhancement

## Before

https://user-images.githubusercontent.com/16015833/191781289-0afc08c1-8cad-49ce-ac7f-aacd2b9fb782.mp4

## After

https://user-images.githubusercontent.com/16015833/191781355-5ddcff4b-dc20-4538-ae02-b24e24bf11e9.mp4


chambo-e:
Should we remove the List variant in Placeholder ?

src/components/List/LoadingPlaceholder.tsx
```diff
@@ -0,0 +1,28 @@
...
+  columns,
+}: LoadingPlaceholderProps<DataType>) => {
+  const rows = useMemo(() => Array.from({ length: totalRows }), [totalRows])
+
+  return (
+    <>
+      {rows.map((_, index) => (
+        <Row id={index.toString()}>
+          {columns.map(column => (
+            <Cell>{column.label ? <Placeholder variant="line" /> : null}</Cell>
```
matthprost, line 22: key might be missing as we do a map

matthprost (changes requested):

baptisteArno:
> Should we remove the List variant in Placeholder ?

We don't use it anywhere at the moment, so it's removable. Do we have things in mind for the `list` Placeholder variant? @matthprost

src/components/List/LoadingPlaceholder.tsx
baptisteArno: Should I set column index as the key? I think it's useless in that case

matthprost:
> > Should we remove the List variant in Placeholder ?
> 
> We don't use it anywhere at the moment, so it's removable. Do we have things in mind for the `list` Placeholder variant? @matthprost

Variant list is used on others project so we might want to keep it :)

DorianMaliszewski (approved):

matthprost (approved):

PR #1902, opened 2022-09-26
fix(pagination): do not update max page when list is loading

baptisteArno:
## Summary

Make sure that pagination isn't updated while the parent list is loading.

## Type

- Enhancement

### DIFF
--- a/src/components/List/index.tsx
+++ b/src/components/List/index.tsx
@@ -228,7 +228,7 @@ function List<DataType extends Record<string, unknown>>(
 
   const pagination = usePagination({
     data: sortedData,
-    isListLoading: isLoading,
+    isLoading,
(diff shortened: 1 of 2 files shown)


src/components/List/index.tsx
```diff
@@ -228,6 +228,7 @@ function List<DataType extends Record<string, unknown>>(
 
   const pagination = usePagination({
     data: sortedData,
+    isListLoading: isLoading,
```
matthprost, line 231: use `isLoading` would be a better fit no?

matthprost (changes requested):

DorianMaliszewski (approved):
LGTM after @matthprost feedback

matthprost (approved):

PR #1762, opened 2022-08-11
refactor: remove typography usage

baptisteArno:
## Summary

This PR makes sure components don't use Typography internally except Typo with `command` variant

## Type

- Refactor

## Related to

#1755

PR #1903, opened 2022-09-26
refactor(typography): remove typography

matthprost:
Closes #1762 

## Type

- Refactor

#### What is expected?

Removal of typography and it migration page.


src/components/MarkDown/index.tsx
```diff
@@ -38,7 +37,9 @@ const headingRenderer = ({
 }
 
 const inlineCodeRenderer = ({ children }: { children: ReactNode }) => (
-  <Typography variant="command">{children}</Typography>
+  <Text as="code" variant="code">
```
chambo-e, line 40: We'll miss a background here no ?

baptisteArno (approved):

src/components/MarkDown/index.tsx
matthprost: @chambo-e yes you're right I forgot it :)

chambo-e (approved):

PR #1904, opened 2022-09-27
fix: radio checkbox and selectable card visible input

matthprost:
## Type

- Bug

#### What is expected?

To remove visible input in radio / checkbox / selectable card / switch button.

I made the input as same size as svg but with no opacity so when user click on svg he is actually clicking on hidden input. This way we keep accessibility and tick customisation.

I also fixed hovered state of selectable card, when disabled border was still getting primary color.

| url  | Screenshot 2022-09-27 at 14 04 47 | Screenshot 2022-09-27 at 14 04 53 |


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@scaleway/ui",
-  "version": "0.194.1",
+  "version": "2.194.1",
```
chambo-e, line 3: wut ?

chambo-e (changes requested):

src/components/Checkbox/index.tsx
```diff
@@ -56,14 +56,10 @@ const StyledReakitCheckbox = styled(ReakitCheckbox, {
...
-  width: 1px;
-  height: 1px;
-  padding: 0;
-  margin: -1px;
-  overflow: hidden;
-  clip: rect(0, 0, 0, 0);
   white-space: nowrap;
+  height: ${({ size }) => size}px;
+  width: ${({ size }) => size}px;
+  opacity: 0;
```
chambo-e, line 62: If I remember correctly, opacity 0 on safari and some other browser is considered non present element, so you won't be able to click on the checkbox

chambo-e (changes requested):

package.json
matthprost: my bad

src/components/Checkbox/index.tsx
matthprost: @chambo-e it should be fine it is the label who's clickable, the input is inside the label

chambo-e (approved):

baptisteArno (approved):

PR #1925, opened 2022-10-04
fix: updated theme documentation and extendTheme function

matthprost:
## Type

- Enhancement

#### What is expected?

Documentation on explanation of the theme composition and how to customize it. It includes customization of typography, shadows, colors, spaces, radii and how to set dark theme. This documentation will be useful for future integration & customization of Scaleway UI in different projects.

I also update `extendTheme` function and make light exported as `theme` instead of `lightTheme` as it is the theme by default we don't need to precise.

Screenshot 2022-10-04 at 15 30 19


chambo-e (commented):
I thought we said we'd remove extendTheme once previous theme was done ? 🤔 

What is the point of allowing theme customization ?

matthprost:
> I thought we said we'd remove extendTheme once previous theme was done ? 🤔
> 
> What is the point of allowing theme customization ?

@chambo-e `extendTheme` is very useful for generating themes even for dark theme we are using it to generate it. The function is also made for any user who wants to customise theme colors or any other property to do it trough this helper.

The goal is still to make the library as much customisable as possible for external users. If we remove this function it will be more difficult to juste change one property in the theme (or a few little). Also this function is made to evolve over time and to provide even simpler ways of customising theme.

This function wasn't using old theme or extending old theme so I'm not sure when you say we would remove it once old theme is replace 😄

Lawndlwd (approved):

vincentaudebert (approved):

baptisteArno (approved):

chambo-e (approved):

PR #1928, opened 2022-10-05
fix: force text align in counter

DorianMaliszewski:
## Type

- Bug

#### What is expected?

Force counter to be centered. In our console other styles override the default behaviour

1. Add text-align: center


matthprost (approved):

chambo-e (approved):

Exayy:
counter text was supported to have a stronger fontWeight no ?

PR #1942, opened 2022-10-11
refactor(menu): improve stories snippet parsing

baptisteArno:
## Summary

No need to parse custom code snippets anymore.

## Type

- Refactor


matthprost (approved):

matthprost (commented):
Screenshot 2022-10-12 at 11 09 42

In the code snippet we have function brackets, is it supposed to be like that ?

baptisteArno:
> In the code snippet we have function brackets, is it supposed to be like that ?

@matthprost yes, I better understand how story parsing is done. Now we never have to write custom snippet.

If you give your `ComponentStory` a function argument, then Storybook will set the source type to `dynamic`. If you don't give an argument, it will set the source type to `code`.

The `dynamic` source type will try to "understand" the inner component. But for complex story it will not render something useful

The `code` source type is just printing out your component as-is, with the function brackets included (I couldn't find a way to ignore those)

More info: https://storybook.js.org/docs/react/writing-docs/doc-block-source

QuiiBz (approved):

PR #1956, opened 2022-10-13
fix(test): add color global values support in css

matthprost:
## Type

- Enhancement

#### What is expected?

Text component now accepts global values for color property: `inherit`, `initial`, etc.

I also fixed SelectNumber story category to be Data Entry.


chambo-e (commented):
I don't understand why we would need this ? 🤔

matthprost:
> I don't understand why we would need this ? 🤔

We had many use case where we needed those properties with text and not set specific color or a color base on parent so to to avoid making:

```tsx
const StyledText = styled(Text)`
   color: inherit;
`

const Component = () => <StyledText />
```

we can directly pass the prop to the component.

chambo-e:
> > I don't understand why we would need this ? 🤔
> 
> We had many use case where we needed those properties with text and not set specific color or a color base on parent so to to avoid making:
> 
> ```tsx
> const StyledText = styled(Text)`
>    color: inherit;
> `
> 
> const Component = () => <StyledText />
> ```
> 
> we can directly pass the prop to the component.

So only inherit I assume ?

Can't we have color:inherit by default on Text and have the color props optional instead ?

matthprost:
> 

Okay sounds another good solution :)

src/components/Text/index.tsx
```diff
@@ -45,13 +46,13 @@ const generateStyles = ({
...
 
-  const themeColor = theme.colors[color]
+  const themeColor = theme.colors[color as Color]
   const text = `text${definedProminence}${
     disabled ? 'Disabled' : ''
   }` as keyof typeof themeColor
 
   return `
-    color: ${theme.colors[color][text]};
+    color: ${color === 'inherit' ? color : theme.colors[color][text]};
```
chambo-e, line 55: We could have 
```suggestion
    color: ${color ? theme.colors[color][text] : 'inherit'};
```

And avoid everything else :)

matthprost: @chambo-e If I do that types won't show that inherit is the default value for color

chambo-e: We don't need to, that's already how css works :)

baptisteArno:
Can we make the `color` prop optional?

baptisteArno (changes requested):
Can we make the `color` prop optional?

matthprost:
> Can we make the `color` prop optional?

That's gonna be a two step migration, we make color required to add neutral to all `Text` then we will remove required and remove all text with color `inherit`

baptisteArno:
What a pain @matthprost

PR #1965, opened 2022-10-18
fix: more display in navtab

DorianMaliszewski:
## Type

- Bug

#### What is expected?

Recompute if we need to display the More tab when children change

1. useEffect add dependency

2. Fix some styles according to our design system

3. Split stories

4. Enable canvas mode in dev 

| /?path=/docs/components-navigation-tabs--playground  | 
Capture d’écran 2022-10-18 à 12 17 57
 | 
Capture d’écran 2022-10-18 à 12 16 28
 |


matthprost (approved):

matthprost (commented):
just a little thing, I dunno if there anything we can do about nothing being displayed in the arg for tab

Screenshot 2022-10-18 at 13 40 05

Exayy (approved):

PR #1968, opened 2022-10-19
feat: new ContainerV2 component

Exayy:
## Type

- Bug
- Feature ✅ 
- Enhancement
- Documentation
- Migration
- Refactor

### Summarise concisely:

New component : ContainerV2, which is the same component than Container (v1) without Box props and margin-top


src/components/ContainerV2/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <Stack gap={1} direction="row" alignItems="center">
+          <Text variant="heading" as="h2">
+            {title}
+          </Text>
+          {subtitle}
+        </Stack>
+        <div>{rightTitle}</div>
+      </Stack>
+    ) : null}
+    {header}
```
chambo-e, line 65: header seems to be used only once which could be replaced by subtitle

Maybe nuke this props ?

matthprost (changes requested):
Can you remove loki update as we are using chromatic ?

src/components/ContainerV2/index.tsx
Exayy: actually `subtitle` hasn't the same positionning it's next (on the right) to the title, header is below

Props are really confusing :
- Title & subtitle are close, mostly subtitle has another font style
- right title sticks to the right
- header : below everything

`header` isn't bad IMO even if it's use only once so far, i've more doubt about `subtitle`

chambo-e (approved):

Lawndlwd (approved):
💜

PR #1973, opened 2022-10-20
docs(Table): fixed table crashing story and updated to new way

matthprost:
## Type

- Documentation

#### What is expected?

Fixed cash on storybook story of table due to last changes. @chambo-e I had to revert to keep function that return a styled, it seems like storybook doesn't want a styled as component in their stories.

Also I updated stories to be formatted correctly.

### DIFF
--- a/src/components/Table/index.tsx
+++ b/src/components/Table/index.tsx
@@ -18,14 +18,16 @@ const Table: ((props: BoxProps) => JSX.Element) & {
   BodyCell: typeof BodyCell
 } = props => <StyledTable {...props} />
 
-export const Head = styled(Box.withComponent('thead'))`
+const StyledHead = styled(Box.withComponent('thead'))`
   border: 0;
   border-bottom-width: 1px;
   border-color: ${({ theme }) => theme.colors.neutral.borderWeak};
   border-style: solid;
 `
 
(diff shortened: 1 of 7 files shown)


src/components/Table/index.tsx
```diff
@@ -18,14 +18,16 @@ const Table: ((props: BoxProps) => JSX.Element) & {
...
-export const Head = styled(Box.withComponent('thead'))`
+const StyledHead = styled(Box.withComponent('thead'))`
   border: 0;
   border-bottom-width: 1px;
   border-color: ${({ theme }) => theme.colors.neutral.borderWeak};
   border-style: solid;
 `
 
-export const Row = styled(Box.withComponent('tr'), {
+export const Head = (props: BoxProps) => <StyledHead {...props} />
```
chambo-e, line 28: What happens if we do 
?
```ts
export const Head = StyledHead 
```

matthprost: @chambo-e Still same issue

chambo-e: Can we raise an issue on emotion or docgen in the meantime ? This is opposite of everything we are trying to do and external tools should not impact our way of doing things :)

baptisteArno (approved):

chambo-e (approved):

PR #1975, opened 2022-10-21
refactor(markdown): remove box from Markdown

baptisteArno:
Remove Box from `MarkDown`

Also rename the component to `Markdown`


chambo-e (approved):

src/components/MarkDown/index.tsx
```diff
@@ -20,12 +19,10 @@ const Container = styled.div`
   padding: ${({ theme }) => theme.space['1']};
 `
 
-const headingRenderer = ({
-  node,
+const MdHeading = ({
```
matthprost, line 22: Possible to use full word instead of acronyms ?

```suggestion
const MarkDownHeading = ({
```

matthprost (changes requested):

chambo-e (approved):

matthprost (approved):

PR #1987, opened 2022-10-24
refactor(BarChart,BarStack): rewrite stories

QuiiBz:
## Type

- Refactor

#### What is expected?

Rewrite stories for BarChart and BarStack. Badge was already done.

### DIFF
--- a/src/components/BarChart/__stories__/PositiveNegative.stories.tsx
+++ b/src/components/BarChart/__stories__/PositiveNegative.stories.tsx
@@ -3,8 +3,9 @@ import { format } from 'date-fns'
 import BarChart from '..'
 import { barChartPositiveNegativeData } from './mockData'
 
-export const PositiveNegative: ComponentStory<typeof BarChart> = () => (
+export const PositiveNegative: ComponentStory<typeof BarChart> = props => (
   <BarChart
+    {...props}
     height={200}
     data={barChartPositiveNegativeData}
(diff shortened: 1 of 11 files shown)


src/components/BarChart/__stories__/PositiveNegative.stories.tsx
```diff
@@ -0,0 +1,40 @@
+import { ComponentStory } from '@storybook/react'
+import { format } from 'date-fns'
+import BarChart from '..'
+import { barChartPositiveNegativeData } from './mockData'
+
+export const PositiveNegative: ComponentStory<typeof BarChart> = () => (
```
matthprost, line 6: Can you add props here to have more beautiful code snippet and remove those:

Screenshot 2022-10-24 at 15 27 57

matthprost (changes requested):

src/components/BarChart/__stories__/PositiveNegative.stories.tsx
matthprost: Should be something like:

```ts
export const PositiveNegative: ComponentStory<typeof BarChart> = props => <BarChart {...props} otherProp={value} ... />
```

matthprost (changes requested):

matthprost (approved):

baptisteArno (approved):
👌

PR #1995, opened 2022-10-26
refactor(DateInput,Description): migrate stories

QuiiBz:
## Type

- Refactor

#### What is expected?

Migrate `DateInput` and `Description` to the new stories format.

`Description`'s code example can't be properly formatted by storybook because it's only made of `styled`


src/components/DateInput/__stories__/Uncontrolled.stories.tsx
```diff
@@ -0,0 +1,14 @@
+import { Story } from '@storybook/react'
+import { ComponentProps } from 'react'
+import DateInput from '..'
+
+export const Uncontrolled: Story<ComponentProps<typeof DateInput>> = props => (
+  <DateInput {...props} onChange={() => {}} label="Date" />
+)
```
matthprost, line 7: Can you remove onChange as it is uncontrolled?

src/components/DateInput/__stories__/Controlled.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { ComponentProps, useState } from 'react'
+import DateInput from '..'
+
+export const Controlled: Story<ComponentProps<typeof DateInput>> = () => {
+  const [value, setValue] = useState<Date | null>(
+    new Date('December 17, 1995 03:24:00'),
+  )
+
+  return <DateInput label="Date" value={value as Date} onChange={setValue} />
+}
```
matthprost, line 11: If we spread props here won't it remove `() =>` in code snippet of storybook?

matthprost (changes requested):

src/components/DateInput/__stories__/Controlled.stories.tsx
QuiiBz: Yes, but it will also remove the `useState` and such the `value/setValue` pair:

Screenshot 2022-10-27 at 10 39 47

IMO it's better to keep it like this.

matthprost: Okay then let's keep it like that 😄

matthprost (approved):

chambo-e (approved):

PR #1997, opened 2022-10-27
feat(Snippet): new component

matthprost:
## Type

- Feature

#### What is expected?

Snippet component to display part of code or commands. There is a single line mode, multiline mode and you can add prefixes (command line or line number). In multiline mode snippet will show a extend button to display all code.

Screenshot 2022-10-27 at 10 38 49
Screenshot 2022-10-27 at 10 38 58
Screenshot 2022-10-27 at 10 39 03
Screenshot 2022-10-27 at 10 39 42

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -105,13 +105,18 @@ const adjustedTheme = (ancestorTheme: Theme, theme: Theme) => ({
   ),
 })
 
-export const globalStyles = (theme: Theme) => css`
-  ${normalize()}
+export const globalStyles = (mode: 'light' | 'dark') => (theme: Theme) =>
+  css`
+    ${normalize()}
 
-  body {
-    color: ${theme.colors.neutral.text};
-  }
-`
+    body {
+      color: ${theme.colors.neutral.text};
+    }
+
+    :root {
+      color-scheme: ${mode === 'dark' ? 'dark' : 'light'};
+    }
+  `
 
 const fonts = css`
   @font-face {
@@ -168,7 +173,7 @@ export const decorators = [
         supportedLocales={['en', 'fr', 'es']}
       >
         <ThemeProvider theme={generatedTheme}>
-          <Global styles={[globalStyles, fonts]} />
+          <Global styles={[globalStyles(mode), fonts]} />
           <StoryComponent />
         </ThemeProvider>
       </I18n>
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -114,7 +114,7 @@ export const globalStyles = (mode: 'light' | 'dark') => (theme: Theme) =>
     }
 
     :root {
-      color-scheme: ${mode === 'dark' ? 'dark' : 'light'};
+      color-scheme: ${mode};
     }
   `
 
--- a/src/components/CopyButton/index.tsx
+++ b/src/components/CopyButton/index.tsx
@@ -44,7 +44,7 @@ const StyledButton = styled('button', {
   }
 `
 
-type CopyButtonProps = {
+export type CopyButtonProps = {
   size?: keyof typeof SIZES
   value: string
   copyText?: string
--- a/src/components/CopyButton/index.tsx
+++ b/src/components/CopyButton/index.tsx
@@ -44,7 +44,7 @@ const StyledButton = styled('button', {
   }
 `
 
-export type CopyButtonProps = {
+type CopyButtonProps = {
   size?: keyof typeof SIZES
   value: string
   copyText?: string
--- a/src/components/Snippet/__tests__/index.tsx
+++ b/src/components/Snippet/__tests__/index.tsx
@@ -1,0 +1,58 @@
+import userEvent from '@testing-library/user-event'
+import {
+  renderWithTheme,
+  shouldMatchEmotionSnapshot,
+} from '../../../helpers/jestHelpers'
+import Snippet from '../index'
+
+const TEST_VALUE = `Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
+`
+
+describe('Snippet', () => {
+  test('renders correctly', () =>
+    shouldMatchEmotionSnapshot(<Snippet value={TEST_VALUE} />))
+
+  test('renders correctly with multiline', () =>
+    shouldMatchEmotionSnapshot(<Snippet value={TEST_VALUE} multiline />))
+
+  test('renders correctly with multiline with prefix lines number', () =>
+    shouldMatchEmotionSnapshot(
+      <Snippet value={TEST_VALUE} multiline prefix="lines" />,
+    ))
+
+  test('renders correctly with multiline with prefix command', () =>
+    shouldMatchEmotionSnapshot(
+      <Snippet value={TEST_VALUE} multiline prefix="command" />,
+    ))
+
+  test('renders correctly with single line with prefix command', () =>
+    shouldMatchEmotionSnapshot(<Snippet value={TEST_VALUE} prefix="command" />))
+
+  test('renders correctly with single line with prefix lines number', () =>
+    shouldMatchEmotionSnapshot(<Snippet value={TEST_VALUE} prefix="lines" />))
+
+  test('renders correctly with copyText', () =>
+    shouldMatchEmotionSnapshot(<Snippet value={TEST_VALUE} copyText="Test" />))
+
+  test('renders correctly with copiedText', () =>
+    shouldMatchEmotionSnapshot(
+      <Snippet value={TEST_VALUE} copiedText="Test" />,
+    ))
+
+  test('renders correctly with hideText', () =>
+    shouldMatchEmotionSnapshot(<Snippet value={TEST_VALUE} hideText="Test" />))
+
+  test('renders correctly with showText', () =>
+    shouldMatchEmotionSnapshot(<Snippet value={TEST_VALUE} showText="Test" />))
+
+  it('should click on extend button to display full content on multiline', async () => {
+    const node = renderWithTheme(<Snippet multiline value={TEST_VALUE} />)
+
+    await userEvent.click(node.getByRole('button', { name: 'Show' }))
+  })
+})
--- a/src/components/Snippet/__tests__/index.tsx
+++ b/src/components/Snippet/__tests__/index.tsx
@@ -1,11 +1,8 @@
 import userEvent from '@testing-library/user-event'
-import {
-  renderWithTheme,
-  shouldMatchEmotionSnapshot,
-} from '../../../helpers/jestHelpers'
+import { shouldMatchEmotionSnapshot } from '../../../helpers/jestHelpers'
 import Snippet from '../index'
 
-const TEST_VALUE = `Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+const TEST_VALUE_MULTILINE = `Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
 Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
 Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
 Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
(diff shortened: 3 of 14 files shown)


PetmePant (approved):

.storybook/preview.tsx
```diff
@@ -105,13 +105,18 @@ const adjustedTheme = (ancestorTheme: Theme, theme: Theme) => ({
...
-  body {
-    color: ${theme.colors.neutral.text};
-  }
-`
+    body {
+      color: ${theme.colors.neutral.text};
+    }
+
+    :root {
+      color-scheme: ${mode === 'dark' ? 'dark' : 'light'};
```
QuiiBz, line 117:
```suggestion
      color-scheme: ${mode};
```

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+  copiedText,
+  multiline = false,
+  showText = 'Show',
+  hideText = 'Hide',
+  prefix,
+}: SnippetProps) => {
+  const theme = useTheme()
+  const [showMore, setShowMore] = useState(false)
+
+  const lines = value.split(/\r\n|\r|\n/).filter(Boolean)
```
QuiiBz, line 158: Could we extract to a const so we can provide more context about what this regex is?

```suggestion
const LINE_BREAK_REGEX = /\r\n|\r|\n/
...
  const lines = value.split(LINE_BREAK_REGEX).filter(Boolean)
```

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+            lines.map((line, index) => (
+              <StyledSpan multiline={multiline} key={line}>
+                {prefix ? (
+                  <Prefix
+                    as="span"
+                    variant="code"
+                    prefix={prefix}
+                    prominence="weak"
+                  >
+                    {prefix === 'lines' ? (index + 1).toString() : '$'}
```
QuiiBz, line 191:
```suggestion
                    {prefix === 'lines' ? `${index + 1}` : '$'}
```

QuiiBz (changes requested):

.storybook/preview.tsx
```diff
@@ -105,13 +105,18 @@ const adjustedTheme = (ancestorTheme: Theme, theme: Theme) => ({
...
-  body {
-    color: ${theme.colors.neutral.text};
-  }
-`
+    body {
+      color: ${theme.colors.neutral.text};
+    }
+
+    :root {
+      color-scheme: ${mode};
```
chambo-e, line 117: Should also support system setting 

https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme#adapting_to_color_schemes

src/components/CopyButton/index.tsx
```diff
@@ -44,7 +44,7 @@ const StyledButton = styled('button', {
   }
 `
 
-type CopyButtonProps = {
+export type CopyButtonProps = {
```
chambo-e, line 47: Use ComponentProps instead

src/components/Snippet/__tests__/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  test('renders correctly with hideText', () =>
+    shouldMatchEmotionSnapshot(<Snippet value={TEST_VALUE} hideText="Test" />))
+
+  test('renders correctly with showText', () =>
+    shouldMatchEmotionSnapshot(<Snippet value={TEST_VALUE} showText="Test" />))
+
+  it('should click on extend button to display full content on multiline', async () => {
+    const node = renderWithTheme(<Snippet multiline value={TEST_VALUE} />)
+
+    await userEvent.click(node.getByRole('button', { name: 'Show' }))
```
chambo-e, line 56: Use transform on shouldMatchEmotionSnapshot

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+  const numberOfLines = lines.length
+  const hasShowMoreButton = numberOfLines > 4 && multiline
+
+  /**
+   * This is used to calculate the max height of the snippet in order to add an animation when the show more button is clicked.
+   * The max height is calculated by multiplying the number of lines by the line height of the text plus the padding (2 times).
+   */
+  const maxHeightExtended =
+    numberOfLines * Number(theme.typography.code.lineHeight.replace('px', '')) +
+    Number(theme.space['2'].replace('px', '')) * 2
```
chambo-e, line 170: Seems hella complicated, could we instead have a height: fit-content and an animation with a keyframes ?

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+  return (
+    <Container multiline={multiline}>
+      <StyledStack>
+        <PreText
+          as="pre"
```
chambo-e, line 176: Why not code 🤔 ?

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+  prefix?: Prefixes
+  showText?: string
+  hideText?: string
+} & Pick<CopyButtonProps, 'copyText' | 'copiedText'>
+
+const Snippet = ({
+  value,
+  copyText,
+  copiedText,
+  multiline = false,
```
chambo-e, line 152: Can't we infer this with the value provided ? 

EDIT: we seems to already do that on line 160-162 🤔

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+   * For `lines` prefix it will display the line number.
+   * For `command` prefix it will display a `$` sign.
+   */
+  prefix?: Prefixes
+  showText?: string
+  hideText?: string
+} & Pick<CopyButtonProps, 'copyText' | 'copiedText'>
+
+const Snippet = ({
+  value,
```
chambo-e, line 149: Why not children ?

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+  copyText,
+  copiedText,
+  multiline = false,
+  showText = 'Show',
+  hideText = 'Hide',
+  prefix,
```
chambo-e, line 155: no default on prefix ?

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+          as="pre"
+          variant="code"
+          multiline={multiline}
+          hasShowMoreButton={hasShowMoreButton}
+          showMore={showMore}
+          maxHeightExtended={maxHeightExtended}
+        >
+          {multiline ? (
+            lines.map((line, index) => (
+              <StyledSpan multiline={multiline} key={line}>
```
chambo-e, line 185:
```suggestion
              <StyledSpan multiline key={line}>
```

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+                    prominence="weak"
+                  >
+                    {prefix === 'lines' ? `${index + 1}` : '$'}
+                  </Prefix>
+                ) : null}
+                {line}
+              </StyledSpan>
+            ))
+          ) : (
+            <StyledSpan multiline={multiline}>
```
chambo-e, line 200:
```suggestion
            <StyledSpan>
```

src/components/Snippet/index.tsx
chambo-e, line 185: What happens if two lines are the same ? Children.map might be a better solution :)

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+              <StyledSpan multiline={multiline} key={line}>
+                {prefix ? (
+                  <Prefix
+                    as="span"
+                    variant="code"
+                    prefix={prefix}
+                    prominence="weak"
+                  >
+                    {prefix === 'lines' ? `${index + 1}` : '$'}
+                  </Prefix>
```
chambo-e, line 194: Might be better to use css counters :)
Meaning user will be able to select text by selection it

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Counter_Styles/Using_CSS_counters

.storybook/preview.tsx
matthprost: @chambo-e we don't have such option on storybook, its or light or dark. Also I don't really see the purpose of supporting it as you change theme on purpose to see component on each individual themes.

src/components/CopyButton/index.tsx
matthprost: To get props from where?

src/components/Snippet/index.tsx
matthprost: @chambo-e Prefix is placed at the begging of each lines but by default we do not want to show any, it has to be specifically written

src/components/Snippet/index.tsx
matthprost: @chambo-e Yes its a good idea

src/components/Snippet/index.tsx
matthprost: @chambo-e `code` is for a short fragment of computer code while `pre` is for pre-formatted which is a style applied to the text. As we can't really know when its code or just information that is copiable, I decided to put `pre` instead of `code`.

src/components/Snippet/index.tsx
matthprost: @chambo-e that's the tricky thing about `height` animation. I cannot use `height` as value for animation I have to use `max-height` so I cannot set `height: fit-content`. I have set animation with `height` from `120px` to `fit-content` here is the result:

https://user-images.githubusercontent.com/15812968/199985245-9b939d00-d280-4b0c-aafb-9ba681024f55.mov

I believe that `auto`, `fit-content`, etc values are excluded from transition and animations.
Here is what I've found about it:
Screenshot 2022-11-04 at 14 38 47
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Transitions/Using_CSS_transitions#defining_transitions

The best way I found is to calculate the number of lines and so the size of the height and sending it to the CSS to be able to make a smooth animation. 

https://user-images.githubusercontent.com/15812968/199985874-b69f13b7-402e-4093-b32c-0482b7beb518.mov

src/components/Snippet/index.tsx
matthprost: I will change it

src/components/Snippet/index.tsx
matthprost: Nice one

src/components/Snippet/index.tsx
matthprost: @chambo-e I changed it, now we put those values into a pseudo element `:before` with CSS counter or `$` text depending on `prefix` property. It also make the JSX smaller and I cloud remove a styled const 🎉

src/components/CopyButton/index.tsx
QuiiBz: I think he mean don't export the type, and when you want to use it in another file, use `ComponentProps<typeof CopyButton>` @matthprost

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+const ButtonContainer = styled('div', {
+  shouldForwardProp: prop => !['multiline'].includes(prop),
+})<{ multiline?: boolean }>`
+  position: absolute;
+  top: 0;
+  right: 0;
+  padding: ${({ theme, multiline }) =>
+    multiline
+      ? `${theme.space['2']} ${theme.space['2']} 0 0`
+      : `${theme.space['1']} ${theme.space['2']} 0 0`};
```
QuiiBz, line 92: I think we can simplify a bit:
```suggestion
      `${theme.space[multiline ? '2' : '1']} ${theme.space['2']} 0 0`
```

QuiiBz (approved):

src/components/CopyButton/index.tsx
matthprost: @QuiiBz Ahh okay makes sense now

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+      : ''};
+`
+
+const ShowMoreContainer = styled('div', {
+  shouldForwardProp: prop => !['showMore'].includes(prop),
+})<{ showMore?: boolean }>`
+  width: 100%;
+  box-shadow: ${({ theme, showMore }) =>
+    !showMore
+      ? `0px -22px 19px -6px
```
chambo-e, line 106: No theme shadow ?

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+  top: 0;
+  right: 0;
+  padding: ${({ theme, multiline }) =>
+    `${theme.space[multiline ? '2' : '1']} ${theme.space['2']} 0 0`};
+  background: ${({ theme }) => theme.colors.neutral.backgroundStrong};
+  border-radius: ${({ theme }) => theme.radii.default};
+  border: 2px solid transparent;
+  ${({ multiline, theme }) =>
+    !multiline
+      ? `box-shadow: -27px 0 19px -11px ${theme.colors.neutral.backgroundStrong}`
```
chambo-e, line 96: No theme shadow ?

src/components/Snippet/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+          />
+        </ButtonContainer>
+        {hasShowMoreButton ? (
+          <ShowMoreContainer showMore={showMore}>
+            <StyledButton type="button" onClick={() => setShowMore(!showMore)}>
+              <AlignCenterText as="span" variant="bodySmallStrong">
+                {showMore ? hideText : showText}
+                &nbsp;
+                <AnimatedArrowIcon name="arrow-down" showMore={showMore} />
+              </AlignCenterText>
```
chambo-e, line 210: (pseudo) 
```jsx
              <Stack>
                <Text>{showMore ? hideText : showText}</Text>
                &nbsp;
                <AnimatedArrowIcon name="arrow-down" showMore={showMore} />
              </Stack>
```

better syntax, less components

chambo-e (changes requested):
Screenshot 2022-11-07 at 09 38 04

This huge difference in margin between the two prefix is by design ?

src/components/Snippet/index.tsx
DorianMaliszewski: Your solution is less semantically correct than the one @matthprost proposed I think

src/components/Snippet/index.tsx
matthprost: @chambo-e this is specific to this component so there is no shadow from theme associated with it

src/components/Snippet/index.tsx
matthprost: @chambo-e same as said before as it is not generic shadow we cannot add it to theme

matthprost:
> Screenshot 2022-11-07 at 09 38 04
> 
> This huge difference in margin between the two prefix is by design?

The huge margin is done on purpose when displaying line number. Why? Because our lines are aligned on right side, meaning we have to set a fixed width (or margin). If we don't fix a width here is what happen:
Screenshot 2022-11-07 at 10 37 54

BUT when setting a fixed width how big should it be? I decided that we may display maximum 9999 lines (which is a lot) so I needed enough space for it:
Screenshot 2022-11-07 at 10 39 54

src/components/Snippet/index.tsx
chambo-e: Beg to differ, not much of semantic difference

```
<span>
  {text}
  <svg />
</span>
```
vs
```
<div>
  <p>{text}</p>
  <svg />
</div>
```

But it'll avoid a styled for nothing, and use our positioning components for what they were meant to do

matthprost: @chambo-e Those are into a Button, Button expect to have phrasing content and cannot have `div` as children
cf https://www.w3.org/TR/2012/WD-html5-20120329/the-button-element.html#the-button-element

That's why I made a styled 😄

chambo-e (approved):

PR #1951, opened 2022-10-12
ci: add chromatic support

matthprost:
## Type

-  CI

#### What is expected?

Added chromatic to the project.

PR #1998, opened 2022-10-27
Revert "ci: add chromatic support (#1951)"

matthprost:
Revert of chromatic.


vincentaudebert:
What's the story with this? Should we dump any visual testing then?

matthprost:
I will close this PR, we can come back here if one day we need to enable visual testing again.

PR #2002, opened 2022-10-31
docs: rewrite `Stack`, `Stepper` and `Tooltip` stories

baptisteArno:
## Type

- Documentation


matthprost (approved):

src/components/Stack/__stories__/AlignItems.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import { Template } from './Template.stories'
+import { coloredChildren } from './helper'
+
+export const AlignItems = Template.bind({})
+
+AlignItems.parameters = {
+  docs: {
+    storyDescription:
+      'prop `alignItems` support every value of css property `align-items`',
```
QuiiBz, line 9: I think we should add `The` at the beginning of each description:

```suggestion
      'The prop `alignItems` supports every value of css property `align-items`',
```

src/components/Tooltip/__stories__/Placement.stories.tsx
```diff
@@ -0,0 +1,32 @@
+import { DecoratorFunction } from '@storybook/addons'
+import { ComponentProps } from 'react'
+import Tooltip from '..'
+
+export const Placement = (props: ComponentProps<typeof Tooltip>) =>
+  ['auto', 'top', 'bottom', 'left', 'right'].map(placement => (
+    <Tooltip
+      {...props}
+      key={placement}
+      placement={placement as ComponentProps<typeof Tooltip>['placement']}
```
QuiiBz, line 10: We should be able to avoid the cast by narrowing the array:

```suggestion
  (['auto', 'top', 'bottom', 'left', 'right'] as const).map(placement => (
    <Tooltip
      {...props}
      key={placement}
      placement={placement}
```

Or we could even type the array properly:
```suggestion
const PLACEMENTS: Array<ComponentProps<typeof Tooltip>['placement']> = ['auto', 'top', 'bottom', 'left', 'right']

  PLACEMENTS.map(placement => (
    <Tooltip
      {...props}
      key={placement}
      placement={placement}
```

QuiiBz (changes requested):

QuiiBz (approved):

PR #2009, opened 2022-11-03
fix(Badge): consistent element sizing across variants

chambo-e:
## Type

- Bug

### Summarise concisely:

Neutral variant has a border while the other doesn't, meaning there would be a shift in size if we switch variants at runtime

### DIFF
--- a/src/components/Badge/index.tsx
+++ b/src/components/Badge/index.tsx
@@ -58,7 +58,8 @@ const generateStyles = ({
       };
       background: ${
         theme.colors[sentiment][background as keyof typeof theme.colors.primary]
-      }
+      };
+      border: 1px solid transparent;
     `,
       }),
       {},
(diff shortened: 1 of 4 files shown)


src/components/Badge/index.tsx
```diff
@@ -58,7 +58,8 @@ const generateStyles = ({
       };
       background: ${
         theme.colors[sentiment][background as keyof typeof theme.colors.primary]
-      }
+      };
+      border: 1px solid transparent;
```
matthprost, line 62: What about we use same color for border? There might be a size different for neutral as it has 1px more into its border

```suggestion
      border: 1px solid ${theme.colors[sentiment][background as keyof typeof theme.colors.primary]};
```

matthprost (changes requested):

QuiiBz (approved):

PR #2018, opened 2022-11-07
fix: radio and checkbox properties and removed reakit usages

matthprost:
## Type

- Enhancement

#### What is expected?

I remove most of `aria-checked` that were not useful + I removed reakit. This impact `SelectableCard` and `SwitchButton`.

### DIFF
--- a/src/components/Checkbox/index.tsx
+++ b/src/components/Checkbox/index.tsx
@@ -7,8 +7,8 @@ import {
   ReactNode,
   forwardRef,
   useCallback,
+  useEffect,
   useId,
-  useMemo,
   useState,
 } from 'react'
 import Loader from '../Loader'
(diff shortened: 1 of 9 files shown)


chambo-e (approved):

src/components/Checkbox/index.tsx
```diff
@@ -185,7 +176,11 @@ type CheckboxProps = Pick<
   checked?: boolean | 'indeterminate'
   className?: string
   ['data-visibility']?: string
-} & Required<Pick<ReakitCheckboxProps, 'onChange'>>
+} & Required<Pick<InputHTMLAttributes<HTMLInputElement>, 'onChange'>> &
```
DorianMaliszewski, line 179: Pas obligé de le mettre en required non ?

src/components/Checkbox/index.tsx
```diff
@@ -208,7 +203,7 @@ const Checkbox = forwardRef(
     ref: ForwardedRef<HTMLInputElement>,
   ) => {
     const hasChildren = !!children
-    const { state, setState } = useCheckboxState({ state: checked })
+    const [state, setState] = useState<boolean | 'indeterminate'>(checked)
```
DorianMaliszewski, line 206: Should we remove this behaviour ? This component doesn't need to handle its own state. The user should use checked prop and onChange to change it ?

src/components/Checkbox/index.tsx
```diff
@@ -247,20 +239,19 @@ const Checkbox = forwardRef(
...
-          <StyledReakitCheckbox
+          <CheckboxInput
+            type="checkbox"
             aria-invalid={!!error}
             aria-describedby={error ? `${computedName}-hint` : undefined}
-            aria-checked={
-              state === 'indeterminate' ? 'mixed' : (state as boolean)
-            }
-            checked={state === 'indeterminate' ? false : (state as boolean)}
+            aria-checked={state === 'indeterminate' ? 'mixed' : undefined}
```
QuiiBz, line 253: I'm not sure about this, but I think it should behave the same as `checked`?
```suggestion
            aria-checked={state === 'indeterminate' ? 'mixed' : state}
```

matthprost: @QuiiBz the best is not to use `aria-checked` as it is for a11y. In that case we need it only for `mixed` state as it is the only way to differentiate for true/false state :)

QuiiBz (approved):

PR #2022, opened 2022-11-09
feat(tooltip): nested scrolls, replaced transition by animation and cleaned code

matthprost:
## Type

- Enhancement

#### What is expected?

I made tooltip position recompute on nested scroll. Now tooltip should always follow his children. I also replace all transition by animation which make it easier to manipulate tooltip when scrolling, transition was making a delay on tooltip position due to `transform` property. With animation, only hover and unhover action trigger an animation.


Lawndlwd (approved):
super ✔️

baptisteArno (approved):
LGTM

PR #2032, opened 2022-11-15
refactor(Button): remove box

Exayy:
## Type

- Refactor

## What is expected?

Remove Box and related props from Button components

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -2,18 +2,16 @@ import { Theme, css } from '@emotion/react'
 import styled from '@emotion/styled'
 import { transparentize } from 'polished'
 import {
+  AnchorHTMLAttributes,
   ButtonHTMLAttributes,
   ComponentProps,
   ElementType,
-  FocusEventHandler,
-  MouseEventHandler,
   ReactNode,
   Ref,
   forwardRef,
   isValidElement,
   useMemo,
 } from 'react'
-import Box, { BoxProps } from '../Box'
 import Icon from '../Icon'
 import Link from '../Link'
 import Loader from '../Loader'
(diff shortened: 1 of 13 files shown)


src/components/Button/index.tsx
```diff
@@ -287,16 +286,16 @@ type StyledButtonProps = {
...
+  Pick<
+    AnchorHTMLAttributes<HTMLAnchorElement>,
+    'href' | 'target' | 'download' | 'rel'
+  >
 
-const StyledButton = styled(Box, {
+const StyledButton = styled('button', {
   shouldForwardProp: prop =>
-    !['action', 'variant', 'extend', 'icon', 'download'].includes(prop),
+    !['action', 'variant', 'extend', 'icon', 'download', 'as'].includes(prop),
```
chambo-e, line 298: why? This should be forwarded to emotion no ?

Exayy: otherwise it goes to html & the `as` isn't applied, feel free to try without on "Download" story but without `a` marker don't replace `button` marker

src/components/Button/index.tsx
```diff
@@ -401,31 +400,28 @@ const FwdButton = ({
...
-  const type = as === 'button' ? elementType : undefined
 
   return (
     <Tooltip id={tooltipBaseId} text={tooltip}>
       <StyledButton
         {...props}
         href={href}
         download={download}
-        ref={innerRef}
+        ref={innerRef as Ref<HTMLButtonElement>}
```
matthprost, line 424: Instead maybe you can type innerRef as `HTMLButtonElement` avoiding this type assertion

matthprost (changes requested):

src/components/Button/index.tsx
Exayy: We could but with href prop it will be `HTMLAnchorElement`  and we will have more typing issue outside the component I guess ? So I can because i'm not a big fan of transform a `button` in a `a` but since the feature exist so far I think it's better to keep it the same way no ?

matthprost: okay I think we would change that when migrating the component fully

matthprost (approved):

PR #2034, opened 2022-11-16
fix(Tags): focus input after onChange

QuiiBz:
## Type

- Bug

#### What is expected?

Focus the input after calling `onChange`

### DIFF
--- a/src/components/Tags/index.tsx
+++ b/src/components/Tags/index.tsx
@@ -101,7 +101,7 @@ type TagsProps = {
   id?: string
   manualInput?: boolean
   name?: string
-  onChange?: (tags: string[]) => void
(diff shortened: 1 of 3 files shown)


src/components/Tags/index.tsx
```diff
@@ -138,7 +138,15 @@ const Tags = ({
       typeof tag === 'object' ? tag?.label : tag,
     )
 
-    onChange?.(changes)
+    const promise = onChange?.(changes)
+
+    if (promise instanceof Promise) {
+      promise
+        .then(() => {
+          inputRef.current?.focus()
```
matthprost, line 146: We need to add some unit testing for this part

matthprost (changes requested):

matthprost (approved):

PR #2061, opened 2022-11-25
feat(toaster): add more options and update stories

matthprost:
## Type

- Enhancement

#### What is expected?

I added more option to toaster container, such as `className`, `position`, `newestOnTop`, `limit` and `id`.
I also updated stories to match our new way of doing.

### DIFF
--- a/src/components/Toaster/index.tsx
+++ b/src/components/Toaster/index.tsx
@@ -110,7 +110,6 @@ const toast = {
 }
 
 type ToastContainerProps = {
-  className?: string
   /**
    * Whether or not to display the newest toast on top.
    * `Default: false`
@@ -124,18 +123,12 @@ type ToastContainerProps = {
    * Position of the toast container
    */
   position?: ToastOptions['position']
-  /**
-   * You can set an id if you have multiple toast containers
-   */
-  id?: string
 }
 
 const ToastContainer = ({
-  className,
   newestOnTop,
   limit,
   position,
-  id,
 }: ToastContainerProps) => {
   const theme = useTheme()
 
@@ -151,9 +144,7 @@ const ToastContainer = ({
             icon={false}
             newestOnTop={newestOnTop}
             limit={limit}
-            className={className}
             position={position}
-            containerId={id}
           />
         )}
       </ClassNames>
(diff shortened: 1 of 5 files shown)


src/components/Toaster/index.tsx
```diff
@@ -122,6 +149,11 @@ const ToastContainer = (): JSX.Element => {
             toastClassName={localCss(styles.toast(theme))}
             autoClose={AUTOCLOSE_DELAY}
             icon={false}
+            newestOnTop={newestOnTop}
+            limit={limit}
+            className={className}
+            position={position}
+            containerId={id}
```
chambo-e, line 156: should be sent by toast rather ?

src/components/Toaster/index.tsx
```diff
@@ -122,6 +149,11 @@ const ToastContainer = (): JSX.Element => {
             toastClassName={localCss(styles.toast(theme))}
             autoClose={AUTOCLOSE_DELAY}
             icon={false}
+            newestOnTop={newestOnTop}
+            limit={limit}
+            className={className}
```
chambo-e, line 154: Why ? 🤔

src/components/Toaster/index.tsx
matthprost: @chambo-e it's the container that manage all those settings

src/components/Toaster/index.tsx
matthprost: @chambo-e why for which one? All of them? I added some options for customizing a little more this component plus we need to allow `className` for each components.

src/components/Toaster/index.tsx
chambo-e: ContainerId is only there if we want multiple Container (which I assume we don't :))
https://fkhadra.github.io/react-toastify/multi-containers

If we want uniq toast we need to use toastId on the toast itself https://fkhadra.github.io/react-toastify/use-a-custom-id

src/components/Toaster/index.tsx
chambo-e: className yeah, why would we need it ?

QuiiBz (approved):

PR #2067, opened 2022-11-29
fix: a11y parameters for text and link

matthprost:
## Type

- Enhancement

#### What is expected?

I added possibility to set and `id` for text to be able to anchor some titles.
I also added `aria-label` for links as apparently links should always have label: https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names#interactive_elements_must_be_labeled

### DIFF
--- a/src/components/Link/index.tsx
+++ b/src/components/Link/index.tsx
@@ -26,6 +26,7 @@ type LinkProps = {
   href: string
   // For react router shouldn't be used directly
   onClick?: MouseEventHandler<HTMLAnchorElement>
+  'aria-label'?: string
 }
 
 const ICON_SIZE = 16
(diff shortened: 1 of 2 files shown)


src/components/Link/index.tsx
```diff
@@ -125,6 +126,7 @@ const Link = forwardRef(
       rel,
       className,
       onClick,
+      'aria-label': ariaLabel,
```
chambo-e, line 129: > for links, buttons and clickable elements, the text content of the element typically provides the label

as children is mandatory that should be enough no ?

matthprost: @chambo-e Here is a good example why it is not enough sometimes:

Screenshot 2022-11-30 at 11 19 13

https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA8

chambo-e: Yeah but we don't do that right ? 🤔

chambo-e (approved):

src/components/Link/index.tsx
matthprost: @chambo-e we actually do with learn more links for example

PR #2069, opened 2022-11-30
fix(website): update component usages

matthprost:
## Type

- Enhancement

#### What is expected?

I updated website with new components and reviewed all design (mobile version included) to make website more beautiful and accessible.

| url  | Screenshot 2022-11-30 at 11-06-55 Scaleway UI | Screenshot 2022-11-30 at 11-06-35 Scaleway UI |


src/theme/index.ts
```diff
@@ -52,16 +52,12 @@ type SCWUITheme = typeof theme & {
...
-  [P in keyof T]?: RecursivePartial<T[P]>
-}
-
 /**
  * Will extend theme with new theme properties
  * @param {SCWUITheme} baseTheme the theme you want to extend from, by default it is set to light theme
  * @param {RecursivePartial<SCWUITheme>} extendedTheme the properties of a new theme you want to apply from baseTheme
  */
-const extendTheme = (extendedTheme: RecursivePartial<SCWUITheme>) =>
+const extendTheme = (extendedTheme: Record<string, unknown>) =>
```
chambo-e, line 60: Why ? 🤔

website/global.d.ts
```diff
@@ -13,5 +13,8 @@ declare global {
 }
 
 declare module '@emotion/react' {
-  export interface Theme extends SCWUITheme {}
+  export interface Theme extends SCWUITheme {
```
chambo-e, line 16: also why ? 🤔

website/src/components/CopyBoxCommand.tsx
```diff
@@ -1,34 +1,26 @@
...
+import { Stack, Tabs } from '@scaleway/ui'
+import { Children, ReactElement, isValidElement, useState } from 'react'
 import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter'
-import { darcula } from 'react-syntax-highlighter/dist/esm/styles/prism'
+import {
+  dracula,
+  oneLight,
+} from 'react-syntax-highlighter/dist/esm/styles/prism'
+
+const StyledDiv = styled(Stack)`
```
chambo-e, line 11:
```suggestion
const StyledStack = styled(Stack)`
```

no ?

src/theme/index.ts
matthprost: @chambo-e We can override the theme with any type of variable. For instance if I want to store `setTheme` in my theme, I need to be able to do it without having type error saying I am not allowed to add it.

website/src/components/CopyBoxCommand.tsx
matthprost: @chambo-e yes you are right I'm changing it

src/theme/index.ts
chambo-e: That's so error prone, our theme should match our predefined scheme, if an user want to set a custom theme he can do it through emotion, not through us

emotion.d.ts
```ts
type MyCustomTheme = {
  setTheme: () => void
}

export interface Theme extends ScwUITheme & MyCustomTheme {}
```tsx

index.tsx
```
import { theme } from '@scaleway/ui'
import { ThemeProvider } from '@emotion/react'

const App = () => {
  const theme: Theme = { ...theme, setTheme: () => { switch() } }

  return (
    <ThemeProvider theme={theme}>{...}</Theme>
  )
}
```

chambo-e (approved):

QuiiBz (approved):

PR #2071, opened 2022-11-30
fix(Pagination): fix getPageNumbers in Pagination v1, add unit tests

ModuloM:
## Type

- Bug

### Summarise concisely:

Fix a bug in `Pagination` v1 that use to display a supernumerary page.

#### What is expected?

Now the pagination correctly display the last page as the max pageCount if smaller than current + gap

| url  | image | image (1) |


ModuloM:
As fixing this comes to implement PaginationV2, I close this one.

PR #2075, opened 2022-12-01
refactor: rewrite stories textbox and timeinput

vincentaudebert:
## Type

- Refactor

#### What is expected?

Rewrite stories for Textbox and TimeInput


src/components/TextBox/__stories__/Template.stories.tsx
```diff
@@ -0,0 +1,6 @@
+import { ComponentStory } from '@storybook/react'
+import UncontrolledTextBox from './UncontrolledTextBox'
+
+export const Template: ComponentStory<typeof UncontrolledTextBox> = args => (
+  <UncontrolledTextBox {...args} />
+)
```
matthprost, line 6: Can you import `UncontrolledTextBox` content in here instead and remove the file? We don't want to have such snippet of code as it doesn't represent how to use the component:

Screenshot 2022-12-01 at 15 39 43

src/components/TextBox/__stories__/Error.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import { Story } from '@storybook/react'
+import ErrorTransition from '../../../__stories__/components/ErrorTransition'
+import UncontrolledTextBox from './UncontrolledTextBox'
+
+export const Error: Story = () => (
+  <ErrorTransition error="An error">
+    {error => <UncontrolledTextBox error={error} placeholder="Placeholder" />}
+  </ErrorTransition>
+)
```
matthprost, line 9: Same here we don't know what is `ErrorTransition` and what does it do while looking at code snippet.

EDIT: After checking it seems like `ErrorTransition` is showing an error after a certain time, we don't need such thing for storybook. Instead you can do this:

```suggestion
export const Error: Story =  Template.bind({})

Error.args = {
  error: 'An error'
}
```

src/components/TimeInput/__stories__/Controlled.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+      {({ value, onChange }) => (
+        <TimeInput
+          name="timeinput-test-controlled"
+          onChange={onChange}
+          placeholder="Time"
+          value={value}
+        />
+      )}
+    </ControlValue>
+)
```
matthprost, line 17: Do not use `ContolValue` but useState instead. It will show a snippet on how to make it work with hook state form react instead of one of our custom components.

src/components/TimeInput/__stories__/Schedule.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import { Story } from '@storybook/react'
+import TimeInput, { schedules } from '..'
+
+export const Schedule: Story = props => (
+  <>
+    {Object.keys(schedules).map(schedule => (
+      <div style={{ marginBottom: '16px' }}>
```
matthprost, line 7: This div should be in a decorator and you can use `Stack` instead:

```tsx
Schedule..decorators = [
  StoryComponent => (
    <Stack gap={2}>
      <StoryComponent />
    </Stack>
  ),
]
```

It will not show `div` into code snippet which will make it easier to read

matthprost (changes requested):

matthprost (changes requested):
We cannot do anything about `function noRefCheck(){}`?

Screenshot 2022-12-05 at 15 27 08

src/components/TimeInput/__stories__/index.stories.tsx
```diff
@@ -14,137 +11,22 @@ export default {
...
+// export { ReadOnly } from './ReadOnly.stories'
+export { Error } from './Error.stories'
+export { Schedule } from './Schedule.stories'
+// export { ToggleablePassword } from './ToggleablePassword.stories'
+// export { Unit } from './Unit.stories'
+// export { Randomize } from './Randomize.stories'
+// export { ForceEdit } from './ForceEdit.stories'
+// export { Multiline } from './Multiline.stories'
+// export { DisableResize } from './DisableResize.stories'
+// export { TabIndex } from './TabIndex.stories'
```
QuiiBz, line 32: Why do we have those comments?

vincentaudebert: oopsie :(

matthprost (approved):

src/components/DateInput/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -2514,7 +2514,7 @@ exports[`DateInput renders correctly with date-fns locale es 1`] = `
                         viewBox="0 0 16 16"
                       >
                         <path
-                          d="M5.29289 8.70735C4.90237 8.31683 4.90237 7.68366 5.29289 7.29314L9.29289 3.29314C9.68342 2.90261 10.3166 2.90261 10.7071 3.29314C11.0976 3.68366 11.0976 4.31683 10.7071 4.70735L7.41421 8.00024L10.7071 11.2931C11.0976 11.6837 11.0976 12.3168 10.7071 12.7074C10.3166 13.0979 9.68342 13.0979 9.29289 12.7074L5.29289 8.70735Z"
+                          d="M5.3 8.7a1 1 0 0 1 0-1.4l4-4a1 1 0 1 1 1.4 1.4L7.42 8l3.3 3.3a1 1 0 0 1-1.42 1.4l-4-4Z"
```
chambo-e, line 2517: Why ? 🤔

vincentaudebert: Fails if I don't update. I don't have any explanation. Sick...

vincentaudebert: Fixed...

QuiiBz (approved):

PR #2079, opened 2022-12-02
refactor(alert): change alert default icon size to 24px

matthprost:
## Type

- Refactor

#### What is expected?

Change default size of icon in Alert from 32px to 24px.

Before:

Screenshot 2022-12-02 at 11 10 19

After:

Screenshot 2022-12-02 at 11 10 23


src/components/Alert/index.tsx
```diff
@@ -117,20 +85,21 @@ type AlertProps = {
...
-    />
+  <StyledStackContainer
+    direction="row"
+    alignItems="center"
+    justifyContent="flex-start"
+    type={type}
+    className={className}
+    gap={2}
+  >
+    <Icon name={icon || typesDefaultIcons[type]} size={24} aria-hidden="true" />
```
chambo-e, line 102: Should we really allow customisation of icon here ? 🤔

matthprost: @chambo-e you're right it is not a good behavior, I will remove it

matthprost: @chambo-e my bad I've read too fast, there might be use case where icon will be different especially for `info` variant

matthprost: example: Screenshot 2022-12-02 at 16 19 41

matthprost: We need to define this with design, I think it will come later on with a bigger refactor for this component.

chambo-e: Gotcha 

(omg that icon is awful)

ModuloM (approved):

PR #2095, opened 2022-12-06
refactor(alert): type to variant prop

matthprost:
## Type

- Refactor

#### What is expected?

Replace `type` prop by `variant` prop in Alert component.


chambo-e (approved):

vincentaudebert (approved):
aaah je dis oui!

PR #2097, opened 2022-12-07
fix: remove box textbox

vincentaudebert:
## Type

- Refactor

#### What is expected?

Removes box from TextBox


matthprost (approved):
> :warning: It will require changes on all repositories including Scaleway Form

chambo-e (approved):

PR #2100, opened 2022-12-07
fix(tabs): vertical scroll triggred on page load

matthprost:
## Type

- Bug

#### What is expected?

There is an issue with Tabs that scroll vertically when loading a page (like on website). This is because we have a scroll event on Tabs to make it go to selected tabs. As we don't want this effect on vertical scroll I blocked it on horizontal scroll avoiding to cause side effects.

Before:

https://user-images.githubusercontent.com/15812968/206156368-25956682-bcce-41c2-a0f0-895ccd5ce953.mov

After:

https://user-images.githubusercontent.com/15812968/206156428-3468d936-4c7b-47c2-8c9d-a635617d5550.mov

### DIFF
--- a/src/components/Tabs/index.tsx
+++ b/src/components/Tabs/index.tsx
@@ -96,13 +96,10 @@ const Tabs = ({
   useEffect(() => {
     const tab = tabsRef.current.querySelector(
       `[role='tab'][aria-selected='true']`,
-    )
-    if (tab && tab.scrollIntoView) {
-      tab.scrollIntoView({
-        behavior: 'smooth',
-        block: 'nearest',
-        inline: 'center',
-      })
+    ) as HTMLElement
+
+    if (tab) {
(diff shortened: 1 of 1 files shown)


src/components/Tabs/index.tsx
```diff
@@ -96,13 +96,10 @@ const Tabs = ({
...
-    if (tab && tab.scrollIntoView) {
-      tab.scrollIntoView({
-        behavior: 'smooth',
-        block: 'nearest',
-        inline: 'center',
-      })
+    ) as HTMLElement
+
+    if (tab && tabsRef.current.scrollTo) {
+      tabsRef.current.scrollTo({ left: tab.offsetLeft, behavior: 'smooth' })
```
chambo-e, line 102: this means tab is useless no ? 🤔

matthprost: useless? Why? tab is used to know where to scroll by using `tab.offsetLeft`

PR #2101, opened 2022-12-07
fix(Checkbox): remove margin when empty

johnrazeur:
## Type

- Bug

### Summarise concisely:

It should not have a margin-right when there is no text provided in the Checkbox component.## Relevant logs and/or screenshots


DorianMaliszewski (approved):

src/components/Checkbox/index.tsx
```diff
@@ -199,6 +209,7 @@ const Checkbox = forwardRef(
       autoFocus = false,
       className,
       'data-visibility': dataVisibility,
+      'aria-label': ariaLabel,
```
chambo-e, line 212: This should probably be an XOR children | aria-label no ?

src/components/Checkbox/index.tsx
```diff
@@ -33,8 +33,11 @@ const PaddedText = styled(Text)`
   padding: ${({ theme }) => `0 ${theme.space['0.5']}`};
 `
 
-const StyledIcon = styled.svg<{ size: number }>`
-  margin-right: ${({ theme }) => theme.space['1']};
+const StyledIcon = styled('svg', {
+  shouldForwardProp: prop => !['hasChildren'].includes(prop),
+})<{ size: number; hasChildren: boolean }>`
+  margin-right: ${({ theme, hasChildren }) =>
```
matthprost, line 39: Instead of using a margin we can use a gap on parent label as we did for radio component. Using gap will also remove usage of `hasChildren` 😄

matthprost (changes requested):

src/components/Checkbox/index.tsx
johnrazeur: Done

src/components/Checkbox/index.tsx
johnrazeur: Done

matthprost (approved):

src/components/Checkbox/index.tsx
```diff
@@ -104,6 +104,7 @@ const CheckboxContainer = styled.label`
   position: relative;
   display: inline-flex;
   align-items: center;
+  gap: ${({ theme }) => theme.space['1']};
```
chambo-e, line 107: Meaning we can remove the margin-right on StyledActivityContainer too :)

chambo-e (changes requested):

src/components/Checkbox/index.tsx
johnrazeur: Good catch !
Fixed.

chambo-e (approved):

PR #2102, opened 2022-12-07
feat: new list component

DorianMaliszewski:
## Type
- Feature

#### What is expected?

New ListV2 Component

1. New Component

## Relevant logs and/or screenshots (not up to date because of feedbacks)

Capture d’écran 2022-12-07 à 18 28 44
Capture d’écran 2022-12-07 à 18 29 11
Capture d’écran 2022-12-07 à 18 30 09
Capture d’écran 2022-12-07 à 18 30 32
Capture d’écran 2022-12-07 à 18 31 10
Capture d’écran 2022-12-07 à 18 31 37
Capture d’écran 2022-12-07 à 18 33 43
Capture d’écran 2022-12-07 à 18 34 03


src/components/ListV2/ListLoadingPlaceholder.tsx
```diff
@@ -0,0 +1,39 @@
...
+  cols?: number
+  className?: string
+}
+
+export const ListLoadingPlaceholder = ({
+  cols = 1,
+  rows = 5,
+  className,
+}: ListLoadingPlaceholderProps) => (
+  <>
```
chambo-e, line 21: useless no ?

src/components/ListV2/ListBody.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from '@emotion/styled'
+import { ReactNode } from 'react'
+
+const StyledTbody = styled.div`
```
chambo-e, line 4: So a Stack ?

src/components/ListV2/ListContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+  children: ReactNode
+  onSelectedIdsChange?: (selectedIds: string[]) => void
+  selectedIds?: string[]
+  data: T[]
+  idKey?: keyof T
+  autoClose?: boolean
+}
+
+export const ListProvider = ({
+  template = 'repeat(12, 1fr)',
```
chambo-e, line 41: Why is this based on a 12 grid system ? 🤔

src/components/ListV2/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { ListHeaderRow } from './ListHeaderRow'
+import { ListHeaders } from './ListHeaders'
+import { ListLoadingPlaceholder } from './ListLoadingPlaceholder'
+import { ListRow } from './ListRow'
+
+const StyledTable = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: ${({ theme }) => theme.space['1']};
+`
```
chambo-e, line 17: Stack too

src/components/ListV2/ListLoadingPlaceholder.tsx
DorianMaliszewski: No because it need to be typed as JSX.Element and not JSX.Element[] so I did this to avoid as JSX.Element

src/components/ListV2/ListBody.tsx
DorianMaliszewski: Should we reuse it ?

matthprost (changes requested):
Disabled row get selected on multi select

Screenshot 2022-12-08 at 10 20 17

src/components/ListV2/ListCell.tsx
```diff
@@ -0,0 +1,47 @@
+import styled from '@emotion/styled'
+import { MouseEventHandler, ReactNode } from 'react'
+
+const StyledCellDiv = styled.div<{ colSpan?: number }>`
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
+  min-height: 60px;
+  ${({ colSpan }) => (colSpan ? `grid-column: span ${colSpan} / span 12;` : '')}
+`
```
matthprost, line 10: Can be a Stack?

src/components/ListV2/ListHeader.tsx
```diff
@@ -0,0 +1,97 @@
...
+}
+
+const ArrowDownIcon = styled(Icon)``
+const ArrowUpIcon = styled(Icon)``
+
+const StyledIconContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  color: ${({ theme }) => theme.colors.neutral.textWeak};
+`
```
matthprost, line 22: Here too you can use Stack directly

src/components/ListV2/ListHeader.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const StyledTh = styled('div', {
```
matthprost, line 24: Might need to be renamed if it is not a Th, also you could use `Text` component and put it as `div` so you won't need to define `font-size`, `font-weight`, etc.

src/components/ListV2/ListRow.tsx
```diff
@@ -0,0 +1,216 @@
...
+  position: relative;
+  display: grid;
+  grid-template-columns: ${({ template }) => template};
+  align-items: center;
+  border: 1px solid ${({ theme }) => theme.colors.neutral.borderWeak};
+  border-radius: ${({ theme }) => theme.radii.default};
+  transition: box-shadow 200ms ease, border-color 200ms ease;
+  background-color: ${({ theme }) => theme.colors.neutral.background};
+  cursor: auto;
+  font-size: 14px;
```
matthprost, line 39: Two possibilities:
1. You get from theme
2. You make a styled of `Text` as `div`

src/components/ListV2/ListRow.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const computedTemplate = isExpandable ? `${template} 25px` : template
+
+  return (
+    <StyledRow
+      className={className}
+      data-disabled={isDisabled ?? false}
+      data-hoverable={isHoverable ?? false}
+      data-highlight={!!isHighlighted || isSelected}
+      data-variant={variant}
+      role={isExpandable ? 'button row' : 'row'}
```
matthprost, line 188: `button row` as role? I don't think you can have multiple roles like that

src/components/ListV2/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+  children,
+  columns,
+  template,
+  selectable,
+  selectedIds,
+  onSelectedIdsChange,
+  data,
+  idKey,
+  isLoading,
+  autoClose,
```
matthprost, line 66: Adding `className` prop on `<Stack role="table"...>` might be interesting if we want to change gap or others properties

matthprost (changes requested):

src/components/ListV2/ListCell.tsx
DorianMaliszewski: No because of colSpan prop is not allowed by Stack

src/components/ListV2/ListHeader.tsx
DorianMaliszewski: I cant because Text doesnt allow onClick prop :/

vincentaudebert:
https://user-images.githubusercontent.com/3360430/206433116-a4a2499a-a716-498e-a294-ccba9023584d.mov

Could be nice to align right the action buttons?

vincentaudebert:
https://user-images.githubusercontent.com/3360430/206433231-8f30740b-ccb0-46e4-a412-702480b0e469.mov

Sortable is broken for me?

And extra note it's a bit disturbing to have the row 2 with the hide arrow, it looks like it's a bug but it's not. 

Maybe make it an extra story ? Not too sure.

src/components/ListV2/ListContext.tsx
```diff
@@ -0,0 +1,107 @@
...
+  setSelectedIds: Dispatch<SetStateAction<string[]>>
+  expandedIds: string[]
+  setExpandedIds: Dispatch<SetStateAction<string[]>>
+  data: T[]
+  idKey: string
+  autoClose: boolean
+  disabledRowsRef: MutableRefObject<string[]>
+}
+
+const ListContext = createContext<ListContextValue>({} as ListContextValue)
```
QuiiBz, line 27:
```suggestion
const ListContext = createContext<ListContextValue | undefined>(undefined)
```
And then `useListContext` should throw an error if the context is undefined (meaning it has been used without being wrapped in a `ListProvider`)

src/components/ListV2/ListHeader.tsx
```diff
@@ -0,0 +1,96 @@
+import styled from '@emotion/styled'
+import { HTMLProps, MouseEvent, ReactNode } from 'react'
+import Icon from '../Icon'
+import Stack from '../Stack'
+
+const getAriaSort = (
+  sort: 'asc' | 'desc' | 'none',
```
QuiiBz, line 7: Could we extract this type to a type and use it here and in `ListHeaderProps`?

src/components/ListV2/ListHeader.tsx
```diff
@@ -0,0 +1,96 @@
...
+): HTMLProps<HTMLDivElement>['aria-sort'] => {
+  if (sort === 'asc') return 'ascending'
+
+  if (sort === 'desc') return 'descending'
+
+  return sort === 'none' ? 'none' : 'other'
+}
+
+const ArrowDownIcon = styled(Icon)``
+const ArrowUpIcon = styled(Icon)``
```
QuiiBz, line 17: Why is this empty?

src/components/ListV2/ListContext.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+type ListContextValue<T = Record<string, unknown>> = {
+  template: string
+  selectable?: boolean
+  selectedIds: string[]
+  setSelectedIds: Dispatch<SetStateAction<string[]>>
+  expandedIds: string[]
+  setExpandedIds: Dispatch<SetStateAction<string[]>>
+  data: T[]
+  idKey: string
```
QuiiBz, line 22: Should be `keyof T[number]` I think?

src/components/ListV2/ListLoadingPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+    {Array.from({ length: rows }, (_, index) => index).map(index => (
```
QuiiBz, line 23: Could we extract those `Array.from` out of the JSX?

src/components/ListV2/ListRow.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  &[role='button row'] {
+    cursor: pointer;
+  }
+
+  ${({ theme }) =>
+    LIST_ROW_VARIANTS.map(
+      color => `
+    &[data-variant="${color}"] {
+      color: ${theme.colors[color as Color].text};
```
QuiiBz, line 74: Why do we have to cast here? I think it can be avoided

src/components/ListV2/ListRow.tsx
```diff
@@ -0,0 +1,306 @@
...
+    }
+  }, [isDisabled, id, disabledRowsRef, checkboxDisabled])
+
+  useEffect(() => {
+    if ((isDisabled || checkboxDisabled) && selectedIds.includes(id)) {
+      setSelectedIds(current => {
+        const indexOfItem = current.indexOf(id)
+        current.splice(indexOfItem, 1)
+
+        return [...current]
```
ModuloM, line 259: Why mutate current array then copy in a new one? If the purpose was to preserve orginal array, it will maybe  be better to do the oppsite ` return [...current].splice(indexOfItem, 1)` or just returning mutated array.

src/components/ListV2/ListRow.tsx
```diff
@@ -0,0 +1,306 @@
...
+      })
+    }
+  }, [id, isDisabled, selectedIds, setSelectedIds, checkboxDisabled])
+
+  return (
+    <Tooltip text={tooltip}>
+      <StyledRow
+        className={className}
+        data-disabled={isDisabled ?? false}
```
ModuloM, line 268: Why using nullish coalescing operator here, instead of default `false` like for `isSelected`?

src/components/ListV2/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+  width?: string
+  id?: string
+}
+
+type ListProps<T> = {
+  children: ReactNode
+  /**
+   * Add checkboxes on the list
+   * */
+  selectable?: boolean
```
ModuloM, line 27: Would be better if the prop name implies that's a boolean?
`isSeletable`?

src/components/ListV2/ListCell.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+type ListCellProps = {
+  colSpan?: number
+  children?: ReactNode
+  onClick?: MouseEventHandler<HTMLTableCellElement>
+  className?: string
+  /**
+   *  Use this if you want to prevent onClick to be handle by parents (Like when you have an expandable content)
+   * */
+  isClickable?: boolean
```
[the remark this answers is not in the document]
ModuloM: And maybe set default value (I always prefer to rely on explicite booleans)?

vincentaudebert (approved):

src/components/ListV2/ListRow.tsx
DorianMaliszewski: splice return an array of removed elements not the remaining ones

src/components/ListV2/ListHeader.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+const StyledIconContainer = styled(Stack)`
+  color: ${({ theme }) => theme.colors.neutral.textWeak};
+`
+
+const StyledHeader = styled('div', {
+  shouldForwardProp: prop => prop !== 'sortable',
```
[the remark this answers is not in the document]
DorianMaliszewski: colSpan is an html attribute that to the browser that the cell is taking two columns imo

ModuloM (approved):

matthprost (approved):

chambo-e (approved):

src/components/ListV2/ListContext.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export const useListContext = <
+  T extends Record<string, unknown> = Record<string, unknown>,
+>() => {
+  const context = useContext(ListContext) as ListContextValue<T>
+  if (!context) {
+    throw new Error('useListContext should be used inside a List component')
+  }
+
+  return context
```
QuiiBz, line 115: We should cast at the `return`, because context can be undefined

Lawndlwd (approved):

QuiiBz (approved):

PR #2108, opened 2022-12-08
fix(icon): image needs to be labelled

matthprost:
## Type

- Enhancement

#### What is expected?

All images (including svg) should have a label associated with it to be accessible: https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names#content_with_images_must_be_labeled

### DIFF
--- a/src/components/Alert/__tests__/__snapshots__/index.test.tsx.snap
+++ b/src/components/Alert/__tests__/__snapshots__/index.test.tsx.snap
@@ -80,6 +80,7 @@ exports[`Alert renders correctly unknow theme type and fallback to defaults 1`] 
     class="cache-bhu4eo-Stack-StyledStackContainer-alertStyles eenhiht2"
   >
     <svg
+      aria-label="circle"
       class="cache-wp3con-StyledIcon-sizeStyles etwatq50"
       viewBox="0 0 24 24"
     >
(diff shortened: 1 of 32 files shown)


chambo-e (commented):
Soooooo, I kinda disagree (again sorry)

According to the link you sent it's for interactive elements, which a bunch of our icons are not :)

Might be better to add aria-label to Menu/Nav/Button/... maybe ? With proper label 

Because filling a label by an icon name is not much better than no a11y :)

matthprost:
> Soooooo, I kinda disagree (again sorry)
> 
> According to the link you sent it's for interactive elements, which a bunch of our icons are not :)
> 
> Might be better to add aria-label to Menu/Nav/Button/... maybe ? With proper label
> 
> Because filling a label by an icon name is not much better than no a11y :)

@chambo-e you may be right I was refering to "[Content with images must be labeled](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names#content_with_images_must_be_labeled:~:text=This%20includes%20SVG%20images)" but as you mention it if we already have a label on a link + a label on the icon it will be too much

matthprost:
> Soooooo, I kinda disagree (again sorry)
> 
> According to the link you sent it's for interactive elements, which a bunch of our icons are not :)
> 
> Might be better to add aria-label to Menu/Nav/Button/... maybe ? With proper label
> 
> Because filling a label by an icon name is not much better than no a11y :)

@chambo-e after quick research I still think it is needed to describe the icon it self to text to speech, @DorianMaliszewski also confirm this is the good way of doing

PR #2111, opened 2022-12-09
fix: typing of new list

DorianMaliszewski:
## Type

- Bug

#### What is expected?

Typing was wrong dont need to extend Record<string, unknown> instead just check on idKey that T is extending it

1. Improve typing

2.

3.

### DIFF
--- a/src/components/ListV2/ListContext.tsx
+++ b/src/components/ListV2/ListContext.tsx
@@ -10,7 +10,7 @@ import {
   useRef,
   useState,
 } from 'react'
-import { ListDataObject } from './types'
+import type { ListDataObject } from './types'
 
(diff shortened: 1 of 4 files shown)


chambo-e (approved):

src/components/ListV2/ListContext.tsx
```diff
@@ -10,36 +10,37 @@ import {
   useRef,
   useState,
 } from 'react'
+import { ListDataObject } from './types'
```
QuiiBz, line 13:
```suggestion
import type { ListDataObject } from './types'
```

matthprost (approved):

QuiiBz (approved):

PR #2117, opened 2022-12-12
fix: remove box list

vincentaudebert:
## Type

- Refactor

#### What is expected?

Removes box from list

### DIFF
--- a/src/components/ActionBar/index.tsx
+++ b/src/components/ActionBar/index.tsx
@@ -17,6 +17,8 @@ const StyledDiv = styled.div<{ rank: number }>`
   transform: translate(-50%, 0);
   width: 600px;
   z-index: 2;
+  display: flex;
+  align-items: center;
 `
 
 type ActionBarProps = {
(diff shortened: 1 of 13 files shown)


src/components/ActionBar/index.tsx
```diff
@@ -17,6 +17,8 @@ const StyledDiv = styled.div<{ rank: number }>`
   transform: translate(-50%, 0);
   width: 600px;
   z-index: 2;
+  display: flex;
+  align-items: center;
```
chambo-e, line 21: intended ?

src/components/List/index.tsx
```diff
@@ -56,6 +55,8 @@ export type ListBodyRenderProps<DataType> = {
 
 type BodyProps<DataType> = {
   children: (props: ListBodyRenderProps<DataType>) => JSX.Element
+  className?: string
+  style?: React.CSSProperties
```
matthprost, line 59: We don't do `React.XX` anymore, you can import `CSSProperties` directly

matthprost (changes requested):

src/components/ActionBar/index.tsx
vincentaudebert: yes select bar is broken otherwise :/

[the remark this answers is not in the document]
vincentaudebert: good point. fixed

chambo-e (approved):

philibea (approved):

matthprost (approved):

PR #2119, opened 2022-12-12
fix: rerender loop

DorianMaliszewski:
## Type

- Bug

#### What is expected?

There was a rerender loop by when having both onSelectedIdsChange and selectedIds prop. Instead it need to act like all html input : when there are onChange present use it instead of local onChange

1. Wrap the setSelectedIds in a function that choose between both of them

2.

3.

### DIFF
--- a/src/components/ListV2/ListContext.tsx
+++ b/src/components/ListV2/ListContext.tsx
@@ -58,7 +58,7 @@ export const ListProvider = <T,>({
 
   useEffect(() => {
     setSelectedIds(current =>
-      selectedIdsProp && current && current !== selectedIdsProp
+      selectedIdsProp && current !== selectedIdsProp
         ? selectedIdsProp
         : current,
     )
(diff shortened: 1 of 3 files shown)


matthprost (approved):

src/components/ListV2/ListContext.tsx
```diff
@@ -78,14 +73,27 @@ export const ListProvider = <T,>({
...
 
+  const onChangeSelectedIds: Dispatch<SetStateAction<string[]>> = useCallback(
+    (value: string[] | ((currentIds: string[]) => string[])) => {
+      if (onSelectedIdsChangeRef.current) {
+        onSelectedIdsChangeRef.current(
+          typeof value !== 'function' ? value : value(selectedIds),
+        )
+      } else {
+        setSelectedIds(typeof value !== 'function' ? value : value(selectedIds))
+      }
```
QuiiBz, line 84: To avoid duplication:
```suggestion
      const newSelectedIds = typeof value !== 'function' ? value : value(selectedIds)
      
      if (onSelectedIdsChangeRef.current) {
        onSelectedIdsChangeRef.current(newSelectedIds)
      } else {
        setSelectedIds(newSelectedIds)
      }
```

QuiiBz (approved):

PR #2125, opened 2022-12-15
fix(Tags): revert focusing input onChange

QuiiBz:
## Summary

Revert https://github.com/scaleway/scaleway-ui/pull/2034/

It's causing issues when `onChange` is triggered: the input is always focused and we can't escape it


matthprost:
Can we have explanation?

matthprost (approved):

Lawndlwd (approved):

PR #2129, opened 2022-12-16
fix(label): update label properties

matthprost:
## Type

- Enhancement

#### What is expected?

I updated Label component in order to follow tokens.

### DIFF
--- a/src/components/Text/index.tsx
+++ b/src/components/Text/index.tsx
@@ -1,6 +1,6 @@
 import type { Theme } from '@emotion/react'
 import styled from '@emotion/styled'
-import type { ElementType, ReactNode } from 'react'
+import type { ReactNode } from 'react'
 import { useEffect, useRef, useState } from 'react'
 import recursivelyGetChildrenString from '../../helpers/recursivelyGetChildrenString'
 import type { Color } from '../../theme'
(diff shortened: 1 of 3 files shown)


src/components/Text/index.tsx
```diff
@@ -81,40 +81,42 @@ type TextProps = {
...
+    disabled: boolean
+    italic: boolean
+    underline: boolean
+  }>(generateStyles)
 
 const Text = ({
   variant,
   children,
-  as,
+  as = 'div',
```
chambo-e, line 119: I assumed we always wanted to specify it ?

src/components/Text/index.tsx
```diff
@@ -136,9 +139,11 @@ const Text = ({
     }
   }, [oneLine])
 
+  const StyledTextComponent = StyledText(as)
```
chambo-e, line 142: Wuut ? That is really ugly, and kinda dangerous this

No way to have it work with the native styled as ?

matthprost: @chambo-e The problem with native one is that component doesn't accept any others props that the ones of the default tag (here it as a `div`) so `htmlFor` won't be accepted as it doesn't exist in div properties. That's the only workaround I've found for this :/

src/components/Text/index.tsx
matthprost: @chambo-e You're right I was confused by control value in storybook

philibea (approved):

vincentaudebert (approved):

PR #2136, opened 2022-12-19
fix(Button): remove Link animation when using href

QuiiBz:
## Type

- Bug

#### What is expected?

Fix `Button` hover bug when using `href` and `target="_blank"`

Before:

https://user-images.githubusercontent.com/43268759/208443774-930c1fb1-1528-4294-afd7-1cd522a32a44.mov

After:

https://user-images.githubusercontent.com/43268759/208443641-36a78a08-c452-4bc4-bef9-53372ac94eb0.mov

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -24,13 +24,13 @@ const StyledContent = styled.div`
   pointer-events: none;
 `
 
-// It will remove animation between Link and left icon that is implemented into Link component
+// Remove the animation of the Link
 const StyledLink = styled(Link)`
-  gap: 0;
-
   &:hover,
   &:focus {
-    gap: 0;
+    svg {
+      transform: none !important;
+    }
   }
 `
 
--- a/src/components/Button/__tests__/__snapshots__/index.tsx.snap
+++ b/src/components/Button/__tests__/__snapshots__/index.tsx.snap
@@ -344,7 +344,7 @@ exports[`Button should render correctly loading with an icon 1`] = `
 
 exports[`Button should render correctly when acting as Link 1`] = `
 <DocumentFragment>
-  .cache-ktsz49-StyledLink-StyledLink-StyledButton {
+  .cache-1rri44g-StyledLink-StyledLink-StyledButton {
   background-color: transparent;
   border: none;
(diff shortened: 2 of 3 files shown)


src/components/Button/index.tsx
```diff
@@ -24,13 +24,13 @@ const StyledContent = styled.div`
...
 
-// It will remove animation between Link and left icon that is implemented into Link component
+// Remove the animation of the Link
 const StyledLink = styled(Link)`
-  gap: 0;
-
   &:hover,
   &:focus {
-    gap: 0;
+    svg {
```
chambo-e, line 31: Can we target the Icon directly ?

QuiiBz: It works if we export and use the `StyledIcon` from the `Icon` component, but not if we use the `Icon` component directly (maybe because it's wrapped in a `forwardRef`?)

Screenshot 2022-12-20 at 09 55 41
Screenshot 2022-12-20 at 09 55 51

matthprost (approved):

src/components/Button/index.tsx
chambo-e: What about 

```jsx
const StyledIcon = styled(Icon)``

const StyledLink = styled(Link)`
  &:hover,
  &:focus {
    ${StyledIcon} {
      transform: none;
    }
  }
`
```

QuiiBz: Just tested and it doesn't work, maybe because the style of this empty `StyledIcon` isn't the same as the one inside `Link`

chambo-e: Well you'll also need to use StyledIcon in Link 😄

QuiiBz: Oh I didn't saw the difference between the styled of icon and styled of svg. Using a `styled.svg` in both places still doesn't work btw

philibea (approved):

chambo-e (approved):

src/components/Button/index.tsx
QuiiBz: As discussed we'll keep the `svg`

PR #2138, opened 2022-12-19
refactor: remove ErrorTransition and ControlValue from stories

matthprost:
## Type

- Refactor

#### What is expected?

Remove `ErrorTransition` that is no more used and `ControlValue` from left stories.


chambo-e (commented):
The only place left where ControlValue is used is in ListV2, can we remove it also ?

vincentaudebert:
Why do we keep control value if we should not use it?

QuiiBz:
If we remove it from the ListV2 test, we can also remove completely the component because it won't have any usage left.

QuiiBz (approved):

chambo-e (approved):

PR #2148, opened 2022-12-28
fix(useOnFieldChange): change order of params condition and callback

philibea:
## Type

- Refactor

#### What is expected?

Change the order of the condition parms and callback parms and add condition = true 

```
 useOnFieldChange(
          'name',
         condition,
          callback,    
        ),
```

```
 useOnFieldChange(
          'name',
          callback,
          condition
        ),
```

### DIFF
--- a/packages/form/src/hooks/useOnFieldChange.ts
+++ b/packages/form/src/hooks/useOnFieldChange.ts
@@ -9,7 +9,7 @@ type CallbackFn<FieldValue, AllValues> = (
 export const useOnFieldChange = <FieldValue = unknown, AllValues = unknown>(
   name: string,
   callback: CallbackFn<FieldValue, AllValues>,
-  condition = true,
+  enabled = true,
 ): void => {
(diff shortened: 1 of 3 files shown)


chambo-e (approved):

DorianMaliszewski (approved):

packages/form/src/hooks/useOnFieldChange.ts
```diff
@@ -8,8 +8,8 @@ type CallbackFn<FieldValue, AllValues> = (
 
 export const useOnFieldChange = <FieldValue = unknown, AllValues = unknown>(
   name: string,
-  condition: boolean,
   callback: CallbackFn<FieldValue, AllValues>,
+  condition = true,
```
johnrazeur, line 12: It's maybe better to name this `enabled` ?

philibea: right !

johnrazeur (approved):

PR #2154, opened 2022-12-29
fix(Button): proper layout on extend buttons

chambo-e:
## Type

- Enhancement

Before

https://user-images.githubusercontent.com/2802867/209998029-67fce1f9-fc2b-4f2d-872f-62fab8925e09.mp4

After

https://user-images.githubusercontent.com/2802867/209998052-f0a030d6-0c0e-463e-a5f4-392a2e04e816.mp4

### DIFF
--- a/packages/ui/src/components/Button/__stories__/Extend.stories.tsx
+++ b/packages/ui/src/components/Button/__stories__/Extend.stories.tsx
@@ -1,12 +1,75 @@
-import { Template } from './Template.stories'
+import type { ComponentStory } from '@storybook/react'
+import { Button } from '..'
+import { Stack } from '../../Stack'
 
-export const Extend = Template.bind({})
+const Buttons: ComponentStory<typeof Button> = ({ ...props }) => (
+  <>
+    <div>
+      <Button icon="plus" iconSize={24} extend size="large" {...props}>
+        Normal Button
+      </Button>
+    </div>
+    <div>
+      <Button icon="plus" iconSize={24} extend size="medium" {...props}>
+        Normal Button
+      </Button>
+    </div>
+    <div>
+      <Button icon="plus" iconSize={24} extend size="small" {...props}>
+        Normal Button
+      </Button>
+    </div>
+    <div>
+      <Button icon="plus" iconSize={24} extend size="xsmall" {...props}>
+        Normal Button
+      </Button>
+    </div>
+    <div>
+      <Button icon="plus" iconSize={24} extend size="xxsmall" {...props}>
+        Normal Button
+      </Button>
+    </div>
+    <div>
+      <Button
+        icon="plus"
+        iconSize={24}
+        extend
+        as="a"
+        href="https://scaleway.com"
+        {...props}
+      >
+        Link Button
+      </Button>
+    </div>
+    <div>
+      <Button
+        icon="plus"
+        iconSize={24}
+        extend
+        as="a"
+        href="https://scaleway.com"
+        size="small"
+        {...props}
+      >
+        Link Button
+      </Button>
+    </div>
+  </>
+)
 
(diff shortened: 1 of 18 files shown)


packages/ui/src/components/Button/index.tsx
```diff
@@ -257,19 +255,9 @@ const SmartIcon = ({
...
 
-type StyledIcon = {
-  margin: number
-  position?: 'left' | 'right'
-}
-
-const StyledIconContainer = styled('div', {
-  shouldForwardProp: prop => !['margin', 'position'].includes(prop),
-})<StyledIcon>`
+const StyledIconContainer = styled('div')`
```
matthprost, line 258:
```suggestion
const StyledIconContainer = styled.div`
```

Not much difference it is just shorter

packages/ui/src/components/Button/__stories__/Extend.stories.tsx
```diff
@@ -1,11 +1,75 @@
...
+      <Buttons {...props} />
+    </Stack>
+    <Stack gap={1} alignItems="center">
+      <Buttons {...props} />
+    </Stack>
+    <Stack gap={1} alignItems="end">
+      <Buttons {...props} />
+    </Stack>
+  </Stack>
+)
```
matthprost, line 72: I'm afraid you cannot use this method as example in storybook, it doesn't show properties on `Button` component itself but on the `Buttons` you created.

Screenshot 2023-01-03 at 10 34 22

Instead I will recommend something shorter but more clear:
```ts
export const Extend: ComponentStory<typeof Button> = ({ ...props }) => (
  <Stack gap={1} style={{ width: '100%' }}>
    <Stack gap={1}>
      <Button icon="plus" iconSize={24} extend size="medium" {...props}>
        Normal Button
      </Button>
    </Stack>
    <Stack gap={1} alignItems="center">
      <Button icon="plus" iconSize={24} extend size="medium" {...props}>
        Normal Button
      </Button>
    </Stack>
    <Stack gap={1} alignItems="end">
      <Button icon="plus" iconSize={24} extend size="medium" {...props}>
        Normal Button
      </Button>
    </Stack>
  </Stack>
)
```

This way the example will be filly shown in code snippet, it is also shorter so easier to read.

matthprost (changes requested):

packages/ui/src/components/Button/index.tsx
```diff
@@ -315,48 +304,43 @@ const StyledButton = styled('button', {
...
     css`
-      display: inline-flex;
-      & ${StyledContent} {
-        transition: max-width 450ms ease, padding 150ms ease, margin 150ms ease;
+      & ${InhibitedStack} {
+        transition: max-width 250ms ease, margin-left 250ms ease;
         max-width: 0;
-        margin-right: 0;
-        ${icon ? 'padding-right: 0;' : 'padding-left: 0;'};
+        margin-left: -${theme.space['1']}px; // To counteract parent gap
```
QuiiBz, line 337: Are these comments valid in emotion? Because a normal CSS comment should be:
```suggestion
        margin-left: -${theme.space['1']}px; /* To counteract parent gap */
```

packages/ui/src/components/Button/index.tsx
```diff
@@ -369,7 +353,7 @@ const StyledButton = styled('button', {
       flex-shrink: 0;
       ${action === 'rounded' && `border-radius: 16px;`}
       > svg {
-        // safari issue prevent event propgation
+        /* safari issue prevent event propgation */
```
QuiiBz, line 356:
```suggestion
        /* safari issue prevent event propagation */
```

QuiiBz (approved):

vincentaudebert (approved):

PR #2161, opened 2023-01-02
chore: update `engines` versions in `package.json`

QuiiBz:
## Type

- Chore

#### What is expected?

Update `engines` versions in `package.json` to match our other repositories.


package.json
```diff
@@ -7,8 +7,8 @@
   ],
   "packageManager": "pnpm@7.20.0",
   "engines": {
-    "node": ">=18.x",
-    "pnpm": ">=7.0.0"
+    "node": ">=v18.12.1",
+    "pnpm": ">=7.20.0"
```
chambo-e, line 11: This is a rather very strict policy, are we sure we need this ?

QuiiBz: This is what we have in all our other projects. We also already set the `packageManager` field to `pnpm@7.20.0`, so if we want to allow broader versions we should also update it.

package.json
```diff
@@ -7,8 +7,8 @@
   ],
   "packageManager": "pnpm@7.20.0",
   "engines": {
-    "node": ">=18.x",
-    "pnpm": ">=7.0.0"
+    "node": ">=v18.12.1",
```
matthprost, line 10: Why should we force for this specific version instead of `v18.x` ?

matthprost (changes requested):

package.json
QuiiBz: See https://github.com/scaleway/scaleway-ui/pull/2161#discussion_r1060101653

QuiiBz:
@chambo-e @matthprost I've updated the versions to `major.x` as discussed.

package.json
QuiiBz: Update: https://github.com/scaleway/scaleway-ui/pull/2161#issuecomment-1369775010

package.json
QuiiBz: Update: https://github.com/scaleway/scaleway-ui/pull/2161#issuecomment-1369775010

chambo-e (approved):

matthprost (approved):

PR #2165, opened 2023-01-03
fix(actionBar): remove padding and center content from list action bar

matthprost:
## Type

- Bug

#### What is expected?

ActionBar is having a padding of 16px inside of the component while the component shouldn't impose any padding it should be done in children. I removed it and added it into List, it should also fix the alignement issue present on every ActionBar.

| url  | Screenshot 2023-01-03 at 15 07 08 | Screenshot 2023-01-03 at 15 31 42 |

### DIFF
--- a/packages/ui/src/components/ActionBar/__stories__/Ranks.stories.tsx
+++ b/packages/ui/src/components/ActionBar/__stories__/Ranks.stories.tsx
@@ -1,13 +1,24 @@
+import styled from '@emotion/styled'
 import type { ComponentStory } from '@storybook/react'
 import ActionBar from '..'
+import Stack from '../../Stack'
+
+const FullHeightStack = styled(Stack)`
+  height: 100%;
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/ActionBar/__stories__/Ranks.stories.tsx
```diff
@@ -1,13 +1,24 @@
...
+const FullHeightStack = styled(Stack)`
+  height: 100%;
+  padding: 0 ${({ theme }) => theme.space['2']};
+`
 
 export const Ranks: ComponentStory<typeof ActionBar> = props => (
   <>
     <ActionBar {...props} rank={1}>
-      I am an Action Bar with rank 1
+      <FullHeightStack alignItems="center" direction="row">
```
QuiiBz, line 14: Why do we use a `Stack` if we only have a single children?

QuiiBz (approved):

chambo-e (approved):

PR #2166, opened 2023-01-03
fix(TextBox): show required icon with right components

QuiiBz:
## Type

- Bug

#### What is expected?

Allow having a required icon + a right component (randomize, password toggle...)

New stories:

Screenshot 2023-01-03 at 16 30 47

Old ones not impacted:

Screenshot 2023-01-03 at 15 44 04


matthprost (changes requested):
Can you make them all with same width?

Screenshot 2023-01-03 at 15 51 38

matthprost (changes requested):
Screenshot 2023-01-05 at 10 34 03

I was wondering what happen when you have required, randomize and unit. It seems like the right part is broken.

QuiiBz:
> I was wondering what happen when you have required, randomize and unit. It seems like the right part is broken.

Fixed, now we can have as many right components as we want.

matthprost (approved):

chambo-e (approved):

PR #2192, opened 2023-01-10
fix(Text): show tooltip on ellipsis only when needed

matthprost:
## Type

- Bug

#### What is expected?

When using `oneLine` prop in Text it was using tooltip on hover even when not needed. I fixed it.

Before:

https://user-images.githubusercontent.com/15812968/211562821-fc998b9d-85cf-4aad-b0e8-dd1b6edc04e5.mov

After:

https://user-images.githubusercontent.com/15812968/211562889-440eae96-a8ae-48d2-8ac7-999a365ed0be.mov

### DIFF
--- a/packages/ui/src/components/Text/index.tsx
+++ b/packages/ui/src/components/Text/index.tsx
@@ -132,7 +132,7 @@ const Text = ({
   useEffect(() => {
     if (oneLine && elementRef && elementRef.current) {
       const { offsetWidth, scrollWidth } = elementRef.current
-      setIsTruncated(offsetWidth <= scrollWidth)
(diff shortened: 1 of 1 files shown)


QuiiBz:
I think we should also be able to always have a tooltip, even when the text isn't truncated.

matthprost:
> I think we should also be able to always have a tooltip, even when the text isn't truncated.

@QuiiBz Why is that? If you need a tooltip you can wrap your `Text` component into a `Tooltip` 😄

QuiiBz:
Yes never mind you're right :)

QuiiBz (approved):

chambo-e (approved):

PR #2199, opened 2023-01-13
fix(Toggle): use gap instead of margin for `label` content

QuiiBz:
## Type

- Bug

#### What is expected?

Fix `<Toggle />` not having spacing when not using a `string` as a `label`

Use a `gap` in the parent instead of using `margin-left` / `margin-right`

### DIFF
--- a/packages/ui/src/components/Toggle/index.tsx
+++ b/packages/ui/src/components/Toggle/index.tsx
@@ -100,6 +100,7 @@ const StyledLabel = styled.label<{
   size: 'small' | 'large'
 }>`
   display: flex;
+  gap: ${({ theme }) => theme.space['1']};
   align-items: center;
   width: fit-content;
   cursor: pointer;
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/Toggle/index.tsx
```diff
@@ -114,30 +115,6 @@ const StyledLabel = styled.label<{
...
-    labelPosition === 'left'
-      ? `margin-right: ${theme.space['1']}`
-      : `margin-left: ${theme.space['1']}`}
-`
-
-const LabelContent = ({
-  label,
-  labelPosition,
-}: Required<Pick<ToggleProps, 'label' | 'labelPosition'>>) => {
-  if (typeof label === 'string') {
```
chambo-e, line 130: Why lose this handling of string vs Component ?

QuiiBz: Because we always want to have a gap between the Toggle itself and the label.

chambo-e: This can still be the case while keeping the handling of string vs component no ?

QuiiBz: @matthprost do we want to always have a gap between the toggle and label? If yes, handling string differently from components becomes useless

chambo-e: I don't understand, either way we are going to have a gap between them


```tsx
<Toggle label="sample label" />
```

will render a proper span with a gap of 8 inside the Toggle

```tsx
<Toggle label={<Text as="span" variant="body">sample label</Text>} />
```

Will render the same gap with a custom component 


I fail to understand why the LabelContent is related to the gap ? :)

QuiiBz: Resolving since as explained we won't care about making a difference between string vs components, as we always want a gap between the toggle and the label.

matthprost (approved):

chambo-e (approved):

PR #2205, opened 2023-01-18
fix(List): update List usages

fabienhebert:
## Type

- Enhancement

### Summarise concisely:

Update list usage about expandable row and selectable

**Selectable rows :**
- Remove the <List> properties :  `idKey` and `data` which was use only for selecting allRow, instead each row selectable row register to the context
- Remove the <List> property `isSelectable`, to know if we can select row, instead we detect if props `selectedIds` and `onSelecedIdsChange` are provided or not

** Expandable rows :**
- No more `isExpandable` on `List.Row`, instead List.Expandable register to the listContext
- `isExpanded` to control/force the expanding of a `List.Row` is replaced by `forceExpand` on the related `List.Expandable`
- Bug fix: if a row can be extended by clicking, a column is created in header to fix headerCell / bodyCell alignment
- Bug fix: the last cell wasn't properly identified some type, so the last cell available space was sometime too high
- Remove `hideArrow` 
- Force `isHoverable` is a row can be expanded

No visual change is expected, except that "expand arrow" is no more display when we forceExpand


matthprost (changes requested):
There is an issue with the expandable arrow, it is not closing when row is closed

https://user-images.githubusercontent.com/15812968/213134972-98f14013-1fd6-4cdf-bc0d-14d3c1388ffb.mov

packages/ui/src/components/ListV2/__stories__/Selectable.stories.tsx
```diff
@@ -1,7 +1,84 @@
...
+              {selectedPlanets.map(planet => planet.name).join(', ')}
+            </Text>
+            <Button
+              action
+              icon="delete"
+              onClick={() => onDelete(selectedIds)}
+            />
+          </Stack>
+        </ActionBar>
+      ) : null}
```
matthprost, line 74: Can we add a height and a padding in order to have a clear action bar:

Screenshot 2023-01-18 at 10 34 48

```suggestion
          <Stack
            direction="row"
            justifyContent="space-between"
            alignItems="center"
            gap={2}
            style={{ height: 100%, padding: '0 8px'}}
          >
            <Text as="p" variant="bodySmall">
              Planets selected :{' '}
              {selectedPlanets.map(planet => planet.name).join(', ')}
            </Text>
            <Button
              action
              icon="delete"
              onClick={() => onDelete(selectedIds)}
            />
          </Stack>
        </ActionBar>
      ) : null}
```

packages/ui/src/components/ListV2/__stories__/Perf.stories.tsx
```diff
@@ -0,0 +1,79 @@
+import type { Story } from '@storybook/react'
+import { useMemo, useState } from 'react'
+import { List } from '..'
+import { Button, PaginationV2, Stack } from '../..'
+
+const ITEM_PER_PAGE = 50
```
matthprost, line 6: Can we reduce to 25 per pages? Or is it on purpose to show performance? This seems quite a long story

packages/ui/src/components/ListV2/__stories__/Perf.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const Perf: Story = () => {
```
matthprost, line 8: Could we call the story `Performance` instead? To avoid having a reduced naming

matthprost (changes requested):

packages/ui/src/components/ListV2/index.tsx
```diff
@@ -47,50 +30,59 @@ type ListProps<T> = {
...
   columns,
   template,
-  isSelectable,
   selectedIds,
   onSelectedIdsChange,
-  data,
-  idKey,
   isLoading,
   autoClose,
+  showExpandArrow = false,
```
matthprost, line 57: Shouldn't `showExpandArrow` be defined by the list itself when there is an Expandable children?

packages/ui/src/components/ListV2/ListExpandable.tsx
```diff
@@ -1,28 +1,51 @@
...
+  &[aria-expanded='true'] {
+    max-height: 9999px;
+  }
 `
 
 type ListExpandableProps = {
   children: ReactNode
   className?: string
+  relatedRowId?: string // Actually always defined because cloned by List.Row
+  forceExpand?: boolean
```
matthprost, line 32: could we get a little comment about this prop? So it is shown in storybook

matthprost (changes requested):

packages/ui/src/components/ListV2/__stories__/Perf.stories.tsx
fabienhebert: yes I think we can delete the story itself before merginig, it was more for checking perfs

packages/ui/src/components/ListV2/index.tsx
DorianMaliszewski: Acutally we doesn't know if all our expandable rows will have an arrow, may be @fabienhebert talk about that with @PetmePant  ? 🤔

PR #2207, opened 2023-01-18
refactor(components): rename TextBox and TextBoxField to TextInput an…

BABAK0T0:
TextInput and TextInputField

## Summary
Renaming TextBox and TextBoxField to TextInput and TextInputField

## Type

- Enhancement

#### What is expected?

Now we can use TextInput and TextInputField in order to be more consistent with our codebase


packages/form/src/components/DateField/__tests__/index.tsx
```diff
@@ -32,6 +32,7 @@ describe('DateField', () => {
       {
         transform: async node => {
           const select = node.getByRole('textbox')
+          console.log(select)
```
chambo-e, line 35: 💩

chambo-e (approved):

packages/form/src/components/DateField/__tests__/index.tsx
BABAK0T0: oops 😅

matthprost (approved):

PR #2209, opened 2023-01-19
feat(tooltip): improve tooltip behavior and fix some issues

matthprost:
## Type

- Enhancement

#### What is expected?

There was multiple issues with tooltip I fixed them and improved the opening / closing behavior:

- Fix: tooltip not closing on hover of disabled radio button (linked to: https://github.com/facebook/react/issues/18753)
- Fix: tooltip appearing at the top left of the screen
- Feat: added a debounce timer on tooltip avoid to flickering when hovering and unhovering fast

https://user-images.githubusercontent.com/15812968/213479952-19ad7f35-03ff-4e9c-b3ef-b3828836806d.mov

### DIFF
--- a/packages/ui/src/components/Tooltip/index.tsx
+++ b/packages/ui/src/components/Tooltip/index.tsx
@@ -15,6 +15,8 @@ import { ARROW_WIDTH, DEFAULT_POSITIONS, computePositions } from './helpers'
 
 const ANIMATION_DURATION = 230 // in ms
 
+function noop() {}
+
 const animation = (positions: PositionsType) => keyframes`
   0% {
     opacity: 0;
@@ -229,6 +231,10 @@ export const Tooltip = ({
 
     return () => {
       window.removeEventListener('scroll', onScrollDetected, true)
+      if (timer.current) {
+        clearTimeout(timer.current)
+        timer.current = undefined
+      }
     }
   }, [generatePositions, onScrollDetected, visibleInDom])
 
@@ -248,10 +254,10 @@ export const Tooltip = ({
   const renderChildren = useCallback(() => {
     if (typeof children === 'function') {
       return children({
-        onBlur: !isControlled ? onPointerEvent(false) : () => {},
-        onFocus: !isControlled ? onPointerEvent(true) : () => {},
-        onPointerEnter: !isControlled ? onPointerEvent(true) : () => {},
-        onPointerLeave: !isControlled ? onPointerEvent(false) : () => {},
+        onBlur: !isControlled ? onPointerEvent(false) : noop,
+        onFocus: !isControlled ? onPointerEvent(true) : noop,
+        onPointerEnter: !isControlled ? onPointerEvent(true) : noop,
+        onPointerLeave: !isControlled ? onPointerEvent(false) : noop,
         ref: childrenRef,
       })
     }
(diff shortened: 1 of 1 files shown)


ModuloM (approved):

packages/ui/src/components/Tooltip/index.tsx
```diff
@@ -204,42 +226,49 @@ const Tooltip = ({
...
-    visibleInDom,
-    positions.tooltipPosition,
-    generatePositions,
-    placement,
-    text,
-    onScrollDetected,
-  ])
+
+    return () => {
+      window.removeEventListener('scroll', onScrollDetected, true)
```
QuiiBz, line 231: Shouldn't we also clear the timeout here?

packages/ui/src/components/Tooltip/index.tsx
```diff
@@ -204,42 +226,49 @@ const Tooltip = ({
...
    * Will render children conditionally if children is a function or not.
    */
   const renderChildren = useCallback(() => {
     if (typeof children === 'function') {
       return children({
-        onBlur: onMouseEvent(false),
-        onFocus: onMouseEvent(true),
-        onMouseEnter: onMouseEvent(true),
-        onMouseLeave: onMouseEvent(false),
+        onBlur: !isControlled ? onPointerEvent(false) : () => {},
```
QuiiBz, line 251: Let's extract all of these empty functions to a top-level `noop` function to avoid creating them every time

```suggestion
function noop() {}
...
        onBlur: !isControlled ? onPointerEvent(false) : noop,
```

QuiiBz (changes requested):

packages/ui/src/components/Tooltip/index.tsx
```diff
@@ -164,6 +174,7 @@ const Tooltip = ({
   const unmountTooltip = useCallback(() => {
     setVisibleInDom(false)
     setReverseAnimation(false)
+    timer.current = undefined
```
ModuloM, line 177: Should we also set debounceTimer.current to undefined ?

packages/ui/src/components/Tooltip/index.tsx
matthprost: Which timeout? This one is just a fallback if the tooltip wrapper is being removed from the dom unexpectedly, we don't want to have an opened listener

QuiiBz: I'm talking about `timer.current`, but not sure if it's necessary or not

packages/ui/src/components/Tooltip/index.tsx
matthprost: @ModuloM `unmountTooltip()` function is being called once debounce timer is done:

```tsx
  const unmountTooltip = useCallback(() => {
    setVisibleInDom(false)
    setReverseAnimation(false)
    timer.current = undefined

    window.removeEventListener('scroll', onScrollDetected, true)
  }, [onScrollDetected])

  const onPointerEvent = useCallback(
    (isVisible: boolean) => () => {
      if (!isVisible && tooltipRef.current && !debounceTimer.current) {
        debounceTimer.current = setTimeout(() => { // <== DEBOUNCE TIMEOUT IS SET HERE
          setReverseAnimation(true)
          timer.current = setTimeout(() => unmountTooltip(), ANIMATION_DURATION) // <=== AND UNMOUNT IS CALLED RIGHT HERE
        }, 200)
      } else {
        if (timer.current) {
          setReverseAnimation(false)
          clearTimeout(timer.current)
          timer.current = undefined
        }
        if (debounceTimer.current) {
          clearTimeout(debounceTimer.current)
          debounceTimer.current = undefined
        }
        setVisibleInDom(isVisible)
      }
    },
    [unmountTooltip],
  )
```

packages/ui/src/components/Tooltip/index.tsx
matthprost: @QuiiBz we could remove it there can be a case where time is still set and the component is being unmounted, good point

packages/ui/src/components/Tooltip/index.tsx
```diff
@@ -172,19 +185,30 @@ const Tooltip = ({
...
       } else {
-        // If a timeout is already set it means tooltip didn't have time to close completely and be removed from dom,
-        // so we clear timeout and set back opacity of tooltip to 1, so it can be visible on screen.
+        // This condition is for when we want to mount the tooltip
+        // If the timer exists it means the tooltip was about to umount, but we hovered the children again,
+        // so we clear the timer and the tooltip will not be unmounted
         if (timer.current) {
           setReverseAnimation(false)
           clearTimeout(timer.current)
+          timer.current = undefined
```
philibea, line 205: can you add a test unit on this line ?

QuiiBz (approved):

PR #2219, opened 2023-01-24
feat(Text): add `dir` prop

QuiiBz:
## Type

- Feature

#### What is expected?

Add a new `dir` prop to `<Text />` to change the direction of the text: https://developer.mozilla.org/fr/docs/Web/HTML/Global_attributes/dir

Screenshot 2023-01-24 at 15 38 24


packages/ui/src/components/Text/index.tsx
```diff
@@ -87,6 +87,7 @@ type TextProps = {
   italic?: boolean
   underline?: boolean
   id?: string
+  dir?: string
```
matthprost, line 90: can't you use the type of `dir` from `p` or `div`?

matthprost (changes requested):

matthprost (approved):

BABAK0T0 (approved):

vincentaudebert:
not sure to understand clearly what you mean by "the direction of the ... " ?

QuiiBz:
@vincentaudebert it's when you use the `oneLine` prop, the text is truncated with '...' at the end

PR #2247, opened 2023-02-06
feat: split theme into another package

matthprost:
## Type

- Enhancement

#### What is expected?

In order to be future proof we need to split themes into another package. From now on we only have one version of a theme which is main one but later design is planning to have multiple theme for multiple projects. Scaleway UI don't need to implement all those theme, in fact each project should pick needed theme from `@scaleway/themes`.


packages/themes/src/index.ts
```diff
@@ -0,0 +1 @@
+export { consoleLightTheme, consoleDarkTheme } from './themes'
```
chambo-e, line 1: Why call that consoleThemes ? scw-ui should not be linked to shire right ?

packages/ui/src/theme/index.ts
```diff
@@ -1,6 +1,5 @@
+import { consoleDarkTheme, consoleLightTheme } from '@scaleway/themes'
```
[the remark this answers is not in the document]
matthprost: @chambo-e From now on it doesn't make much sense as we only have 2 themes available. In a soon future, design wants to make 2 themes per product (documentation, website, etc). All those themes won't be handle in Scaleway UI, so in order to avoid pollution we need to put themes in another package.

This way, documentation could install Scaleway UI, import documentation light and dark theme and inject it into their ThemeProvider.

packages/themes/src/index.ts
matthprost: @chambo-e Right, maybe I can call it otherwise, like general theme or productTheme

scripts/figma-synchronise-tokens.sh
```diff
@@ -92,12 +92,13 @@ for i in "${themes[@]}"
...
+   capitalizeTheme="$(tr '[:lower:]' '[:upper:]' <<< ${i:0:1})${i:1}"
 
    # Generate colors tokens
    (printf "/**
  * Provides all tokens of a specific theme edited by design team.
  * This file is automatically generated from /scripts/figma-synchronise-tokens.sh.
  * PLEASE DO NOT EDIT HERE
  */
-export default "; echo "${FINAL_RESULT}") > packages/ui/src/theme/tokens/"${i}".ts
+export const console"${capitalizeTheme}"Theme = "; echo "${FINAL_RESULT}") > packages/themes/src/themes/console/"${i}".ts
```
philibea, line 103: actually this script can handle only console theme ?

matthprost: @philibea yep from now I have no way to detect what product is targeted for each theme as there is only console. Later the script will be updated to be able to do the same for any themes.

ModuloM (approved):

vincentaudebert (approved):

PR #2255, opened 2023-02-10
fix: correct non scalar default arguments as props

chambo-e:
First PR to prepare for upcoming eslint config update

### DIFF
--- a/packages/form/src/components/DateField/index.tsx
+++ b/packages/form/src/components/DateField/index.tsx
@@ -26,8 +26,8 @@ type DateFieldProps = BaseFieldProps<Date> &
     required?: boolean
     locale?: string
     onChange?: (value: Date | null) => void
-    onBlur?: (event: FocusEvent<HTMLElement, Element>) => void
-    onFocus?: (value: FocusEvent<HTMLElement, Element>) => void
+    onBlur?: (event: FocusEvent<HTMLElement>) => void
+    onFocus?: (value: FocusEvent<HTMLElement>) => void
     autoFocus?: boolean
   }
 
@@ -101,11 +101,11 @@ export const DateField = ({
           input.onChange(newDate)
         }
       }}
-      onBlur={(e: FocusEvent<HTMLElement, Element>) => {
+      onBlur={(e: FocusEvent<HTMLElement>) => {
         input.onBlur(e)
         onBlur?.(e)
       }}
-      onFocus={(e: FocusEvent<HTMLElement, Element>) => {
+      onFocus={(e: FocusEvent<HTMLElement>) => {
         input.onFocus(e)
         onFocus?.(e)
       }}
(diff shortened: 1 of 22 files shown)


packages/ui/src/components/LineChart/index.tsx
```diff
@@ -30,22 +30,27 @@ type LineChartProps = {
   chartProps?: Partial<LineSvgProps>
 }
 
+const defaultMargin = { bottom: 50, left: 60, right: 25, top: 50 }
```
matthprost, line 33: Can all constant be UPPERCASE? As we use to do everywhere else

matthprost (changes requested):

vincentaudebert (approved):

packages/form/src/components/DateField/index.tsx
```diff
@@ -101,11 +101,11 @@ export const DateField = ({
           input.onChange(newDate)
         }
       }}
-      onBlur={(e: FocusEvent<HTMLElement, Element>) => {
+      onBlur={(e: FocusEvent<HTMLElement>) => {
```
QuiiBz, line 104: I believe we shouldn't have to type again the parameters

packages/ui/src/components/BarChart/index.tsx
```diff
@@ -29,13 +29,18 @@ type BarChartProps = {
   chartProps?: Partial<BarSvgProps<BarDatum>>
 }
 
+const defaultMargin = { bottom: 50, left: 60, right: 25, top: 50 }
```
QuiiBz, line 32: Can you put these constants in uppercase?

QuiiBz (changes requested):

packages/form/src/components/DateField/index.tsx
chambo-e: @QuiiBz 😢 

Screenshot 2023-03-02 at 15 49 45

packages/ui/src/components/BarChart/index.tsx
```diff
@@ -29,16 +29,21 @@ type BarChartProps = {
   chartProps?: Partial<BarSvgProps<BarDatum>>
 }
 
+const DEFAULT_MARGIN = { bottom: 50, left: 60, right: 25, top: 50 }
+const DEFAULT_DATA: BarDatum[] = []
+const DEFAULT_AXISFORMATTER = {}
+const DEFAULT_KEYS = ['value']
```
QuiiBz, line 35: Could we use the type from the props?
```suggestion
const DEFAULT_MARGIN: BarChartProps['margin'] = { bottom: 50, left: 60, right: 25, top: 50 }
const DEFAULT_DATA: BarChartProps['data'] = []
const DEFAULT_AXISFORMATTER: : BarChartProps['axisFormatters'] = {}
const DEFAULT_KEYS: : BarChartProps['keys'] = ['value']
```

packages/ui/src/components/DateInput/index.tsx
```diff
@@ -165,11 +165,14 @@ type DateInputProps = Pick<
   value?: Date | string
 }
 
+const DEFAULT_FORMAT = (value?: Date | string) =>
```
QuiiBz, line 168: Same here
```suggestion
const DEFAULT_FORMAT: DateInputProps['format'] = (value) =>
```

matthprost (approved):

PR #2257, opened 2023-02-13
fix(toggle): add forward ref

johnrazeur:
## Type

- Bug

Add `forwardRef` to the Toggle component.

### DIFF
--- a/packages/ui/src/components/Toggle/index.tsx
+++ b/packages/ui/src/components/Toggle/index.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
-import type { ChangeEvent, ChangeEventHandler, ReactNode } from 'react'
-import { useCallback, useEffect, useState } from 'react'
+import { forwardRef, useCallback, useEffect, useState } from 'react'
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Toggle/index.tsx
```diff
@@ -140,66 +140,76 @@ type ToggleProps = {
...
+      name,
+      onChange,
+      size = 'large',
+      tooltip,
+      labelPosition = 'right',
+      label,
+      required,
+      className,
+    }: ToggleProps,
+    ref: Ref<HTMLInputElement>,
```
matthprost, line 158: What is the use case for it?

matthprost (changes requested):

packages/ui/src/components/Toggle/index.tsx
johnrazeur: Currently we can't use `Toggle` in the `Modal` disclosure prop without creating a callback.

matthprost (approved):

QuiiBz (approved):

PR #2269, opened 2023-02-21
fix(NumberInput): primary color on hover on disabled state

matthprost:
## Type

- Bug

#### What is expected?

When hovering `NumberInput` with disabled state we show primary border. It shouldn't.

| Capture d’écran 2023-02-20 à 16 02 09 (1)  |  | Screenshot 2023-02-21 at 11 32 15 |

### DIFF
--- a/packages/ui/src/components/NumberInput/index.tsx
+++ b/packages/ui/src/components/NumberInput/index.tsx
@@ -1,5 +1,4 @@
 import type { Theme } from '@emotion/react'
-import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 import type {
   ChangeEventHandler,
@@ -22,14 +21,7 @@ const roundStep = (value: number, step: number, direction: number) =>
     ? Math.floor(value / step) * step
     : Math.ceil(value / step) * step
 
-const disabledStyles = ({
-  disabled,
-  theme,
-}: {
-  disabled: boolean
-  theme: Theme
-}) =>
-  disabled &&
+const disabledStyles = ({ theme }: { theme: Theme }) =>
   `
     background-color: ${theme.colors.neutral.backgroundDisabled};
     border: none;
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/NumberInput/index.tsx
```diff
@@ -270,7 +257,7 @@ export const NumberInput = ({
   const isPlusDisabled = (maxValue && plusRoundedValue > maxValue) || disabled
 
   return (
-    <StyledContainer disabled={disabled} size={size} className={className}>
+    <StyledContainer data-disabled={disabled} size={size} className={className}>
```
fabienhebert, line 260: Optional : We may use `aria-disabled` for that kind of case mb and still use the attibute as css identifier

packages/ui/src/components/NumberInput/index.tsx
```diff
@@ -291,6 +278,7 @@ export const NumberInput = ({
         }}
         aria-live="assertive"
         role="status"
+        data-disabled={disabled}
```
fabienhebert, line 281: Optional too :
Do we need to duplicate the prop instead of something like 

const StyledContainer = styled`
   &[data-disabled=true] ${StyledCenterBox} {
       border: none;
   }
`

fabienhebert (approved):

packages/ui/src/components/NumberInput/index.tsx
ModuloM: I'm not expert, far from it, but it seems that it's better to keep `data-` as `aria-disabled` will disabled all the children, and apparently it's better to keep it for leaves elements (div with button role, etc.)
(I've used https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled for ref)

ModuloM (approved):

packages/ui/src/components/NumberInput/index.tsx
matthprost: @fabienhebert @ModuloM In fact, we shouldn't use `aria-disabled` it should be explicit from native tags such as `button`, `input`, etc by using prop `disabled`. Here we just want to say to the styled div that the container is disabled, so I passed a `data-disabled`

PR #2270, opened 2023-02-21
feat: update listV2 usages

fabienhebert:
## Type

- Enhancement

#### To understand changes
The main difference between ListV1 and ListV2 is the removal of render prop for List elements

```js
// V1
<List>
   {({ list }) =>
        <list.Row>
   )}
   
 // V2
<List>
    <List.Row>
```
   
##### About Expandable rows
1. Instead of `List.Expandable`, we now have an `expandable` prop (the positionning of expandable content to let be authorized a random placing among the cell)
2. `expandable` is only render when row is expanded (necessary if data fetch or heavy ops are done inside the expandable content
3. No more `forceExpand` usage, it seems overkill at the moment but can be easily restore if needed in the future

##### About Selectable rows
1. Like on V1, doesn't expose a `selectedIds` & `onSelectedIdsChange` but keep it as internal state.
2. Instead of having `data-key` and `idKey` on List and `id` on ListRow,  ListV2 now only use List.Row `id` to register its "selectability" (used for select all)
3. Since there is no more render Props  `SelectBar` must be a bit different. Now `List.SelectBar` is an `utility` accepting `data` and  `idKey` to compute selected data items using the selected row ids.
```js
// V1
<List data={data} idKey="id">
   {({ list }) =>
        <list.SelectBar>
             {({ selectedItems, unselectAll }) => ...
   )}
   
 // new V2
<List>
    <List.SelectBar data={data} idKey="id" >
         {({ selectedItems, unselectAll }) => ...
```
4. `isSelectable` is now named `areRowsSelectable`

##### About Sorting
1.  On this version the sorting is more basic but should be enought and simple to understand.
```js
 // V1
 <List 
    onSortClick={({ field, order}) => { ... }} // order was the new order
    columns={{
        sort : "name" // the field
    }} >

 // current V2
  <ListV2
    columns={{
        sort : asc | desc | none | undefined
        id : "name"
        onClick : ({ columnId, order  } => { ... } // order was the current order
    }} >

 // new V2
 <ListV2
    columns={{
        isOrdered : true/false (say if orderDirection should be considered)
        orderDirection : "asc" | "desc"
        onOrder : (newOrder) => { ... }
    }} >
```

##### Other modifications
1. `autoClose` is now named `autoCollapse`
2. No more padding in cell (first & last), so column width is always the one asked (if space available)
3. Fixing colums alignment with expandableContent
5. Prevent column with auto size to reduce specified width columns
6. No more JSX headers
7. Stories improvements, lot of explanations about props and more realistic example
8. Row prop `variant` has been renamed `sentiment`
9. The border/separator between cells and expandable is now the same than row border (if you specify a sentiment)
10. No more `isHoverable` and `isHighlighted` on Row (diff with V2) 
11. No more expand arrow indicator (diff with V2, like V1) 😞 
12. Select/Unselect all is now disabled if there is no selectable rows

##### Update since MR beginning
1. Sorting -> Ordering : a more flexible system


matthprost (changes requested):
A gap is missing between header and row + the header has a padding left and right that shouldn't have
Screenshot 2023-02-21 at 12 13 39
Screenshot 2023-02-21 at 12 13 44

matthprost (changes requested):
Screenshot 2023-02-21 at 12 15 48
Screenshot 2023-02-21 at 12 15 55

There is too much spacing between the arrows of sorting

packages/ui/src/components/ListV2/HeaderRow.tsx
```diff
@@ -0,0 +1,47 @@
+import styled from '@emotion/styled'
+import type { ReactNode } from 'react'
+import { Checkbox } from '../Checkbox'
+import { HeaderCell } from './HeaderCell'
+import { useListContext } from './ListContext'
+
+const StyledHeaderRow = styled.div`
+  // List itself also apply style about common templating between HeaderRow and other Rows
```
QuiiBz, line 8: Not sure if `//` comments work inside emotion?
```suggestion
  /* List itself also apply style about common templating between HeaderRow and other Rows */
```

packages/ui/src/components/ListV2/__stories__/Context.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <List.Cell>{planet.perihelion}AU</List.Cell>
+          <List.Cell>{planet.aphelion}AU</List.Cell>
+        </List.Row>
+      ))}
+      <SubComponent srcData={data} />
+    </List>
+  )
+}
+
+// Context.parameters = {
```
QuiiBz, line 41: Can we delete this comment?

packages/ui/src/components/ListV2/Row.tsx
```diff
@@ -0,0 +1,197 @@
...
+  children: ReactNode
+  id: string
+  expandable?: ReactNode
+  isSelectDisabled?: boolean
+  selectTooltip?: string
+  isDisabled?: boolean
+  sentiment?: (typeof SENTIMENTS)[number]
+}
+
+export const Row = ({
```
johnrazeur, line 96: I think we should forward ref at least in the component and maybe in others.
I have already have an use case where I needed to attach the ref to manage drag & drop.

fabienhebert: Good idea, i add it on List, List.Row & List.Cell, make less sense on List.SelectBar

matthprost (changes requested):
There is a little issue with mouse cursor on loading and disabled state.
- For loading state the mouse cursor is only in loading on the skeleton line while it should be on the whole list row
- For the disabled state we should have a disabled cursor

packages/ui/src/components/ListV2/HeaderRow.tsx
```diff
@@ -0,0 +1,47 @@
...
+  /* List itself also apply style about common templating between HeaderRow and other Rows */
+  column-gap: ${({ theme }) => theme.space['2']};
+  padding: 0 ${({ theme }) => theme.space['2']};
+`
+
+const StyledCheckbox = styled(Checkbox)`
+  &[aria-disabled='true'] {
+    pointer-events: none;
+  }
+`
```
matthprost, line 17: Please avoid pointer event none, if we need to put a tooltip for some reason it won't work. What was the purpose of this?

packages/ui/src/components/ListV2/Row.tsx
```diff
@@ -0,0 +1,206 @@
...
+export const StyledRow = styled.div`
+  /* List itself also apply style about common templating between HeaderRow and other Rows */
+
+  position: relative;
+  border: 1px solid ${({ theme }) => theme.colors.neutral.borderWeak};
+  border-radius: ${({ theme }) => theme.radii.default};
+  transition: box-shadow 200ms ease, border-color 200ms ease;
+  box-shadow: none;
+  background-color: ${({ theme }) => theme.colors.neutral.background};
+  cursor: auto;
```
matthprost, line 27: It's already the initial value, is there any use for it?

Screenshot 2023-02-22 at 11 30 29

packages/ui/src/components/ListV2/Row.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+const StyledCheckboxContainer = styled.div`
+  display: flex;
+`
+
+const StyledCheckbox = styled(Checkbox)`
+  &[aria-disabled='true'] {
+    pointer-events: none;
+  }
+`
```
matthprost, line 84: Same here, we don't need pointer event, the checkbox should manage itself

packages/ui/src/components/ListV2/Row.tsx
```diff
@@ -0,0 +1,206 @@
...
+                  value={id}
+                  onChange={() => {
+                    if (selectedRowIds[id]) {
+                      unselectRow(id)
+                    } else {
+                      selectRow(id)
+                    }
+                  }}
+                  disabled={isDisabled || isSelectDisabled}
+                  aria-disabled={isDisabled || isSelectDisabled}
```
matthprost, line 191: Here there is an a11y issue, if `disabled` is given you don't need `aria-disabled` it is explicit that the input is disabled

matthprost (changes requested):

packages/ui/src/components/ListV2/HeaderRow.tsx
fabienhebert: was similar in the current ListV2 (ListheaderRow), i change

packages/ui/src/components/ListV2/Row.tsx
fabienhebert: was similar in the current ListV2 (ListRow), i change

packages/ui/src/components/ListV2/Row.tsx
fabienhebert: was similar in the current ListV2 (Row), i change

matthprost (approved):

ModuloM (approved):

PR #2271, opened 2023-02-22
fix: use radii from theme

matthprost:
## Type

- Enhancement

#### What is expected?

Most of radius are hardcoded while we have them in theme.

### DIFF
--- a/packages/themes/src/themes/console/index.ts
+++ b/packages/themes/src/themes/console/index.ts
@@ -6,6 +6,7 @@ const radii = {
   small: '1px',
   default: '4px',
   large: '8px',
+  xlarge: '16px',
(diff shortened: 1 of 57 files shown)


packages/themes/src/themes/console/index.ts
```diff
@@ -6,6 +6,7 @@ const radii = {
   small: '1px',
   default: '4px',
   large: '8px',
+  xlarge: '16px',
```
fabienhebert, line 9: not automatically done by importing ? (mb it will ?)

fabienhebert (approved):

Lawndlwd (approved):

QuiiBz (approved):

PR #2274, opened 2023-02-23
feat(Icon): new icon colors from theme

matthprost:
## Type

- Enhancement

#### What is expected?

- Use new `theme.colors.xx.icon`
- Update Icon component to only take real color as sentiment and no more hexadecimal values
- Added `disabled` to Icon component
- Updated all components using Icon

### DIFF
--- a/packages/ui/src/components/DateInput/index.tsx
+++ b/packages/ui/src/components/DateInput/index.tsx
@@ -211,7 +211,9 @@ export const DateInput = ({
                 disabled={disabled}
               />
               <StyledIconContainer>
-                {required && <Icon name="asterisk" color="danger" size={8} />}
+                {required ? (
+                  <Icon name="asterisk" color="danger" size={8} />
+                ) : null}
                 <StyledSeparator direction="vertical" />
                 <Icon
                   name="calendar-range"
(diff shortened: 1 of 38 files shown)


Lawndlwd (approved):

packages/ui/src/components/DateInput/index.tsx
```diff
@@ -212,19 +211,11 @@ export const DateInput = ({
...
               />
               <StyledIconContainer>
-                {required && (
-                  <Icon
-                    name="asterisk"
-                    color={theme.colors.danger.text}
-                    size={8}
-                  />
-                )}
+                {required && <Icon name="asterisk" color="danger" size={8} />}
```
QuiiBz, line 214:
```suggestion
                {required ? <Icon name="asterisk" color="danger" size={8} /> : null}
```

packages/ui/src/components/Icon/__stories__/Color.stories.tsx
```diff
@@ -3,11 +3,7 @@ import type { ComponentProps } from 'react'
 import { Icon } from '..'
 import { Stack } from '../../Stack'
 
-const colors: ComponentProps<typeof Icon>['color'][] = [
-  'primary',
-  'success',
-  'warning',
-]
+const colors = ['primary', 'success', 'warning'] as const
```
QuiiBz, line 6: Constants should be in uppercase:

```suggestion
const COLORS = ['primary', 'success', 'warning'] as const
```

QuiiBz (changes requested):

QuiiBz (approved):

PR #2281, opened 2023-02-28
fix(checkbox): set the right color

BABAK0T0:
## Type

- Bug

### Summarise concisely:

I fixed Checkbox component when is checked and disabled it means now is fulfilled and set with the right color depending on its state

### DIFF
--- a/packages/ui/src/components/Checkbox/index.tsx
+++ b/packages/ui/src/components/Checkbox/index.tsx
@@ -69,7 +69,10 @@ const CheckboxInput = styled('input', {
     cursor: not-allowed;
   }
 
-  &:checked + ${StyledIcon}, &[aria-checked='mixed'] + ${StyledIcon} {
+  &:checked:not(:disabled)
+    + ${StyledIcon},
+    &[aria-checked='mixed']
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Checkbox/index.tsx
```diff
@@ -69,7 +69,10 @@ const CheckboxInput = styled('input', {
     cursor: not-allowed;
   }
 
-  &:checked + ${StyledIcon}, &[aria-checked='mixed'] + ${StyledIcon} {
+  &:checked:not(:disabled)
+    + ${StyledIcon},
+    &[aria-checked='mixed']:not(:disabled)
+    + ${StyledIcon} {
```
matthprost, line 75: Maybe we can simplify it:

```suggestion
&:not(:disabled) {  
  &:checked + ${StyledIcon}, [aria-checked='mixed'] ${StyledIcon} {
    ...
  }
}
```

matthprost (changes requested):

matthprost (approved):

vincentaudebert (approved):

PR #2294, opened 2023-03-08
fix: menu and modal small issues

vincentaudebert:
## Type

- Enhancement

#### What is expected?

some improvements and small fixes on Menu and Modal

(Describe what you did)

1. Menu:

    align prop to be removed
    check arrow down color when placement top (seems primary while should be neutral).
    Also Menu.Item is missing in storybook api table
    Rename baseId en id 

2.
    Modal:
        rename Switch story to Toggle in example
        rename baseId en id et remove hardcoded default value, it should be optional parameter
        Toggle story snippet is not correct we don’t see Toggle inside of the example


packages/ui/src/components/Menu/index.tsx
```diff
@@ -173,9 +163,8 @@ const MenuList = styled.div<MenuListProps>`
 const FwdMenu = forwardRef(
   (
     {
-      align = { left: '50%', right: 'inherit' },
       ariaLabel = 'Menu',
-      baseId = 'menu',
+      id = 'menu',
```
matthprost, line 167: Default value should be removed as it's an idea we can have side effects that many modals has the same id

```suggestion
      id,
```

packages/ui/src/components/Modal/__stories__/Toggle.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+        />
+      )}
+      onBeforeClose={() => {
+        setToggled(false)
+      }}
+    >
+      <div>Content should be present in center of the modal</div>
+    </Modal>
+  )
+}
```
matthprost, line 29: Screenshot 2023-03-08 at 14 16 33

The code snippet of this story is not describing correctly what is happening, we missing the content of the disclosure. To display it in a better view you can remove the spread of props which will make storybook display the snippet exactly like we wrote it.

To do so you can:

```suggestion
export const ToggleStory: Story = () => {
  const [toggled, setToggled] = useState(false)

  return (
    <Modal
      disclosure={dialog => (
        <Toggle
          name="toggle"
          checked={toggled}
          onChange={() => {
            dialog?.toggle?.()
            setToggled(true)
          }}
        />
      )}
      onBeforeClose={() => {
        setToggled(false)
      }}
    >
      <div>Content should be present in center of the modal</div>
    </Modal>
  )
}
```

matthprost (changes requested):

packages/ui/src/components/Modal/index.tsx
```diff
@@ -270,7 +270,7 @@ export const Modal = memo(
     animated = false,
     animation = 'zoom',
     ariaLabel = 'modal',
-    baseId = 'modal',
+    id,
```
matthprost, line 273: You also forgot to remove `unstable_xx` from props, it might be linked to typing of the Modal 😄 

Screenshot 2023-03-08 at 14 19 02

matthprost (changes requested):

matthprost (approved):
Looks perfect to me 👌

QuiiBz (approved):

PR #2295, opened 2023-03-08
fix(Notice): allow complex children

chambo-e:


matthprost (approved):

ModuloM (approved):

ModuloM:
Maybe add an example in Storybook ?

PR #2296, opened 2023-03-08
fix: text to accept hmtlFor

matthprost:
## Type

- Enhancement

#### What is expected?

Added htmlFor property on Text as we are going to remove Label component to use Text instead.

```tsx
// Old usage
<Label htmlFor="myinputid">My Label</Label>

// New usage
<Text as="label" variant="bodyStrong" htmlFor="myinputid">My Label</Text>
```


vincentaudebert (approved):

QuiiBz (approved):
Should be a `feat` instead of `fix` I think?

PR #2300, opened 2023-03-15
feat(RowV2): new component

fabienhebert:
## Type

- Feature

### Summarise concisely:

A utility/display component to simplify our usage :
- We can replace Grid/Row/Col, by using multiples RowV2 (check story example)
- We can replace Stack with styled children, when each children should take a specific width

The component is based on `display:grid` and  templateColumn receive a `grid-template-column` value

### DIFF
--- a/packages/ui/src/components/RowV2/__stories__/AlignItems.stories.tsx
+++ b/packages/ui/src/components/RowV2/__stories__/AlignItems.stories.tsx
@@ -1,27 +1,33 @@
+import styled from '@emotion/styled'
 import type { Story } from '@storybook/react'
 import { RowV2 } from '..'
 import { Separator, Stack } from '../..'
 
+const DivWithBackground = styled.div`
+  background: ${({ theme }) => theme.colors.primary.background};
+  color: ${({ theme }) => theme.colors.primary.text};
+`
+
 export const AlignItems: Story = () => (
   <Stack gap={2}>
     <RowV2 templateColumns="repeat(3, 1fr)">
-      <div style={{ height: '100px', backgroundColor: '#eee' }}>
+      <DivWithBackground style={{ height: '100px' }}>
         100px height & default align
-      </div>
-      <div style={{ height: '50px', backgroundColor: '#eee' }}>
+      </DivWithBackground>
+      <DivWithBackground style={{ height: '50px' }}>
         50px height & default align
-      </div>
-      <div style={{ backgroundColor: '#eee' }}>auto height & default align</div>
+      </DivWithBackground>
+      <DivWithBackground>auto height & default align</DivWithBackground>
     </RowV2>
     <Separator />
     <RowV2 templateColumns="repeat(3, 1fr)" alignItems="center">
-      <div style={{ height: '100px', backgroundColor: '#eee' }}>
+      <DivWithBackground style={{ height: '100px' }}>
         100px height & align center
-      </div>
-      <div style={{ height: '50px', backgroundColor: '#eee' }}>
+      </DivWithBackground>
+      <DivWithBackground style={{ height: '50px' }}>
         50px height & align center
-      </div>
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/RowV2/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+type StyledRowProps = Pick<RowV2Props, 'gap' | 'templateColumns' | 'alignItems'>
+export const StyledRow = styled('div', {
+  shouldForwardProp: prop =>
+    !['templateColumns', 'gap', 'alignItems'].includes(prop),
+})<StyledRowProps>`
+  display: grid;
+  ${({ theme, gap = 2, alignItems = 'normal', templateColumns }) => `
+    grid-template-columns: ${templateColumns};
+    grid-gap: ${theme.space[gap]};
```
chambo-e, line 14:
```suggestion
    gap: ${theme.space[gap]};
```

packages/ui/src/components/RowV2/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  id?: string
+  className?: string
+  dataTestId?: string
+  children: ReactNode
+  templateColumns: string
+  gap?: keyof SCWUITheme['space']
+  alignItems?: CSSProperties['alignItems']
+}
+
+export const RowV2 = forwardRef(
```
chambo-e, line 29: Why the forwardRef ? 🤔

packages/ui/src/components/RowV2/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+    grid-template-columns: ${templateColumns};
+    grid-gap: ${theme.space[gap]};
+    align-items: ${alignItems};
+  `}
+`
+
+type RowV2Props = {
+  id?: string
+  className?: string
+  dataTestId?: string
```
chambo-e, line 22: Why not use the native name ?

packages/ui/src/components/RowV2/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { forwardRef } from 'react'
+import type { SCWUITheme } from '../../theme'
+
+type StyledRowProps = Pick<RowV2Props, 'gap' | 'templateColumns' | 'alignItems'>
+export const StyledRow = styled('div', {
+  shouldForwardProp: prop =>
+    !['templateColumns', 'gap', 'alignItems'].includes(prop),
+})<StyledRowProps>`
+  display: grid;
+  ${({ theme, gap = 2, alignItems = 'normal', templateColumns }) => `
```
chambo-e, line 12: Why the default gap here ?

chambo-e (changes requested):

packages/ui/src/components/RowV2/index.tsx
fabienhebert: I think it's better to have it on every component (we think about that). It's optional but let say you wanna scroll to this row for some reason you can it's convenient and there are multiples use case like that.

packages/ui/src/components/RowV2/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+`
+
+type RowV2Props = {
+  id?: string
+  className?: string
+  dataTestId?: string
+  children: ReactNode
+  templateColumns: string
+  gap?: keyof SCWUITheme['space']
+  alignItems?: CSSProperties['alignItems']
```
matthprost, line 26: no `justifyContent`?

matthprost (changes requested):

chambo-e (approved):

matthprost (changes requested):
Documentation is unreadable in dark theme

Screenshot 2023-03-17 at 15 35 19

matthprost (approved):

PR #2304, opened 2023-03-17
fix(verification_code): set the right color

BABAK0T0:
## Type

- Bug

### Summarise concisely:

Set the right color while hovering and focusing VerificationCode component

| Page |   Screenshot   |
| :--- | :--------: |
| VerificationCode  | Screenshot 2023-03-17 at 12 19 55 |

### DIFF
--- a/packages/ui/src/components/DateInput/__tests__/__snapshots__/index.test.tsx.snap
+++ b/packages/ui/src/components/DateInput/__tests__/__snapshots__/index.test.tsx.snap
@@ -4572,7 +4572,7 @@ exports[`DateInput renders correctly with date-fns locale ru 1`] = `
                   <p
                     class="e17s8tvr0 cache-1re7cvs-StyledText-StyledText e1tg3t120"
                   >
-                    январь 2021 г.
+                    январь 2021 г.
                   </p>
                 </div>
                 <span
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/DateInput/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -4572,7 +4572,7 @@ exports[`DateInput renders correctly with date-fns locale ru 1`] = `
                   <p
                     class="e17s8tvr0 cache-1re7cvs-StyledText-StyledText e1tg3t120"
                   >
-                    январь 2021 г.
+                    январь 2021 г.
```
chambo-e, line 4575: Bad node version ?

chambo-e (changes requested):

packages/ui/src/components/DateInput/__tests__/__snapshots__/index.test.tsx.snap
BABAK0T0: Yep, je suis en 18.14.2 mais normalement @matthprost a fait une MR pour fix ca 🤔

matthprost: @BABAK0T0 you should use 18.15 instead

matthprost (changes requested):
We need to handle error style too:
Screenshot 2023-03-17 at 14 23 39

It should same as for TextInput:
Screenshot 2023-03-17 at 14 23 49

chambo-e (approved):

PR #2336, opened 2023-03-21
fix(ListV2): simplify row selection disabled usage

fabienhebert:
## Type

- Enhancement

### Summarise concisely:

Regroup `isSelectDisabled` and `selectTooltip` in one prop `selectDisabled` which can be a boolean or a string (if string the  string will be displayed as a tooltip on the select checkbox)

### DIFF
--- a/packages/ui/src/components/ListV2/Row.tsx
+++ b/packages/ui/src/components/ListV2/Row.tsx
@@ -84,6 +84,9 @@ type RowProps = {
   children: ReactNode
   id: string
   expandable?: ReactNode
+  /**
+   * Row cannot be selected if this prop is provided. boolean true disabled selection, a string disable selection and a tooltip will be displayed on checkbox hover.
+   * */
   selectDisabled?: boolean | string
   isDisabled?: boolean
   sentiment?: (typeof SENTIMENTS)[number]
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/ListV2/Row.tsx
```diff
@@ -84,8 +84,7 @@ type RowProps = {
   children: ReactNode
   id: string
   expandable?: ReactNode
-  isSelectDisabled?: boolean
-  selectTooltip?: string
+  selectDisabled?: boolean | string
```
matthprost, line 87: Do you think we can some JSDoc here so it will be shown in storybook api table?

matthprost (changes requested):

packages/ui/src/components/ListV2/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -1456,6 +1456,7 @@ exports[`ListV2 Should render correctly with areRowSelectable 1`] = `
             aria-invalid="false"
             aria-label="select all"
             class="cache-1bap3ku-CheckboxInput eu28k4m2"
+            disabled=""
```
QuiiBz, line 1459: This prop shouldn't be rendered in the DOM, right?

QuiiBz (approved):

PR #2337, opened 2023-03-21
fix(NumberInput): enable controlled NumberInput

fabienhebert:
## Type

- Enhancement

### Summarise concisely:

Allow NumberInput to be controlled


packages/ui/src/components/NumberInput/index.tsx
```diff
@@ -175,24 +179,41 @@ export const NumberInput = ({
   size = 'large',
   step = 1,
   text,
+  defaultValue = 0,
```
matthprost, line 182: I'm a bit septic about this, what if we set minValue = 10 ? Will it increase defaultValue to the minimum ?

matthprost (changes requested):

packages/ui/src/components/NumberInput/index.tsx
fabienhebert: I agree, I didn't find the perfect combo for uncontrolled local state

Let say the defaultValue is `0` and minValue `10` should be use min as "real" defaultValue ?

The rules would be :
- defaultValue is 0 if not provided
- if a minValue is provided & defaultValue < minValue, use the minValue (it means minValue shouldn't have default value to 0)
- if a maxValue is provided & defaultValue > minValue, use the maxValue
- Otherwise you use the defaultValue provided or 0

vincentaudebert (approved):

fabienhebert:
> There is an issue, when I set a value I cannot change the value with the input

Wished controlled behavior

PR #2347, opened 2023-03-24
feat(SelectInput): add null to the possible option value type

ModuloM:
## Type

- Enhancement

#### What is expected?

Now `SlectOption` value attribut can be `null`


vincentaudebert:
why?

vincentaudebert:
Pipes are all down?

ModuloM:
> why?

Because to reset SelectOption to default when value is undefined, it doesn't work. So We have to be able to set SelectOption to null.

> Pipes are all down?

Yes, Thanks, I'll have to look why. I'll ping you when It's fixed

ModuloM:
> why?

Because to reset SelectOption to default when value is undefined, it doesn't work. So We have to be able to set SelectOption to null.

> Pipes are all down?

Yes, Thanks, I'll have to look why. I'll ping you when It's fixed

ModuloM:
> why?

Because to reset SelectOption to default when value is undefined, it doesn't work. So We have to be able to set SelectOption to null.

PR #2349, opened 2023-03-24
refactor(Expandable): smart expandable

matthprost:
## Type

- Refactor

### Summarise concisely:

The old component needed the prop `height` in order to make a smooth animation. In fact we need a height to set it to `max-height` to make an smooth animation.

Most of the time we do not set `height` as it is variable to the children size and we do not always know the children size. In most case the animation will look like this:

ezgif com-video-to-gif

---

#### Introducing: Smart Expandable Component ✨

Expandable became intelligent and measure the size of its children. No need to give a property `height` anymore, we can calculate the size of the children automatically and pass it directly to the `max-height` in CSS. This way the animation is way more smooth.

ezgif com-video-to-gif (1)

**But that's not all!** 

Expandable is also handling an Expandable within an Expandable. It seems nothing at first but I can assure you it's not that easy. When a parent Expandable has a `max-height` defined and its children dynamically increase you will need to "refresh" that `max-height` or find another way for the parent to expand as much as the children does.

Do not worry it is handled so you don't have to break your brain the same way as I did, here is the result:

ezgif com-video-to-gif (2)

### DIFF
--- a/packages/form/src/components/DateField/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/form/src/components/DateField/__tests__/__snapshots__/index.tsx.snap
@@ -180,12 +180,12 @@ exports[`DateField should render correctly 1`] = `
   opacity: 1;
 }
 
-.cache-s6tu1l {
+.cache-1lt3r2p {
   -webkit-transition: max-height 300ms ease-out,opacity 300ms ease-out;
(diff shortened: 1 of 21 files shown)


QuiiBz (approved):
Love it!

PR #2359, opened 2023-03-27
fix(tooltip): improve tooltip by hidding when generating positions

matthprost:
## Type

- Enhancement

#### What is expected?

We have an issue with the tooltip that sometimes it is shown for half a second on the top left of the screen. This might be due to the time to compute the positions of it and it is very different depending on each computer.

In order to avoid that problem I changed the initial position of the tooltip in a negative value, so it will be shown hidden on the screen and not more at the top left.


vincentaudebert:
test to regen?

QuiiBz (approved):

PR #2362, opened 2023-03-28
refactor(Card): new component

matthprost:
## Type

- Bug
- Feature
- Enhancement
- Documentation
- Migration
- Refactor

#### What is expected?

(Describe what you did)

1.

2.

3.


PetmePant (approved):

packages/ui/src/components/Card/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  &[data-isActive='true'] {
+    border: 1px solid ${({ theme }) => theme.colors.primary.border};
+  }
+
+  &[data-disabled='true'] {
+    border: 1px solid ${({ theme }) => theme.colors.neutral.borderDisabled};
+  }
+`
+
+export const Card = ({
```
chambo-e, line 47: How to replace Container with subtitle/rightTitle ? :)

packages/ui/src/components/Card/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+    {typeof header === 'string' ? (
+      <Text variant="heading" as="h2" disabled={disabled}>
+        {header}
+      </Text>
+    ) : (
+      header
+    )}
+    <BorderedBox data-isActive={isActive} data-disabled={disabled}>
+      {typeof children === 'function'
+        ? children({ disabled, isActive })
```
chambo-e, line 70: Do we really need this pattern ?

packages/ui/src/components/Card/index.tsx
matthprost: @chambo-e You could pass directly into `header` props:

```tsx
<Card header={<Stack gap={1}>
      <Stack gap={1} direction="row" alignItems="center">
        <Text variant="heading" as="h2">
          Advanced Header
        </Text>
        <Badge variant="success" size="small">
          New
        </Badge>
      </Stack>
      <Alert variant="info">This is an info alert in the header</Alert>
    </Stack>}
>
    This is the content of a Card
</Card>
```

packages/ui/src/components/Card/index.tsx
matthprost: @chambo-e It's not mandatory to have it but might be nice to get the state in the children to make the children follow the parent state. Why do you feel like it is wrong?

chambo-e: It's a pattern we are trying very hard to remove 😄 
Kinda uncomfortable having it available from the get go

Especially since these variables are not internal variables, they are props that will need to be passed from the parent, so the children should already be able to follow state ?

matthprost: @chambo-e I hesitated to do so, as you say we do not compute or do anything with these props, we only transmit them. So I think you're right we don't need to pass them this way, we can directly access them the same way we do for the parent.

I will push a fix to remove this behavior, thanks for feedback 😄

packages/ui/src/components/Card/__stories__/Disabled.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Stack } from '../../Stack'
+import { Text } from '../../Text'
+import { Card } from '../index'
+
+export const Disabled: Story = () => (
+  <Card header="Disabled Card" disabled>
+    <Stack gap={1}>
+      <Text as="p" variant="body" disabled>
+        We are using the disabled state returned from Card to disable its
+        children
```
fabienhebert, line 12: The text isn't good anymore

packages/ui/src/components/Card/__stories__/IsActive.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+      </Stack>
+    </Card>
+  )
+}
+
+IsActive.parameters = {
+  docs: {
+    storyDescription:
+      'You can highlight a Card by passing the `isActive` prop. It will highlight the Card and you can use the `isActive` state returned from Card to highlight its children.',
+  },
```
fabienhebert, line 51: same, update the description since Card does not accept the render props anymore

packages/ui/src/components/Card/__stories__/Disabled.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+      </Text>
+      <Button disabled>Button</Button>
+    </Stack>
+  </Card>
+)
+
+Disabled.parameters = {
+  docs: {
+    storyDescription:
+      'You can disable a Card by passing the `disabled` prop. It will disable the Card and you can use the `disabled` state returned from Card to disable its children.',
```
fabienhebert, line 22: same, update the description since Card does not accept the render props anymore

fabienhebert (approved):

PR #2363, opened 2023-03-28
feat(components): add className and data-testid - Part 1

BABAK0T0:
## Type

- Enhancement

### Summarise concisely:

Add className and data-testid on every components

### DIFF
--- a/packages/ui/src/components/Bullet/index.tsx
+++ b/packages/ui/src/components/Bullet/index.tsx
@@ -91,8 +91,13 @@ export const Bullet = ({
   tooltipBaseId,
   'data-testid': dataTestId,
 }: BulletProps) => (
-  <Tooltip id={tooltipBaseId} text={tooltip} data-testid={dataTestId}>
-    <StyledContainer variant={variant} size={size} className={className}>
+  <Tooltip id={tooltipBaseId} text={tooltip}>
+    <StyledContainer
+      variant={variant}
+      size={size}
+      className={className}
+      data-testid={dataTestId}
+    >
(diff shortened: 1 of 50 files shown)


packages/ui/src/components/Bullet/index.tsx
```diff
@@ -88,8 +89,9 @@ export const Bullet = ({
   text,
   tooltip,
   tooltipBaseId,
+  'data-testid': dataTestId,
 }: BulletProps) => (
-  <Tooltip id={tooltipBaseId} text={tooltip}>
+  <Tooltip id={tooltipBaseId} text={tooltip} data-testid={dataTestId}>
     <StyledContainer variant={variant} size={size} className={className}>
```
matthprost, line 95: It should not be applied on tooltip but on the container instead:

```suggestion
  <Tooltip id={tooltipBaseId} text={tooltip}>
    <StyledContainer variant={variant} size={size} className={className} data-testid={dataTestId}>
```

packages/ui/src/components/Carousel/index.tsx
```diff
@@ -156,6 +156,8 @@ export const Carousel = ({
           e.preventDefault()
           e.stopPropagation()
         }}
+        className={className}
+        data-testid="scrollbar-wrapper"
```
matthprost, line 160: Why hardcoded value? Can't we add data-testid as prop and set it here?

matthprost (changes requested):

matthprost (approved):

QuiiBz (approved):

PR #2370, opened 2023-03-29
feat(components): add className and data-testid - Part 2

BABAK0T0:
## Type

- Enhancement

### Summarise concisely:

Add className and data-testid on every components


matthprost (approved):

QuiiBz (approved):

matthprost:
As you added new properties you may need to update unit test too to avoid decreasing coverage

PR #2377, opened 2023-03-30
fix(tooltip): flickering effect due to animation going backwards

matthprost:
## Type

- Bug

#### Flickering Bug on Tooltip when un-hovering 🐞

This is an actual bug we have on Tooltip causing a flicker effect. This is because our animation is going back to its initial state once completed. In some cases this can cause the flickering.

To fix it I simply added `forwards` to the animation in CSS, forcing the animation to keep its last position at the end of the animation solving the issue.

| Before  | After |
| ------------- | ------------- |
|  Screen Recording 2023-03-30 at 17 57 44 | ezgif-4-96aeef49aa |

### DIFF
--- a/.changeset/rare-hotels-try.md
+++ b/.changeset/rare-hotels-try.md
@@ -1,0 +1,6 @@
+---
+'@scaleway/ui': patch
+'@scaleway/form': patch
+---
+
+Fix flickering effect on tooltip when un-hovering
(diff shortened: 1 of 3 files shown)


chambo-e (approved):

QuiiBz (changes requested):
Missing changeset

QuiiBz (approved):

PR #2378, opened 2023-03-30
feat(Popover): new component

matthprost:
## Type

- Feature

#### New Component Popover 🔬

We created a new component named Popover that is an extension of the Tooltip allowing to show more information and with different styles.

Popover can be found in two variants `default` and `primary`

Screen Recording 2023-03-31 at 14 24 20

This incredible component is also available in different sizes: `small`, `medium` and `large` which only defines the size of the width (and not the height).

Screen Recording 2023-03-31 at 14 24 42

#### How to control it?

1. **Uncontrolled** - You can simply pass `visible` to the component and it will handle closing when clicking on close icon button. This only works if you need to Popover to only be displayed on first render if you set `visible={true}`. Example:

```tsx
// This usecase is only when we want to display popover opened on first render
export const MyUncontrolledPopover = () => (
    <Popover
      visible
      title="Popover Title"
      content="Content"
    >
      <Icon name="help-circle-outline" size={24} />
    </Popover>
)
```

If you need the Popover to open and close on click on an external button, you will need to make handle the state by your own, check the seconde possibility. ⬇️ 

2. **Controlled** - It's quite simple you need a state that you pass through `visible={myState}` and you should also handle the on close icon button through `onClose` prop. For example:

```tsx
export const MyControlledPopover = () => {
  const [opened, setOpened] = useState(false)

  return (
    <Popover
      visible={opened}
      title="Popover Title"
      content="Content"
      onClose={() => setOpened(false)}
    >
      <Button onClick={() => setOpened(true)} variant="secondary">
        Open Popover
      </Button>
    </Popover>
  )
}
```


packages/ui/src/components/Tooltip/__stories__/index.stories.tsx
```diff
@@ -15,5 +15,5 @@ export default {
 } as Meta
 
 export { Playground } from './Playground.stories'
-export { Placement } from './Placement.stories'
-export { MaxWidth } from './MaxWidth.stories'
+/* export { Placement } from './Placement.stories'
```
philibea, line 18: ??

packages/ui/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+    setOpened(visible)
+  }, [visible])
+
+  const onCloseCallBack = useCallback(() => {
+    setOpened(false)
+    onClose?.()
+  }, [onClose])
+
+  return (
+    <StyledTooltip
```
philibea, line 138: it's not really a tooltip anymore so the role need to be dialog and not tooltip ?

philibea: https://github.com/scaleway/scaleway-ui/pull/2378/files#diff-4b090bb740787ad99908e13624821cbfbd710c1700a550b5a00603a6e05057c3L297

philibea: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role

chambo-e (approved):

fabienhebert (approved):

matthprost:
Awaiting https://github.com/scaleway/scaleway-ui/pull/2371 to be merge so we can use it for close button.

PetmePant:
Everything seems fine the only thing that I see is: 

- When you trigger popover and you click on another location the popover should close

packages/ui/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  useEffect(
+    () => {
+      document.addEventListener('click', handleClickOutside, true)
+
+      return () => {
+        document.removeEventListener('click', handleClickOutside, true)
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
fabienhebert, line 132: Why do we need that ? can't we useCallback the handleClickOutside ?

packages/ui/src/components/Tooltip/index.tsx
```diff
@@ -125,187 +126,201 @@ type TooltipProps = {
...
+      maxWidth = 232,
+      visible,
+      innerRef,
+      role = 'tooltip',
+      'data-testid': dataTestId,
+    }: TooltipProps,
+    tooltipRef: Ref<HTMLDivElement>,
+  ) => {
+    const childrenRef = useRef<HTMLDivElement>(null)
+    useImperativeHandle(innerRef, () => childrenRef.current)
```
fabienhebert, line 150: innerRef + localRef ?

PR #2402, opened 2023-04-05
feat(ButtonV2): support href

fabienhebert:
## Type

- Feature
- Enhancement

### Summarise concisely:

Make ButtonV2 supports href (and download/target)


packages/ui/src/components/ButtonV2/index.tsx
```diff
@@ -275,7 +297,54 @@ export const ButtonV2 = forwardRef(
...
+        <Component
+          className={className}
+          data-testid={dataTestId}
+          disabled={computeIsDisabled}
+          fullWidth={fullWidth}
+          iconPosition={iconPosition}
+          sentiment={sentiment}
+          size={size}
+          type={type}
+          href={href}
```
matthprost, line 317: Should we support `target` also?

matthprost (changes requested):

matthprost (approved):

PR #2404, opened 2023-04-06
refactor(Alert): new design and properties

matthprost:
## Type

- Enhancement

### New Alert Design 🚨

We created a new Design for Alert to make it more clean and readable. Previous Alert was quite hard to read even more for those who have visual handicap.

**Old Alert visual**
Screenshot 2023-04-06 at 14 43 47

New Alert visual
Screenshot 2023-04-12 at 16 10 52

This new Alert also implement new properties for more possibilities. 

Among them we have:
- `buttonText` and `onButtonClick` which will add a button at the end of the Alert.
- `isClosable` and `onClose` which make the Alert closable without having to handle a state and you can even trigger some actions once close button is clicked with `onClose` property.
Screenshot 2023-04-06 at 14 47 28

The new Alert is also capable of changing the direction of its content depending on the size of its text. This way, when a text is too long to be displayed in a row, we automatically display the content as column.

Screenshot 2023-04-06 at 14 55 37


ModuloM (approved):

packages/ui/src/components/Alert/index.tsx
```diff
@@ -15,11 +17,12 @@ const alertStyles = ({
   theme: Theme
   variant: AlertType
 }): SerializedStyles => {
-  const sentiment = theme.colors[variant] || theme.colors.danger
+  const sentiment = theme.colors[variant]
 
   return css`
     background-color: ${sentiment.background};
     color: ${sentiment.text};
+    border-left: 4px solid ${sentiment.borderWeak};
```
QuiiBz, line 25: Should we use the value from the theme instead of hardcoding `4px`?

QuiiBz (approved):

packages/ui/src/components/Alert/index.tsx
matthprost: @QuiiBz it's a border not a space, so we can put a hardcoded value. Maybe in the futur we will have border thickness in the theme, until then we do hardcoded 😄

PetmePant:
Alert:

- The name need to change to inline alert (will help to differentiate from other type of alerts (global alert etc that will come) 

- Close icon button needs to be from the neutral palette (mixing with semantics from alert is problematic for users, (close icon should not contain any color that can be confused with extra meaning) 

- We need to update the icon of danger/warning with the circle option. ( do I need to make a specific ticket for that? it seems that we missed it from the icon library)

PetmePant (approved):

PR #2411, opened 2023-04-10
fix(form): meta type error function

philibea:
## Type

- Refactor

#### What is expected?

use the right type for meta FieldMetaState instead of FieldState.

This will avoid cast error like this: 

``` javascript
 const { input, meta } = useField(`${name}-buffer`, {
      type,
      value: '',
      validate: validateFn,
    })

    const error = getError({
      meta: meta as FieldState<unknown>,

```

=> 
``` javascript
 const { input, meta } = useField(`${name}-buffer`, {
      type,
      value: '',
      validate: validateFn,
    })

    const error = getError({
      meta,

```

### DIFF
--- a/.changeset/quick-dodos-tan.md
+++ b/.changeset/quick-dodos-tan.md
@@ -1,0 +1,5 @@
+---
+'@scaleway/form': minor
+---
+
+fix type issue with meta field from getErrors function
--- a/.changeset/quick-dodos-tan.md
+++ b/.changeset/quick-dodos-tan.md
@@ -2,4 +2,4 @@
 '@scaleway/form': minor
 ---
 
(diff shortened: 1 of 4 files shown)


matthprost (approved):

ModuloM (approved):

.changeset/quick-dodos-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+'@scaleway/form': minor
+---
+
+fix type issue with meta field from getErrors function
```
QuiiBz, line 5: Let's make the changeset prettier:
```suggestion
Fix type issue with meta field from `getErrors` function
```

PR #2420, opened 2023-04-12
docs(SelectInput): add group story

fabienhebert:
## Type

- Documentation

### Summarise concisely:

Add a story to explain how group options when using SelectInput


.changeset/early-falcons-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@scaleway/ui': patch
```
matthprost, line 2: It's just a documentation change so it can be a minor

matthprost (changes requested):

.changeset/early-falcons-return.md
ModuloM: Not needed as we discuss this morning ;)

ModuloM (approved):

.changeset/early-falcons-return.md
matthprost: @ModuloM to be removed :D

matthprost (changes requested):

johnrazeur (approved):

vincentaudebert (approved):

packages/ui/src/components/SelectInput/__stories__/Group.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <div style={{ height: '120px' }}>
+      <StoryComponent />
+    </div>
+  ),
+]
+
+Group.parameters = {
+  docs: {
+    storyDescription:
+      'By using `options` prop you can regroup options by category/group (no jsx way to do that).',
```
QuiiBz, line 57:
```suggestion
      'By using the `options` prop you can regroup options by category/group (not possible when using JSX).',
```

QuiiBz (approved):

PR #2421, opened 2023-04-12
fix(Card): add forward ref

matthprost:
## Type

- Bug

#### What is expected?

ForwardRef is missing in Card component.


fabienhebert (approved):

philibea:
@matthprost  we don't have a CardField ?

PR #2424, opened 2023-04-13
refactor(VerificationCode): change state management

matthprost:
## Type

- Refactor

#### What is expected?

I remove `data-id` in `VerificationCode` as it was confusing and not type safe. This prop was used to manage state of the input, instead I'm passing `index` directly in the functions who needs it.

### DIFF
--- a/.changeset/calm-moons-judge.md
+++ b/.changeset/calm-moons-judge.md
@@ -1,0 +1,5 @@
+---
+'@scaleway/ui': minor
+---
+
+change state handle of `VerificationCode`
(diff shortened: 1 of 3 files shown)


ModuloM (approved):

vincentaudebert (approved):
Looks good to me, just wondering when we paste, should it focus on the last char pasted or maybe the next one as when we type something?

At the moment it stays focused on the char where we paste

matthprost:
> Looks good to me, just wondering when we paste, should it focus on the last char pasted or maybe the next one as when we type something?
> 
> At the moment it stays focused on the char where we paste

Good point we need to focus the last element, I will check to improve this component behavior thanks for feedback 😄

PR #2464, opened 2023-04-24
feat(PieChart): using new data color

matthprost:
## Type

- Feature

#### What is expected?

PieChart integrate new chart colors we won't need to set them in data anymore they will be get from theme automatically making theme dark theme compatible.

Screenshot 2023-04-24 at 11 46 39

Screenshot 2023-04-24 at 11 51 33

### DIFF
--- a/packages/ui/src/components/PieChart/index.tsx
+++ b/packages/ui/src/components/PieChart/index.tsx
@@ -87,11 +87,8 @@ export const PieChart = ({
       if (Number(a.replace('data', '')) < Number(b.replace('data', ''))) {
         return -1
       }
-      if (Number(a.replace('data', '')) > Number(b.replace('data', ''))) {
-        return 1
-      }
 
-      return 0
+      return 1
     })
     .map(
       key =>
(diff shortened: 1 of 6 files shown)


chambo-e (commented):
Does this mean we loose consistent products Colors ?

matthprost:
> 

@chambo-e yes we loose one color = one product, this is done on purpose :)

packages/ui/src/components/PieChart/index.tsx
```diff
@@ -81,6 +81,23 @@ export const PieChart = ({
     [emptyLegend],
   )
 
+  const localColors = Object.keys(colors.other.data.charts)
+    .filter(key => !['success', 'danger'].includes(key))
+    .sort((a, b) => {
+      if (Number(a.replace('data', '')) < Number(b.replace('data', ''))) {
+        return -1
+      }
+      if (Number(a.replace('data', '')) > Number(b.replace('data', ''))) {
```
QuiiBz, line 90: To avoid duplicating:
```suggestion
      const numberA = Number(a.replace('data', ''))
      const numberB = Number(b.replace('data', ''))
      if (numberA < numberB) {
        return -1
      }
      if (numberA > numberB) {
```

matthprost: @QuiiBz thanks for feedback, I simplified it a bit more so no need for variable now 😛

QuiiBz (approved):

PR #2468, opened 2023-04-24
feat(TableV2): new component

fabienhebert:
## Type

- Feature

### Summarise concisely:

Add a new component TableV2. The endgoal is to merge `Table (v1)`, `List (v1) variant table` & `List (v1) variant explorer` in this new component.

Table (V1) didn't supported ordering & selectable rows


packages/ui/src/components/ListV2/HeaderCell.tsx
```diff
@@ -72,16 +72,27 @@ export const HeaderCell = ({
...
-        onOrder
-          ? () => onOrder(order === 'ascending' ? 'desc' : 'asc')
+      className={className}
+      onClick={handleOrder}
+      onKeyDown={
+        handleOrder
+          ? event => {
+              if (event.key === ' ') {
+                handleOrder()
+                event.preventDefault()
```
matthprost, line 89: I don't think we need the preventDefault 🤔

packages/ui/src/components/TableV2/HeaderCell.tsx
```diff
@@ -0,0 +1,134 @@
...
+  ${({ width, maxWidth, minWidth }) => `
+    ${width ? `width: ${width};` : ''}
+    ${maxWidth ? `max-width: ${maxWidth};` : ''}
+    ${minWidth ? `min-width: ${minWidth};` : ''}
+  `}
+
+  padding: ${({ theme }) => theme.space['1']};
+`
+
+const HeaderCellContent = styled('div')`
```
matthprost, line 57: Why not using a `Text` with prop `as="div"` so you don't need to add CSS related to text style

packages/ui/src/components/TableV2/HeaderCell.tsx
```diff
@@ -0,0 +1,134 @@
...
+    <StyledHeaderCell
+      aria-sort={order}
+      className={className}
+      onClick={handleOrder}
+      onKeyDown={
+        handleOrder
+          ? event => {
+              if (event.key === ' ') {
+                handleOrder()
+                event.preventDefault()
```
matthprost, line 112: Same here I don't think we need preventDefault it should work without

matthprost (changes requested):

packages/ui/src/components/ListV2/HeaderCell.tsx
```diff
@@ -72,16 +72,27 @@ export const HeaderCell = ({
...
       aria-sort={order}
-      onClick={
-        onOrder
-          ? () => onOrder(order === 'ascending' ? 'desc' : 'asc')
+      className={className}
+      onClick={handleOrder}
+      onKeyDown={
+        handleOrder
+          ? event => {
+              if (event.key === ' ') {
```
ModuloM, line 87: Maybe we should also map `enter` key?
As we use this as a button (https://www.w3.org/WAI/ARIA/apg/patterns/button/)

packages/ui/src/components/TableV2/HeaderCell.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  return (
+    <StyledHeaderCell
+      aria-sort={order}
+      className={className}
+      onClick={handleOrder}
+      onKeyDown={
+        handleOrder
+          ? event => {
+              if (event.key === ' ') {
```
ModuloM, line 110: Maybe we should also map enter key?

ModuloM:
Juste quick note on checkbox component, it doesn't listen to keydown `Enter` to change it's state (but not related to TableV2) 🚀

matthprost:
> Juste quick note on checkbox component, it doesn't listen to keydown `Enter` to change it's state (but not related to TableV2) 🚀

@ModuloM This might seems strange but for changing state with accessibility you should use space bar instead of enter. If you check on native checkbox that's how you change it and enter doesn't work 😛

packages/ui/src/components/ListV2/HeaderCell.tsx
```diff
@@ -72,20 +72,34 @@ export const HeaderCell = ({
...
       aria-sort={order}
-      onClick={
-        onOrder
-          ? () => onOrder(order === 'ascending' ? 'desc' : 'asc')
+      className={className}
+      onClick={handleOrder}
+      onKeyDown={
+        handleOrder
+          ? event => {
+              if (event.key === ' ' || event.key === 'Enter') {
```
matthprost, line 87:
```suggestion
              if (event.key === ' ') {
```

packages/ui/src/components/TableV2/HeaderCell.tsx
```diff
@@ -0,0 +1,132 @@
...
+    : undefined
+
+  return (
+    <StyledHeaderCell
+      className={className}
+      onClick={handleOrder}
+      onKeyDown={
+        handleOrder
+          ? event => {
+              if (event.key === ' ' || event.key === 'Enter') {
```
matthprost, line 103:
```suggestion
              if (event.key === ' ') {
```
Screenshot 2023-05-03 at 10 32 43

Here is a nice tool to get the value of each keys: https://www.toptal.com/developers/keycode

packages/ui/src/components/TableV2/__stories__/Style.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+  { stripped: false, bordered: false },
+  { stripped: true, bordered: false },
+  { stripped: false, bordered: true },
+  { stripped: true, bordered: true },
+]
+
+export const Style: Story = () => (
+  <Stack gap={2}>
+    {combos.map(combo => (
+      <Stack gap={1}>
```
matthprost, line 16: A key is missing

matthprost (approved):

PR #2469, opened 2023-04-25
fix(Tooltip): improve accessibility the wrapper can now be focused for voiceover

matthprost:
## Type

- Enhancement

#### What is expected?

Tooltip wrapper can now be focused using the keyboard, once the tooltip wrapper is focused the tooltip itself is shown. It can then be removed  with `Escape` key.

Here are the specifications:
> A tooltip typically becomes visible, after a short delay of generally one to five seconds, in response to a mouse hover, or after the owning element receives keyboard focus. Just as it is opened automatically, without user request, it is closed automatically when the focus is lost or on mouse out. It should also close when the user presses the Escape key.
[https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role)

We have to keep in mind an important statement:
> Because the tooltip itself never receives focus and is not in the tabbing order, a tooltip can not contain interactive elements like links, inputs, or buttons.
[https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role)

https://user-images.githubusercontent.com/15812968/234213731-3358f08f-d69c-42db-962c-0da912a60c99.mov

### DIFF
--- a/.changeset/four-queens-buy.md
+++ b/.changeset/four-queens-buy.md
@@ -1,0 +1,5 @@
+---
+'@scaleway/ui': minor
+---
+
+Tooltip is now focusable for the voice over to be able to read it
--- a/packages/ui/src/components/Button/__tests__/__snapshots__/index.tsx.snap
+++ b/packages/ui/src/components/Button/__tests__/__snapshots__/index.tsx.snap
@@ -1641,6 +1641,7 @@ exports[`Button should render correctly with a tooltip 1`] = `
 <div
     aria-describedby="test"
     class="cache-yb56yg-StyledChildrenContainer ep3agpf0"
+    tabindex="0"
   >
     <button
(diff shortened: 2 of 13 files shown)


vincentaudebert:
Can we reach 100% code coverage btw ? or too hard?

vincentaudebert (approved):

QuiiBz (approved):

matthprost:
> Can we reach 100% code coverage btw ? or too hard?

Quite hard to reach, the last use case are difficult to test

PR #2532, opened 2023-05-18
fix(buttonV2): add white-space: nowwrap

philibea:
## Type

- Bug

#### What is expected?

i don't think we should allow white-space inside a Button by design this is what we have in Button ( V1 )

Before: 
Capture d’écran 2023-05-18 à 00 19 45

After: 
Capture d’écran 2023-05-18 à 00 19 50


QuiiBz (approved):

matthprost (changes requested):
Changeset is missing

matthprost (approved):

PR #2536, opened 2023-05-19
refactor(NumberInput): new parameters and behavior

matthprost:
## Type

- Feature

### NumberInput New Features 🚀

NumberInput needed some updates to accept more properties. First of all NumberInput needed to be empty and show a placeholder. Before this update NumberInput should always have a value and couldn't be empty.

| Before | After |
| :------ | ----: |
| Screenshot 2023-05-22 at 10 38 27 | Screenshot 2023-05-22 at 10 32 17 |

You can also add a placeholder to detail the content that is needed:
Screenshot 2023-05-22 at 10 32 04

Here is what happen when you start increasing the value:
Screenshot 2023-05-22 at 10 39 20

Another missing feature is to be able to show that the content is in error. Now that the input can be empty, if it becomes a required input we need to show an error in case it is not filled.

Screenshot 2023-05-22 at 10 32 23


packages/ui/src/components/NumberInput/index.tsx
```diff
@@ -216,43 +207,72 @@ export const NumberInput = ({
...
-    if (maxValue && currentValue > maxValue) onMaxCrossed?.()
-    if (currentValue < minValue) onMinCrossed?.()
+      if (maxValue && currentValue > maxValue) onMaxCrossed?.()
+      if (minValue && currentValue < minValue) onMinCrossed?.()
 
-    setValue(boundedValue)
+      setValue(boundedValue)
 
-    if (onBlur) onBlur(event)
+      if (onBlur) onBlur(event)
```
QuiiBz, line 274:
```suggestion
      onBlur?.(event)
```

QuiiBz (changes requested):
We can increment/decrement once when the component is disabled:

https://github.com/scaleway/scaleway-ui/assets/43268759/70d6344f-59a9-4556-bb15-04e4cb8aa332

QuiiBz (approved):

PR #2547, opened 2023-05-22
fix: use ButtonV2

fabienhebert:
## Type

- Enhancement
- Refactor

### Summarise concisely:

use ButtonV2 in scw-ui instead of Button (v1)
Remove Button(V1)
Rename ButtonV2 to Button

### Breaking changes:

- Button is not the same component anymore
- scw-form Submit component has not the same props


packages/ui/src/__stories__/ButtonMigration/Variants.tsx
```diff
@@ -1,91 +1,9 @@
...
-  success: { variant: 'filled', sentiment: 'success' },
-  'success-bordered': { variant: 'outlined', sentiment: 'success' },
-  'success-soft-bordered': { variant: 'outlined', sentiment: 'neutral' },
-  transparent: { variant: 'ghost', sentiment: 'neutral' },
-  warning: { variant: 'filled', sentiment: 'danger' },
-  'warning-bordered': { variant: 'outlined', sentiment: 'danger' },
-  'warning-soft-bordered': { variant: 'outlined', sentiment: 'neutral' },
-}
+import ButtonActionMigrationImg from '../assets/button-action-migration.png'
+import ButtonMigrationImg from '../assets/button-migration.png'
```
matthprost, line 2: The migration page on storybook doesn't seems to work anymore, I have a white page

packages/ui/src/components/Menu/__stories__/Template.stories.tsx
```diff
@@ -1,9 +1,9 @@
 import type { ComponentStory } from '@storybook/react'
 import { Menu } from '..'
-import { Button } from '../../Button'
+import { Button } from '../../index'
 
 export const DefaultDisclosure = (
-  <Button action icon="dots-horizontal" iconSize={24} variant="transparent" />
+  <Button icon="dots-horizontal" variant="ghost" size="medium" />
```
matthprost, line 6:
```suggestion
  <Button icon="dots-horizontal" variant="ghost" size="small" />
```

Old disclosure button was 32x32, so it should be small size

packages/ui/src/components/Modal/index.tsx
```diff
@@ -343,12 +343,16 @@ export const Modal = memo(
...
-                    title="close"
-                    variant="transparent"
+                    variant="ghost"
+                    size="small"
                     icon="close"
-                    iconSize={20}
-                    action
+                    sentiment="neutral"
+                    aria-label="close"
+                    tooltip="close"
```
matthprost, line 355: As we have `aria-label` we do not need tooltip here. Previous title was used for a11y, screen reader will be capable of reading the `aria-label`

matthprost (changes requested):

matthprost (approved):

BABAK0T0 (approved):

PR #2549, opened 2023-05-23
feat(Row): add justify-content property

johnrazeur:
## Type

- Feature

Add `justify-content` property to Row.


fabienhebert:
What is the end goal of this addition ? I hesitate adding it when the component has been created but I didn't the a real advantage for 99% of case (otherwhile a styled(Row) could work) ?

matthprost (approved):

fabienhebert (approved):

PR #2551, opened 2023-05-24
refactor(Radio): change label

johnrazeur:
## Type

- Refactor

The radio component can be use without a displayed label, the aria-label prop have been added to ensure accessibility.

### DIFF
--- a/packages/ui/src/components/Radio/index.tsx
+++ b/packages/ui/src/components/Radio/index.tsx
@@ -83,7 +83,8 @@ const RadioContainer = styled.div`
   align-items: flex-start;
   gap: ${({ theme }) => theme.space['1']};
 
-  &[aria-disabled='false'] {
+  &[aria-disabled='false'],
+  &[aria-disabled='false'] > label {
     cursor: pointer;
   }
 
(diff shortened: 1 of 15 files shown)


BABAK0T0 (changes requested):
Better to removes the root label when there is no label and puts the aria-label on the input

packages/ui/src/components/Radio/index.tsx
```diff
@@ -138,7 +140,17 @@ type RadioProps = {
     | 'id'
     | 'name'
     | 'required'
-  >
+  > &
+  (
+    | {
+        'aria-label': string
```
philibea, line 146: This is not typed when we do that right ?

philibea: should we use ariaLabel as props when we control usage with children ?

johnrazeur: It's typed, it's an XOR (the XOR custom type break eslint idk why). The aria-label prop can't be used when `children` is provided.

matthprost (approved):

BABAK0T0 (approved):

PR #2562, opened 2023-05-25
refactor: remove website

matthprost:
## Type

- Refactor

#### What is expected?

Website can be removed as we now have Zero Height documentation that will be the entry point of the design system.


philibea:
can we keep it as example ?

fabienhebert (approved):

QuiiBz (approved):

PR #2566, opened 2023-05-26
feat(Tag,TagList): add copiable props

QuiiBz:
## Type

- Feature
#### What is expected?

Add copiable props to `Tag` & `TagList`, similar to `CopyButton`:
- `copiable` is a boolean defining if the tag is copiable or not
- `copyText` is a string to set the tooltip's copy text
- `copiedText` is a string to set the tooltip's copied text

https://github.com/scaleway/scaleway-ui/assets/43268759/c1c7e72c-e3df-49d0-8021-407be0c30674


packages/ui/src/components/Tag/index.tsx
```diff
@@ -106,5 +126,68 @@ export const Tag = ({
...
+  </>
 )
+
+export const Tag = ({
+  children,
+  isLoading,
+  onClose,
+  icon,
+  copiable = false,
+  copyText = 'Copy',
```
philibea, line 138: i don't think we should put default props for translation no ?

philibea: we should make the props required if copiable is true

QuiiBz: We already hardcode these in `CopyButton`

DorianMaliszewski (approved):

packages/ui/src/components/TagList/index.tsx
```diff
@@ -50,6 +50,9 @@ type TagListProps = {
    * This property define maximum width of each tag. This doesn't apply for tags in tooltip.
    */
   multiline?: boolean
+  copiable?: boolean
+  copyText?: string
+  copiedText?: string
```
matthprost, line 55: Better to take those type from Tag directly 

```ts
& Pick<ComponentProps<typeof Tag>, 'copiable' | 'copyText' | 'copiedText'>
```

matthprost (changes requested):

packages/ui/src/components/Tag/index.tsx
```diff
@@ -106,5 +125,68 @@ export const Tag = ({
...
+  copiedText = 'Copied!',
+  disabled,
+  variant = 'neutral',
+  className,
+  'data-testid': dataTestId,
+}: TagProps) => {
+  const [isCopied, setCopied] = useClipboard(
+    typeof children === 'string' ? children : '',
```
vincentaudebert, line 145: Should disable copy if it's not string then?

QuiiBz: It's already done line 151

matthprost (approved):
Nice work 🚀

PetmePant (approved):

PR #2593, opened 2023-06-05
fix(textinput): style issue on padding

philibea:
## Type

- Bug

#### What is expected?

wrong css style

### DIFF
--- a/packages/ui/src/components/TextInput/index.tsx
+++ b/packages/ui/src/components/TextInput/index.tsx
@@ -289,14 +289,14 @@ const StyledInput = styled('input', {
     multiline &&
     !hasLabel &&
     `
-    padding-top: ${theme.space[1]};
+    padding-top: ${theme.space['1']};
   `}
 
   ${({ hasRightElement, rightElementPadding, theme }) =>
     hasRightElement &&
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/TextInput/index.tsx
```diff
@@ -292,10 +292,10 @@ const StyledInput = styled('input', {
     padding-top: 8px;
   `}
 
-  ${({ hasRightElement, rightElementPadding, theme }) =>
+  ${({ hasRightElement, rightElementPadding }) =>
     hasRightElement &&
     `
-    padding-right: ${rightElementPadding || theme.space['6']}px;
+    padding-right: ${rightElementPadding || 32}px;
```
QuiiBz, line 298: Why we don't use `theme['4']`?

QuiiBz (changes requested):

fabienhebert:
Still have issue when multiple right elements but we can do it in another MR too

Capture d’écran 2023-06-05 à 12 25 33

fabienhebert (approved):

QuiiBz (approved):

PR #2596, opened 2023-06-06
fix(TextInput): padding with multiple right components

QuiiBz:
## Type

- Bug

#### What is expected?

Fix right padding when we have multiple right components (unit, valid, required, ...)

| url  | Screenshot 2023-06-06 at 11 08 33 | Screenshot 2023-06-06 at 11 08 52 |

### DIFF
--- a/.changeset/yellow-bags-doubt.md
+++ b/.changeset/yellow-bags-doubt.md
@@ -1,0 +1,5 @@
+---
+'@scaleway/ui': patch
+---
+
+Fix right padding on TextInput when having multiple right components (unit, valid, required...)


packages/ui/src/components/TextInput/index.tsx
```diff
@@ -551,6 +544,11 @@ export const TextInput = forwardRef<
       dataTestId,
     ])
 
+    const rightElementPadding =
+      rightComponentsArray.length * 32 +
+      (required ? 32 : 0) +
+      (unit || hasRightElement ? 8 : 0)
```
fabienhebert, line 550: I struggle to understand why 8 here mb we can explain the calculation somewhere
+ since the spacing is now in the theme, mb we should just calculate space unit here, then multiply by theme.space['1'] using a calc-width ?

It works for everycase, but the last one isn't perfect, the padding is higher than the other ones

Capture d’écran 2023-06-06 à 12 35 49

QuiiBz: As discussed, I've refactored the way it works by using a factor and multiplying it by a theme value.

philibea (approved):

PR #2608, opened 2023-06-08
refactor(TagList): use Popover component instead of Tooltip

QuiiBz:
## Type

- Refactor

- Use a `Popover` instead of a `Tooltip` for the threshold
- Add a new required `popoverTitle` prop
- Set `Popover` title to `bodyStrong` instead of `headerSmall`

| url  | Screenshot 2023-06-08 at 11 29 47 | Screenshot 2023-06-08 at 11 30 11 |
| url  | Screenshot 2023-06-08 at 11 30 32 | Screenshot 2023-06-08 at 11 30 52 |

Custom Popover content:

Screenshot 2023-06-08 at 11 31 29


ModuloM (approved):

packages/ui/src/components/TagList/index.tsx
```diff
@@ -101,27 +104,35 @@ export const TagList = ({
...
-                >
-                  {tag}
-                  {index < tags.slice(visibleTagsCount).length - 1 ? ',' : null}
-                </span>
-              ))}
-            </StyledManyTagsContainer>
+        <Popover
+          title={popoverTitle}
+          visible={isVisible}
+          onClose={() => setIsVisible(false)}
```
matthprost, line 110: You need to add a test on closing the popover

packages/ui/src/components/TagList/index.tsx
```diff
@@ -101,27 +104,35 @@ export const TagList = ({
...
+                  >
+                    {tag}
+                  </Tag>
+                ))}
+              </StyledTagContainer>
+            )
           }
         >
-          <StyledTooltipReference>
+          <StyledTooltipReference onClick={() => setIsVisible(true)}>
```
matthprost, line 132: Can we rename it differently as we do not use tooltip anymore?

matthprost: Also this line needs to be tested in unit testing

matthprost (changes requested):

packages/ui/src/components/TagList/index.tsx
```diff
@@ -51,8 +45,13 @@ type TagListProps = {
    * This property define maximum width of each tag. This doesn't apply for tags in tooltip.
    */
   multiline?: boolean
+  /**
+   * This propery define the title of the Popover, when some tags are hidden because of the threshold.
+   */
+  popoverTitle: string
   className?: string
   'data-testid'?: string
+  children?: ({ tags }: { tags: string[] }) => ReactNode
```
matthprost, line 54: to remove

matthprost (changes requested):

packages/ui/src/components/TagList/index.tsx
```diff
@@ -51,6 +45,10 @@ type TagListProps = {
    * This property define maximum width of each tag. This doesn't apply for tags in tooltip.
    */
   multiline?: boolean
+  /**
+   * This propery define the title of the Popover, when some tags are hidden because of the threshold.
```
matthprost, line 49:
```suggestion
   * This property define the title of the Popover, when some tags are hidden because of the threshold.
```

matthprost (changes requested):

matthprost (approved):

PetmePant (approved):

PR #2616, opened 2023-06-12
fix: support tooltipInfo on header cell (List & Table)

fabienhebert:
### Type

- Enhancement

### What has been done

Add a prop `tooltipInfo` on headerCell for both Table and List component.

This prop accepts a string and is optionel. If provided it add an info icon wrapped by a tooltip with the value as text

### Relevant logs and/or screenshots

Capture d’écran 2023-06-12 à 16 39 59


packages/ui/src/components/List/HeaderCell.tsx
```diff
@@ -56,6 +57,7 @@ type HeaderCellProps = {
   isOrdered?: boolean
   orderDirection?: 'asc' | 'desc' | 'none'
   onOrder?: (newOrder: 'asc' | 'desc') => void
+  tooltipInfo?: string
```
matthprost, line 60: Can we keep `tooltip` instead? That's the prop name we use for other components

packages/ui/src/components/Table/HeaderCell.tsx
```diff
@@ -70,6 +71,7 @@ type HeaderCellProps = {
   width?: string
   minWidth?: string
   maxWidth?: string
+  tooltipInfo?: string
```
matthprost, line 74: Same here, `tooltip` as prop name is better

packages/ui/src/components/List/HeaderCell.tsx
philibea: info sound better IMHO

matthprost: Let's go with `info` then 😄

JulienSaguez (approved):

matthprost (approved):

PetmePant (approved):

PR #2622, opened 2023-06-14
fix(Button): remove `iconOnly` from dom

matthprost:
## Type

- Bug

#### What is expected?

`iconOnly` prop is added into DOM while it shouldn't as it is an internal prop of Button component used for styling.


QuiiBz (approved):
We might need to update snapshots

chambo-e (approved):

PR #2666, opened 2023-06-27
feat(menu): retrieve onClick events on disclosure component

JulienSaguez:
## Type

- Feature

#### What is expected?

retrieve click event on disclosure function or component

### DIFF
--- a/packages/ui/src/components/Menu/index.tsx
+++ b/packages/ui/src/components/Menu/index.tsx
@@ -1,7 +1,13 @@
 import type { Theme } from '@emotion/react'
 import { css } from '@emotion/react'
 import styled from '@emotion/styled'
-import type { ReactElement, ReactNode, Ref } from 'react'
+import type {
+  ButtonHTMLAttributes,
+  MouseEvent,
+  ReactElement,
+  ReactNode,
+  Ref,
+} from 'react'
 import {
   cloneElement,
   forwardRef,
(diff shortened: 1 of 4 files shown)


philibea (approved):

packages/ui/src/components/Menu/index.tsx
```diff
@@ -183,15 +193,26 @@ const FwdMenu = forwardRef(
...
     return (
       <>
         {disclosure && (
           // @ts-expect-error reakit types are invalid, no need to pass as something, default is div
-          <PopoverDisclosure {...popover} ref={innerRef}>
+          <PopoverDisclosure
+            {...popover}
+            onClick={(e: MouseEvent<HTMLButtonElement>) => {
+              // This allow to use onClick for speficique case like traking.
+              target?.props?.onClick?.(e)
```
matthprost, line 212:
```suggestion
            onClick={(event: MouseEvent<HTMLButtonElement>) => {
              // This allow to use onClick for specific case like tracking.
              target?.props?.onClick?.(event)
```

we avoid using diminutive letters and use the full name of the variable

matthprost: @philibea made the mistake also, can you change it here too: `packages/ui/src/components/Modal/index.tsx`?

matthprost (changes requested):

matthprost (approved):

PR #2671, opened 2023-06-29
feat(Meter): refactor password strength meter to meter

philibea:
## Type

- Refactor

#### What is expected?

We need to refactor PasswordStrengthMeter and rename it into Meter to remove all score calculation and only provide a score directly. All this logic of score calculation should be done outside of the component, it should be more dumb.

We should keep both components and mark PasswordStrengthMeter as deprecated.


packages/ui/src/components/Meter/Meter.tsx
```diff
@@ -0,0 +1,104 @@
+import styled from '@emotion/styled'
+import type { JSX } from 'react'
+import { Text } from '../Text'
+
+const StyledTitle = styled(Text)`
+  display: inline-block;
+  vertical-align: top;
+  line-height: 22px;
```
matthprost, line 8: Why do we change this value here? It should be set in Text depending on variant you choose

packages/ui/src/components/Meter/Meter.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return (
+    <div
+      aria-live="polite"
+      aria-valuemax={strength.length ?? 0}
+      aria-valuemin={0}
+      aria-valuenow={value}
+      className={className}
+      data-testid={dataTestId}
+      id={id}
+      role="progressbar"
```
matthprost, line 79: Why not directly using tag `meter` in html and using the native props? I see you used aria-valuemin/max etc those could be the native props

packages/ui/src/components/PasswordStrengthMeter/index.tsx
```diff
@@ -76,6 +76,9 @@ const DEFAULT_ESTIMATE = () => ({ score: 0 as Score })
 const DEFAULT_FORBIDDEN_WORDS: PasswordStrengthMeterProps['forbiddenInputs'] =
   []
 
+/**
+ * @deprecated use Meter component instead
```
matthprost, line 80: You need to also mark it deprecated on storybook in file `packages/ui/src/components/PasswordStrengthMeter/__stories__/index.stories.tsx` with those parameters:

```ts
export default {
  component: PasswordStrengthMeter,
  parameters: {
      deprecated: true,
      deprecatedReason:
         'This component is deprecated, please use Meter instead.',
  },
  title: 'Components/Feedback/PasswordStrengthMeter',
} as Meta<typeof PasswordStrengthMeter>
```

matthprost (changes requested):

packages/ui/src/components/Meter/Meter.tsx
philibea: The meter element is a complex element to style. We have 2 options in this case: 
- We can hide visually the meter and use our own design
- We can try to style the meter but we will need to check with the Design Team because it's will be impossible to be 100% as the current design.

Or we can stay with a div that respect meter pattern ?

@matthprost

packages/ui/src/components/Meter/Meter.tsx
philibea: I removed the styled and used a Text as span direclty and much better 👍🏻

packages/ui/src/components/PasswordStrengthMeter/__stories__/index.stories.tsx
```diff
@@ -9,6 +9,11 @@ export default {
         component: 'Show strength of a password based on different criteria.',
       },
     },
+    parameters: {
```
matthprost, line 12: There is already parameters at higher level, this way won't work we don't see the component deprecated on storybook

matthprost (changes requested):

matthprost (approved):

packages/ui/src/components/Meter/Meter.tsx
matthprost: @fabienhebert @philibea we can't unify both components as there have different behavior and usages. But your comment makes me think we should change `role="progressbar"` to `role="meter"` instead

fabienhebert (approved):

PR #2781, opened 2023-08-01
feat(ui,form): add `RadioGroup` & `RadioFormGroup` components

QuiiBz:
## Type

- Feature

#### What is expected?

Add a new `RadioGroup` component in `@ultraviolet/ui` and a new `RadioGroupField` component in `@ultraviolet/form`

Screenshot 2023-08-07 at 14 14 42


packages/ui/src/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+    }),
+    [name, value, onChange, required],
+  )
+
+  return (
+    <RadioGroupContext.Provider value={contextValue}>
+      <Stack gap={1}>
+        <FieldSet className={className}>
+          <Stack gap={1.5}>
+            <Text as="p" variant="bodyStrong">
```
philibea, line 122: <Text as="legend"  is not possible ?

packages/ui/src/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+}: RadioGroupRadioProps) => {
+  const context = useContext(RadioGroupContext)
+
+  if (!context) {
+    throw new Error('RadioGroup.Radio can only be used inside a RadioGroup')
+  }
+
+  const { groupName, onChange, groupValue } = context
+
+  const radioName = `${groupName}-${name ?? ''}`
```
matthprost, line 51: If no name is provided, why not generating one?

matthprost (changes requested):

packages/ui/src/components/RadioGroup/index.tsx
matthprost: By the way can't we make it nested, like `${groupName}.${name ?? ''}` so in the form the value of the radio will be inside of the radio group name

packages/ui/src/components/RadioGroup/__stories__/Helper.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import { Template } from './Template.stories'
+
+export const Helper = Template.bind({})
```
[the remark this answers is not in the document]
QuiiBz: That would be the same as the `Controlled` story then?

packages/ui/src/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+  className?: string
+  helper?: ReactNode
+  error?: ReactNode
+  direction?: 'row' | 'column'
+  children: ReactNode
+} & Required<Pick<InputHTMLAttributes<HTMLInputElement>, 'onChange' | 'name'>> &
+  Pick<InputHTMLAttributes<HTMLInputElement>, 'required'>
+
+export const RadioGroup = ({
+  label,
```
matthprost, line 94: Can we rename it `legend` everywhere? This is not a label there is a difference between label and legend in HTML, we should follow the same naming.

matthprost (changes requested):

packages/ui/src/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+  )
+
+  return (
+    <RadioGroupContext.Provider value={contextValue}>
+      <Stack gap={1}>
+        <FieldSet className={className}>
+          <Stack gap={1.5}>
+            <Text as="p" variant="bodyStrong">
+              <legend>{label}</legend>
+            </Text>
```
matthprost, line 122: Let's use `<Text />` as `legend` and add some required icon, I did that for CheckboxGroup 😄 

PS: design was missing that part they will update ZH according to those specs

```suggestion
            <Text as="legend" variant="bodyStrong">
              {label}&nbsp;
              {required ? (
                <StyledRequiredIcon name="asterisk" color="danger" size={8} />
              ) : null}
            </Text>
```

matthprost (changes requested):

.changeset/friendly-baboons-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ultraviolet/ui': patch
+---
+
+Add `RadioGroup` component
```
matthprost, line 5: Shouldn't it be a minor ? On form we have minor for the same addition

matthprost (changes requested):

packages/form/src/components/RadioGroupField/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+      },
+    },
+  },
+  title: 'Form/Components/Fields/RadioGroupField',
+} as Meta
+
+export { Playground } from './Playground.stories'
+export { Error } from './Error.stories'
+export { Helper } from './Helper.stories'
+export { Direction } from './Direction.stories'
```
matthprost, line 27: Can you add a story with required prop and a submit button following what I have done on checkbox: https://github.com/scaleway/ultraviolet/blob/a3aa6381b656cb1a535926b0eb8f839602b94b14/packages/form/src/components/CheckboxGroupField/__stories__/Required.stories.tsx

So we see that when nothing is checked the submit is disabled until a selection is made

matthprost (changes requested):

packages/form/src/components/RadioGroupField/__stories__/index.stories.tsx
QuiiBz: Done:

Screenshot 2023-08-29 at 14 26 57

matthprost (approved):

PetmePant (approved):

PR #2782, opened 2023-08-01
refactor: update checkbox style

Lawndlwd:
## Summary

refactor style definition of checkbox and checkbox Field

## Type
- Refactor

to test the new design please check the [storybook](https://6117947ff8042f003af44001-agtfdutnqr.chromatic.com/?path=/docs/components-data-entry-checkbox--docs), and refer to the ticket on `DS-521`


packages/ui/src/components/Checkbox/index.tsx
```diff
@@ -258,61 +318,84 @@ export const Checkbox = forwardRef(
...
+                <Icon name="asterisk" size={10} color="danger" />
+              </sup>
+            ) : null}
+          </Row>
+
+          {helper ? (
+            <Text variant="bodySmall" as="p" prominence="weak" color="neutral">
+              {helper}
+            </Text>
           ) : null}
```
fabienhebert, line 390: mb we should also change the component to allow content to be well aligned with the label ? we may need a dedicated prop for the `label` and use children as optional content ?

Lawndlwd: Sorry, not sure to understand well
you want that we use label as label props and use the children’s props for optional content ? what is the matter ?

packages/ui/src/components/Checkbox/index.tsx
```diff
@@ -218,6 +277,7 @@ export const Checkbox = forwardRef(
       onBlur,
       error,
       name,
+      helper,
       value,
       size = 24,
```
matthprost, line 282: Can we change the default size to 20 so it follow the design specifications? Current checkbox is too big

Screenshot 2023-08-23 at 14 44 47

packages/ui/src/components/Checkbox/index.tsx
```diff
@@ -82,97 +88,143 @@ const CheckboxInput = styled('input', {
...
+      }
+    }
   }
 
-  ${CheckMark}, ${CheckMixedMark} {
+  ${CheckboxInput}:checked + ${StyledIcon} path {
     transform-origin: center;
-    transition: 200ms transform ease-in-out;
-    transform: scale(0);
+    transition: 100ms transform ease-in-out;
```
matthprost, line 182: Why changing the transition speed? 100ms seems to be quite fast

packages/ui/src/components/Checkbox/index.tsx
```diff
@@ -183,8 +235,15 @@ const StyledActivityContainer = styled.div`
...
   error?: string | ReactNode
+  /**
+   * @deprecated Size prop is deprecated and will be removed in next majour update.
+   */
   size?: number
+  /**
+   * @deprecated Progress prop is deprecated and will be removed in next majour update.
+   */
   progress?: boolean
+  helper?: string | ReactNode
```
matthprost, line 246:
```suggestion
  helper?: ReactNode
```

ReactNode includes string

packages/ui/src/components/Checkbox/index.tsx
```diff
@@ -183,8 +235,15 @@ const StyledActivityContainer = styled.div`
 
 type CheckboxProps = {
   error?: string | ReactNode
+  /**
+   * @deprecated Size prop is deprecated and will be removed in next majour update.
+   */
   size?: number
+  /**
+   * @deprecated Progress prop is deprecated and will be removed in next majour update.
```
matthprost, line 243:
```suggestion
   * @deprecated Size prop is deprecated and will be removed in next major update.
   */
  size?: number
  /**
   * @deprecated Progress prop is deprecated and will be removed in next major update.
```

matthprost (changes requested):

packages/ui/src/components/Checkbox/index.tsx
matthprost: My bad the sizes are correct

JulienSaguez (approved):

fabienhebert (approved):
Looks good to me, but I don't understand why adding `helper` in a lot of story not really related to it, it sounds confusing.

May be it will be better to have a dedicated story and remove others helper usage in stories

matthprost (approved):

PR #2804, opened 2023-08-09
fix: secondary sentiments on some components

fabienhebert:
## Type

- Enhancement

### Summarise concisely:

We added a `secondary` color/sentiment to the THEME however if wasn't accepted by some components


packages/ui/src/components/Text/index.tsx
```diff
@@ -159,7 +166,7 @@ export const Text = ({
         as={as}
         placement={placement}
         prominence={prominence}
-        color={color}
+        color={computedSentiment}
```
QuiiBz, line 169: Should we rename the prop too?

packages/ui/src/components/Badge/index.tsx
```diff
@@ -158,28 +158,53 @@ export const Badge = ({
...
+        <StyledText
+          variant="bodyStrong"
+          as="p"
+          fontSize={TEXT_SIZES[size]}
+          prominence={disabled ? 'weak' : 'default'}
+        >
+          {children}
+        </StyledText>
+      </StyledBox>
+      <StyledBox
```
QuiiBz, line 185: I'm not sure to understand why it's being duplicated?

QuiiBz (changes requested):

packages/ui/src/components/Text/index.tsx
fabienhebert: It's internal to the component but I can check indeed

packages/ui/src/components/Badge/index.tsx
fabienhebert: i forgot to remove this test 🙈

QuiiBz (approved):

philibea (approved):

PR #2808, opened 2023-08-10
test(Modal): add story for Modal with Tooltip

ModuloM:
## Summary
Add z-index to put modal backdrop above other components (like action bar)

## Type

- Bug

#### What is expected?

When there is both modal and action bar in a view, action bar should be now behind the backdrop.

| url  |  Screenshot 2023-08-10 at 11 35 24 | Screenshot 2023-08-10 at 11 35 03 |


QuiiBz (approved):
Does it work correctly when we have a tooltip in a modal?

packages/ui/src/components/Modal/index.tsx
```diff
@@ -211,6 +211,7 @@ const MemoDisclosure = memo(Disclosure)
 const StyledDialogBackdrop = styled(DialogBackdrop)`
   display: flex;
   position: fixed;
+  z-index: 9999;
```
fabienhebert, line 214: would it worth it to add kind of zindex constant file to keep some tracks of the differents index and make sure all make sense ? if bug fix is urgent we can merge like this and add a comment may be ?

fabienhebert (approved):

packages/ui/src/components/Modal/index.tsx
ModuloM: I've done as I've seen in other places. But actually you're right, I will prefer constants that can defined some "layers". So it will be easier to manage layers between each other. 

For now, I just can define a `LAYER_TOP` or something like that and set it up to 9999 (it will be then use as our top level layer). Wdyt?

matthprost: On a long term vision we are planing to remove all z-index as each portal element should be shown depending on when it appears on screen. That said, here the issue is mainly due to the way we handle the modal with reakit and we are also planing on removing reakit to do something homemade. This is a working solution for now and I don't know if it will be worth defining constant for z-index in the project.

matthprost (approved):

ModuloM:
There is indeed a problem with a modal that contains a tooltip
Screenshot 2023-08-29 at 11 40 40

ModuloM:
> There is indeed a problem with a modal that contains a tooltip Screenshot 2023-08-29 at 11 40 40

So we are good to go with this issue for now @matthprost?

packages/ui/src/components/Modal/__stories__/WithTooltip.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import type { StoryFn } from '@storybook/react'
+import { Button } from '../../Button'
+import { Tooltip } from '../../Tooltip'
+import { Modal } from '../index'
+
+export const WithTooltip: StoryFn = props => (
+  <Modal {...props} disclosure={<Button>With a Tooltip</Button>}>
+    <Tooltip
+      placement="top"
+      text="⚠️&nbsp;For now, this does not work, tooltip appear below the Modal!"
```
matthprost, line 10: Can you replace the text by something else like "the tooltip should be on top of the modal". Im afraid we will forget about this text

matthprost (changes requested):

matthprost (approved):

PR #2809, opened 2023-08-10
chore(deprecated): use Icons from icons packages

philibea:
## Type

- Refactor

#### What is expected?

Icon come from @ultraviolet/icons now 
Refactor usage of ui inside icon as it's was used only for theme. Remove cycle dependencies
Update test u on all component using Icons


chambo-e (commented):
This is a breaking change, needs to be planned for V2, we can live with them deprecated for a while :)

packages/ui/src/components/index.ts
```diff
@@ -55,3 +53,10 @@ export { ToastContainer, toast } from './Toaster'
 export { Toggle } from './Toggle'
 export { Tooltip } from './Tooltip'
 export { VerificationCode } from './VerificationCode'
+export {
+  /**
+   * @deprecated This component is deprecated. Please use `Icon` from `@ultraviolet/icons` instead.
+   * The component is same but has been move on another package.
```
QuiiBz, line 59:
```suggestion
   * The component is the same but has been moved to another package.
```

QuiiBz (changes requested):

QuiiBz (approved):

chambo-e (approved):

PR #2836, opened 2023-08-25
Draft: feat:  checkbox group & group field

Lawndlwd:
## Type

- Feature

#### What is expected?

Create new checkbox group and checkbox group field components


packages/ui/src/components/CheckboxGroup/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+import {
+  type ComponentProps,
+  type InputHTMLAttributes,
+  type ReactNode,
+  createContext,
+  useContext,
+  useMemo,
+} from 'react'
+import { Checkbox } from '../Checkbox'
+import { Icon } from '../Icon'
```
QuiiBz, line 11: Shouldn't we import this from `@ultraviolet/icons`?

QuiiBz:
There is no active state? (when we press without releasing)

QuiiBz (approved):

Lawndlwd (commented):
💜

JulienSaguez (approved):

PetmePant (changes requested):
@matthprost the only issue is the spacing between each toggle in the group component. It should be 16px instead of 8px

matthprost:
> @matthprost the only issue is the spacing between each toggle in the group component. It should be 16px instead of 8px

@PetmePant this PR is about checkbox group, I will change on toggle group

PR #2847, opened 2023-08-29
feat(ui): implement a new modal

DorianMaliszewski:
## Type

- Feature

#### What is expected?

Drop reakit usage in our modals and keep behaviour. Moreover, it fix the close icon container because it was above the text and prevent user events at the top of the modal

1. Start from scratch a new modal

2. 

3.


packages/ui/src/components/Modal/Dialog.tsx
```diff
@@ -0,0 +1,151 @@
...
+    opacity: 1;
+  }
+`
+
+type StyledDialogProps = {
+  'data-size': ModalSize
+  'data-placement': ModalPlacement
+}
+
+const StyledDialog = styled.dialog<StyledDialogProps>`
```
chambo-e, line 34: Support is relatively recent on Safari, do we have a way to check if this is going to be ok with our user base ?

matthprost: @chambo-e there is 93% of the global users that are compatible with this, what is our minimum in order to use it?

Source: https://caniuse.com/?search=dialog

packages/ui/src/components/Modal/Disclosure.tsx
```diff
@@ -0,0 +1,28 @@
...
+      visible,
+      onClose: handleClose,
+      toggle,
+      onOpen: handleOpen,
+      modalId: id,
+    })
+  ) : (
+    <DisclosureContainer onClick={handleOpen} type="button" aria-controls={id}>
+      {disclosure}
+    </DisclosureContainer>
```
matthprost, line 27: There is an issue with this, we are creating a button wrapper around the disclosure which is also a button. The result is 2 buttons in each others in the dom:

Screenshot 2023-08-30 at 11 23 28

matthprost (changes requested):

packages/ui/src/components/Modal/Dialog.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const stopClick: MouseEventHandler = useCallback(event => {
+    event.stopPropagation()
+  }, [])
+
+  // Stop key up : used when having inputs in modals
+  const stopKeyUp: KeyboardEventHandler = useCallback(event => {
+    event.stopPropagation()
+  }, [])
+
+  return createPortal(
```
matthprost, line 130: The create portal should be dynamic when the user open the modal or the dom will be full of useless div. You can refer to Popup internal component which is used by tooltip

Screenshot 2023-08-30 at 11 37 46

matthprost (changes requested):

packages/ui/src/components/Modal/Disclosure.tsx
```diff
@@ -0,0 +1,46 @@
...
+  useEffect(() => {
+    const element = disclosureRef.current
+    const handleClick = () => {
+      handleOpen()
+    }
+
+    element?.addEventListener('click', handleClick)
+
+    return () => {
+      element?.removeEventListener('click', handleClick)
```
QuiiBz, line 24: So we can directly use `handleOpen` instead of having another function?
```suggestion
    element?.addEventListener('click', handleOpen)

    return () => {
      element?.removeEventListener('click', handleOpen)
```

QuiiBz (changes requested):

QuiiBz:
Need a snapshot update:

https://github.com/scaleway/ultraviolet/actions/runs/6023641523/job/16340773711?pr=2847#step:8:252

matthprost (approved):

QuiiBz (approved):

PR #2855, opened 2023-09-01
ci: add github action to auto assign when pull request is created

matthprost:
## Type

- CI

#### What is expected?

In our current workflow the creator of the PR is the assigned to it. With this rule we can safely put a github action that auto assign the PR to the author in case we forget to do it.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,0 +1,10 @@
+name: 'Auto Assign Pull Request'
+on:
+  pull_request:
+    types: [opened, ready_for_review]
+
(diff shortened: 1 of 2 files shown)


.github/workflows/action.yml
```diff
@@ -0,0 +1,10 @@
+name: 'Auto Assign Pull Request'
```
QuiiBz, line 1: We don't need quotes around the name

QuiiBz (approved):

chambo-e (approved):

PR #2885, opened 2023-09-06
fix: dialog add attributes

DorianMaliszewski:
## Type

- Enhancement

#### What is expected?

Modal should have aria-modal=true but actionar shouldn't. Additionally pass more aria props to disclosure

(Describe what you did)

1.

2.

3.


matthprost (approved):

fabienhebert (approved):

fabienhebert:
sounds fine but can't we considerer that ActionBar is kind of modal ?

PR #2889, opened 2023-09-07
fix(modal): handle events and focus

DorianMaliszewski:
## Type

- Enhancement

#### What is expected?

Handle events when modal is open to trap focus and prevent elements outisde of the modal being impacted 

(Describe what you did)

1.

2.

3.


packages/ui/src/components/Modal/Dialog.tsx
```diff
@@ -75,13 +80,15 @@ export const Dialog = ({
   backdropCss,
 }: DialogProps) => {
   const containerRef = useRef(document.createElement('div'))
+  const dialogRef = useRef(document.createElement('dialog'))
```
QuiiBz, line 83: Why do we have to create an element on every render?

```suggestion
  const containerRef = useRef<HTMLElement>(null)
  const dialogRef = useRef<HTMLElement>(null)
```

matthprost (approved):

vincentaudebert (approved):

QuiiBz (approved):

PR #2905, opened 2023-09-11
ci(changeset): add examples project into ignore

matthprost:
There is an issue when renovate tries to update package in examples folder it generate a changeset out of it while it shouldn't. To avoid that I added those projects into ignore of the config file.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -12,5 +12,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@examples/next-advanced", "@examples/next-simple"]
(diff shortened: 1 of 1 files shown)


.changeset/config.json
```diff
@@ -12,5 +12,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": ["@examples/next-advanced", "@examples/next-simple"]
```
chambo-e, line 15: Can't we have a wildcard ?

matthprost: @chambo-e I replaced by: `"ignore": ["@examples/*"]`

PR #2915, opened 2023-09-14
fix(Tooltip): avoid breaking all words and break only when necessary

matthprost:
## Type

- Bug

#### What is expected?

We use to break all words on tooltip in case the content is too big but it cause problems as it will also break words in the middle of them as reported here:

Screenshot 2023-09-13 at 10 02 21

To solve this we removed CSS property `word-break: break-all` and replaced it by `overflow-wrap: anywhere`. This is the correct way to do this for this kind of situation.

Examples:

Screenshot 2023-09-14 at 10 30 53
Screenshot 2023-09-14 at 10 31 17

### DIFF
--- a/packages/ui/src/internalComponents/Popup/index.tsx
+++ b/packages/ui/src/internalComponents/Popup/index.tsx
@@ -67,7 +67,7 @@ const StyledTooltip = styled('div', {
   text-align: center;
   position: absolute;
   max-width: ${({ maxWidth }) => maxWidth}px;
-  overflow-wrap: anywhere;
+  overflow-wrap: break-word;
   font-size: 0.8rem;
   inset: 0 auto auto 0;
   top: 0;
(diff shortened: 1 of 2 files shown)


vincentaudebert (approved):

JulienSaguez (approved):

packages/ui/src/internalComponents/Popup/index.tsx
```diff
@@ -67,7 +67,7 @@ const StyledTooltip = styled('div', {
   text-align: center;
   position: absolute;
   max-width: ${({ maxWidth }) => maxWidth}px;
-  word-break: break-all;
+  overflow-wrap: anywhere;
```
philibea, line 70:
```suggestion
  overflow-wrap: break-word;
```

this is not this one ?

philibea: it's because inside a tooltip we don't want that our word break ?

philibea (approved):

PR #2918, opened 2023-09-14
feat(ui): add MenuV2

QuiiBz:
## Type

- Feature

#### What is expected?

Add a new `MenuV2` component that uses the internal `Popup` component to replace the usage of `reakit`. We have to introduce a new component and deprecate the old one because the types are a bit different, especially the `placement` prop.

1. Add `MenuV2` component with tests and stories similar to `Menu`

2. Deprecated `Menu` component

3. Add migration doc


QuiiBz:
Needs https://github.com/scaleway/ultraviolet/pull/2920

packages/ui/src/components/MenuV2/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+      // Focus the disclosure when the menu is closed
+      disclosureRef.current?.focus()
+    }
+
+    const finalDisclosure = cloneElement(target, {
+      onClick: toggleVisible,
+      'aria-haspopup': 'dialog',
+      'aria-expanded': isVisible,
+      // @ts-expect-error not sure how to fix this
```
philibea, line 137: why ?

packages/ui/src/components/MenuV2/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+      placement = 'bottom',
+      visible = false,
+      className,
+      'data-testid': dataTestId,
+    }: MenuProps,
+    ref: Ref<HTMLButtonElement | null>,
+  ) => {
+    const [isVisible, setIsVisible] = useState(visible)
+    const popupRef = useRef<HTMLDivElement>(null)
+    const disclosureRef = useRef<HTMLDivElement>(null)
```
philibea, line 99: disclosure should not be an htmlDivElement ? It's always a Button ?

packages/ui/src/components/MenuV2/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const onClose = () => {
+      setIsVisible(false)
+
+      // Focus the disclosure when the menu is closed
+      disclosureRef.current?.focus()
+    }
+
+    const finalDisclosure = cloneElement(target, {
+      onClick: toggleVisible,
+      'aria-haspopup': 'dialog',
```
philibea, line 135:
```suggestion
      'aria-haspopup': 'menu',
```

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup

QuiiBz: That's not what reakit uses:

Screenshot 2023-09-18 at 09 38 27

packages/ui/src/components/MenuV2/index.tsx
QuiiBz: `cloneElement` doesn't seem to accept `ref` as a prop

matthprost (changes requested):

matthprost (approved):

packages/ui/src/components/MenuV2/Item.tsx
```diff
@@ -0,0 +1,152 @@
...
+      className,
+      'data-testid': dataTestId,
+    },
+    ref,
+  ) => {
+    if (href && !disabled) {
+      return (
+        <Tooltip text={tooltip}>
+          <StyledLinkItem
+            borderless
```
ModuloM, line 112: Do we always want that Link Items are borderless?

packages/ui/src/components/MenuV2/__stories__/Sentiments.stories.tsx
```diff
@@ -0,0 +1,40 @@
+import type { StoryFn } from '@storybook/react'
+import { MenuV2 } from '..'
+import { DefaultDisclosure } from './Template.stories'
+
+export const Sentiments: StoryFn<typeof MenuV2> = () => (
+  <MenuV2 disclosure={DefaultDisclosure}>
+    <MenuV2.Item>default disabled</MenuV2.Item>
+    <MenuV2.Item sentiment="danger">Danger disabled</MenuV2.Item>
```
ModuloM, line 8: You've switch disable texts for 'default' and 'danger' is this examples

packages/ui/src/components/MenuV2/__stories__/WithModal.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <div style={{ padding: 32 }}>
+        Content should be present in center of the modal
+      </div>
+    </Modal>
+  )
+
+  return (
+    <MenuV2 disclosure={DefaultDisclosure}>
+      <MenuV2.Item>Menu Item</MenuV2.Item>
+      <MenuV2.Item href="/?path=/docs/components-navigation-menu--modal">
```
ModuloM, line 21: It's not very important but the link seems broken

`/?/?path=/docs/components-navigation-menu"`

packages/ui/src/components/MenuV2/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    const onClose = () => {
+      setIsVisible(false)
+
+      // Focus the disclosure when the menu is closed
+      disclosureRef.current?.focus()
+    }
+
+    const finalDisclosure = cloneElement(target, {
```
ModuloM, line 133: 👍

ModuloM (approved):

packages/ui/src/components/MenuV2/__stories__/Sentiments.stories.tsx
QuiiBz: It's only borderless when there is an `href`, similar to `Menu` so I guess it's fine.

packages/ui/src/components/MenuV2/Item.tsx
QuiiBz: It's only borderless when there is an href, similar to Menu so I guess it's fine.

PR #2920, opened 2023-09-14
fix(Tooltip): implement esc key without focus

matthprost:
## Type

- Enhancement

#### What is expected?

We use to have ESC key handled on tooltip but only on focus, it is now handled without focus too. The same features is implemented in Popover as it uses Popup.

### DIFF
--- a/packages/ui/src/internalComponents/Popup/index.tsx
+++ b/packages/ui/src/internalComponents/Popup/index.tsx
@@ -215,6 +215,20 @@ export const Popup = forwardRef(
     }, [onScrollDetected])
 
     /**
+     * This function is called when we need to hide tooltip. A timeout is set to allow animation end, then remove
+     * tooltip from dom.
+     */
+    const hideTooltip = useCallback(() => {
+      debounceTimer.current = setTimeout(() => {
+        setReverseAnimation(true)
+        timer.current = setTimeout(() => {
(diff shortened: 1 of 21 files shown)


packages/ui/src/internalComponents/Popup/index.tsx
```diff
@@ -291,6 +293,40 @@ export const Popup = forwardRef(
...
+        if (event.key === 'Escape') {
+          event.preventDefault()
+          event.stopPropagation()
+          debounceTimer.current = setTimeout(() => {
+            setReverseAnimation(true)
+            timer.current = setTimeout(() => {
+              unmountTooltip()
+              onClose?.()
+            }, ANIMATION_DURATION)
+          }, 200)
```
QuiiBz, line 308: Could we avoid duplicating this whole block (lines 227-233)?

QuiiBz (changes requested):

packages/ui/src/internalComponents/Popup/index.tsx
matthprost: @QuiiBz fixed!

QuiiBz (approved):

Lawndlwd (approved):

JulienSaguez (approved):

PR #2938, opened 2023-09-19
refactor(Notice): changed to span and reduced icon size

matthprost:
## Type

- Refactor

#### What is expected?

Current Notice is a paragraph while it should be a span avoiding the margin top and bottom of the paragraph. Plus we reduced the size of the icon fitting better into a TextInput.

| url  | Screenshot 2023-09-19 at 16 32 01 | Screenshot 2023-09-19 at 16 32 12 |
| url  | Screenshot 2023-09-19 at 16 32 35 | Screenshot 2023-09-19 at 16 32 40 |
| url  | Screenshot 2023-09-19 at 16 33 13 | Screenshot 2023-09-19 at 16 33 29 |

### DIFF
--- a/.changeset/pretty-cameras-push.md
+++ b/.changeset/pretty-cameras-push.md
@@ -1,0 +1,6 @@
+---
+'@ultraviolet/form': minor
+'@ultraviolet/ui': minor
+---
+
+Refactoring of Notice component to be a span with smaller icon so it fits better in TextInput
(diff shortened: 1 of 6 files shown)


QuiiBz (approved):
It's technically a breaking change since we change the semantic of the component, right?

philibea (approved):

matthprost:
> It's technically a breaking change since we change the semantic of the component, right?

@QuiiBz Technically it should yes, in reality here the change is so little and improve accessibility so I would consider it as minor.

PR #2947, opened 2023-09-22
feat: theme generator

matthprost:
## Type

- Feature

## Introduction: Theme Generator ✨
There is around 175 token colors in Ultraviolet theme. It can be hard to set each of them one by one and understanding their usage in the components. Theme generator is here to help you create a new theme for your project FAST 🚀.

### How does it work?
You basically just need to fill this form with the name and color of each sentiment you want to use in your project. The color you will set will be the most used color in the theme and the other colors will be generated from it.

It then provide a page with most of our components as example with the new generated theme applied onto it. You can copy/paste the JSON of the theme or even get a React example.

### Other changes

Button now support `onMouseUp`, `onMouseDown`, `onMouseOut` props.

### Demo

https://github.com/scaleway/ultraviolet/assets/15812968/3237fc6d-29d8-4e94-9047-6c22a5b22874


packages/ui/src/__stories__/Tools/ThemeGenerator/helpers.ts
ModuloM, line 1: Empty ?

matthprost: @ModuloM my bad forgot to move the helpers here, done!

packages/ui/src/__stories__/Tools/ThemeGenerator/ThemeResult/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { CodeIntegration } from './CodeIntegration'
+import { Demo } from './Demo' // For some reason Global doesn't work here this is the workaround I found
+
+// For some reason Global doesn't work here this is the workaround I found
+const Container = styled.div`
+  ${normalize()}
+`
+
+type ThemeResultProps = {
+  theme: SCWUITheme
```
QuiiBz, line 17: Not related to this PR, but shouldn't we rename `SCWUITheme` to something like `UltravioletUITheme`?

packages/ui/src/__stories__/Tools/ThemeGenerator/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+    setStep(1)
+  }, [])
+
+  return (
+    <Stack gap={4}>
+      {step === 0 ? (
+        <Stack gap={2}>
+          <Stack>
+            <Text variant="headingLarge" as="h1">
+              ✨Theme generator
```
QuiiBz, line 114:
```suggestion
              ✨ Theme generator
```

QuiiBz (changes requested):

ModuloM (approved):

packages/ui/src/__stories__/Tools/ThemeGenerator/ThemeResult/index.tsx
matthprost: @QuiiBz Good point, I will add a new type instead to avoid breaking changes

packages/ui/src/__stories__/Tools/ThemeGenerator/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+          (secondAcc, shadeKey, index) => ({
+            [shadeKey]: {
+              value: generateShadeContrast(shadeKey, value, index),
+              type: 'color',
+            },
+            ...secondAcc,
+          }),
+          {},
+        ),
+        ...acc,
```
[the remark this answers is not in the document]
matthprost: @ModuloM I'm not sure there is a use for unit testing this if it's checking that the output is the same in both case. I will dig

packages/ui/src/__stories__/Tools/ThemeGenerator/ThemeResult/index.tsx
matthprost: https://github.com/scaleway/ultraviolet/pull/2967

QuiiBz (approved):

PR #2982, opened 2023-10-03
feat: new library for advanced components

matthprost:
## Type

- Feature

### ✨ New Library for complex component

Ultraviolet UI is our simple library for basic components with very low composition. We needed one more library to increase the complexity of our component that couldn't fit there.

One first component is added into it: ContentCard.
A simple component to display information with images, title, description, icon or anything else.

Screenshot 2023-10-03 at 15 45 04

Screenshot 2023-10-06 at 11 55 54


PetmePant (dismissed):
Storybook related:

1) Add a fixed width in the preview component so is not huge
2) Add some custom stories so we showcase the possibilities to our users (slot/children)
3) There is an issue that you can hide image/icon even if you dont have any content in the prop

Component related:
1) Fix icon sizes so it's the same for both directions
2) Add the option of loading state (check specifications)
3) Broder on hover state should be primary

ModuloM (approved):

packages/plus/src/components/ContentCard/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+  title: string
+  description?: string
+  children?: ReactNode
+  /**
+   * The href to link the card to. If not provided, the card will not be clickable.
+   */
+  href?: HTMLAnchorElement['href']
+  target?: HTMLAnchorElement['target']
+  onClick?: MouseEventHandler<HTMLElement>
+  loading?: boolean
```
vincentaudebert, line 96: should we forward className in case of extra styling?

vincentaudebert (approved):
Very very nice work. Love it!

packages/plus/src/components/ContentCard/index.tsx
matthprost: @vincentaudebert Right! I forgot about it, thanks!

JulienSaguez (approved):

PetmePant (changes requested):
Some issues appeared while on review:

1) Height is not adjusted based on the content that is inside 
2)There is a hidden gap that creates a white space at the bottom of the component
3) Description and title seem to broke when a lot content is added

Screenshot 2023-10-10 at 1 00 40 PM
633/8729a2d6-3dc1-40c3-bc85-fbfff0510d5f)
Screenshot 2023-10-10 at 12 52 37 PM

PetmePant (approved):

PR #2984, opened 2023-10-03
feat: add estimate cost into complex component library

matthprost:
## Type

- Feature

#### What is expected?

New component `EstimateCost`.

Estimate cost is a complex component allowing you to display an estimation of a price for a given service. This is quite useful to allow client estimate depending on time what they are going to pay for.

The component is experimental and some stuff will change.

In order to allow different currencies and languages we expose multiple props:

- `locales`: it allows you to change all locales in the component itself, an example of available locales is shown in storybook in "Default locales" section.
- `currency`: it allows you to change the currency according to ISO 4217 currency codes (https://en.wikipedia.org/wiki/ISO_4217#List_of_ISO_4217_currency_codes)
- `numberLocales`: this if for the formatting of numbers that are shown on components. Each country shows number in a different way and it can be customise with this prop

Screenshot 2023-11-02 at 17 31 59

Screenshot 2023-11-02 at 17 37 22
Screenshot 2023-11-02 at 17 37 42
Screenshot 2023-11-02 at 17 38 01
Screenshot 2023-11-02 at 17 38 27
Screenshot 2023-11-02 at 17 39 56


packages/plus/package.json
```diff
@@ -30,9 +30,10 @@
   ],
   "sideEffects": false,
   "type": "module",
+  "types": "./dist/index.d.ts",
```
philibea, line 33: why, it's should not be necessary !

packages/plus/package.json
```diff
@@ -43,20 +44,21 @@
   "devDependencies": {
     "@babel/core": "7.23.2",
     "@emotion/babel-plugin": "11.11.0",
+    "@emotion/react": "11.11.1",
+    "@emotion/styled": "11.11.0",
```
philibea, line 48: not the same version as `@emotion/react` ?

packages/plus/src/components/EstimateCost/constants.ts
```diff
@@ -0,0 +1,33 @@
+import type { Units } from './types'
+
+export const HOURS_IN_MONTH = 730
```
philibea, line 3: APPROXIMATIVE_HOURS_IN_MONTH

packages/plus/src/components/EstimateCost/EstimateCostProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  formatNumber: (number: number, options: FormatNumberOption) => string
+}>({ locales: EstimateCostLocales, formatNumber: () => '' })
+
+export const useEstimateCost = () => useContext(EstimateCostContext)
+
+type EstimateCostProviderProps = {
+  children: ReactNode
+  locales?: typeof EstimateCostLocales
+  currency?: string
+  numberLocales?: string
```
philibea, line 17: i think we can better type this one ?

philibea: And maybe add a comment for our user where and why it's used ?

matthprost: @philibea 
I took the type from `Intl.NumberFormat` which is a `string`, should I import an external lib for getting all of them?
Screenshot 2023-11-03 at 10 49 57

I already described it into comment on types of each of them, it should be visible when checking JSDoc but also on storybook
Screenshot 2023-11-03 at 10 51 31

packages/plus/package.json
matthprost: @philibea this is to be consistent between all package.json config among the repository

packages/plus/package.json
matthprost: @philibea this is the last version available for `@emotion/styled`
> Source: https://www.npmjs.com/package/@emotion/styled?activeTab=versions

Lawndlwd (approved):

fabienhebert (approved):

packages/plus/src/components/EstimateCost/EstimateCostProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  children,
+  locales,
+  currency,
+  numberLocales,
+}: EstimateCostProviderProps) => {
+  const newLocales = locales
+    ? deepMerge(EstimateCostLocales, locales)
+    : EstimateCostLocales
+
+  const formatNumber = useCallback(
```
philibea, line 42: this function is a duplicate of what we have into scaleway/use-i18n right ?

packages/plus/src/components/EstimateCost/OverlayComponent.tsx
```diff
@@ -0,0 +1,161 @@
...
+  height: 120px;
+  background-color: ${({ theme }) => theme.colors.neutral.background};
+  margin: 0 0 0 200px;
+  display: flex;
+  justify-content: center;
+  box-shadow: ${({ inView, theme }) =>
+    inView ? '0' : theme.shadows.defaultShadow};
+  transition:
+    bottom 0.3s,
+    box-shadow 0.3s;
```
philibea, line 29: should we use our theme here ?

packages/plus/src/components/EstimateCost/OverlayComponent.tsx
```diff
@@ -0,0 +1,161 @@
...
+`
+
+const List = styled.ul`
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  list-style: none;
+  margin: 0;
+  padding: 24px 0;
```
philibea, line 38: use theme here ?

packages/plus/src/components/EstimateCost/OverlayComponent.tsx
```diff
@@ -0,0 +1,161 @@
...
+        <List>
+          {OverlayLeft ? (
+            <SideItem>
+              <OverlayLeft disabled={disableOverlayLeft}>
+                {locales['estimate.cost.submit.label']}
+              </OverlayLeft>
+            </SideItem>
+          ) : null}
+          {Children.map(list, (child, index) =>
+            /* @ts-expect-error I'm too dumb to understand this sorcery */
```
philibea, line 114: you can remove this ts-expect error with a check of the child with isValidElement.

It's used inside Breadcrumbs, MenuV2 for exemple

packages/plus/src/components/EstimateCost/locales/en.ts
```diff
@@ -0,0 +1,22 @@
+export default {
```
philibea, line 1: should we add localazy into the CI for this one ?

packages/plus/src/components/EstimateCost/EstimateCostProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export const EstimateCostProvider = ({
+  children,
+  locales,
+  currency,
+  numberLocales,
+}: EstimateCostProviderProps) => {
+  const newLocales = locales
+    ? deepMerge(EstimateCostLocales, locales)
```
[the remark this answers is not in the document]
matthprost: @philibea good point hahaha

packages/plus/src/components/EstimateCost/locales/en.ts
matthprost: @philibea Im not sure there is a need for it as we are not planning to provide any other language than english

packages/plus/src/components/EstimateCost/EstimateCostProvider.tsx
matthprost: @philibea not same but different but still the same..

Yeah it's a bit different as it focus on currency only and I cannot import the one from i18n as it needs a provider so I made home made one for this component but nothing crazy I only use Intl.NumberFormat at it's most raw usage.

packages/plus/src/components/EstimateCost/OverlayComponent.tsx
matthprost: @philibea I think you're mixing with normal box-shadow? here it's a transition

philibea (approved):

PR #2988, opened 2023-10-04
fix: event workflow not consistent

DorianMaliszewski:
## Type

- Bug

#### What is expected?

Handle close/show event properly do not have waterfall effects

(Describe what you did)

1.

2.

3.

### DIFF
--- a/packages/ui/src/components/MenuV2/__stories__/WithModal.stories.tsx
+++ b/packages/ui/src/components/MenuV2/__stories__/WithModal.stories.tsx
@@ -1,6 +1,7 @@
 import type { StoryFn } from '@storybook/react'
 import { MenuV2 } from '..'
 import { Modal } from '../../Modal'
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/MenuV2/__stories__/WithModal.stories.tsx
```diff
@@ -12,6 +13,7 @@ export const WithModal: StoryFn<typeof MenuV2> = () => {
       <div style={{ padding: 32 }}>
         Content should be present in center of the modal
       </div>
+      <TextInput name="test" />
```
matthprost, line 16: Can you add a label also, so we can know what is it about in the story

matthprost (changes requested):

matthprost (approved):

ModuloM (approved):

PR #2994, opened 2023-10-06
refactor(form): migrate to react hook form

johnrazeur:
## Type

- Migration
- Refactor

#### What is expected?

Migrate final-form to react-hook-form.


packages/form/src/components/Form/index.tsx
```diff
@@ -1,75 +1,146 @@
...
+
+  const methods = !methodsProp ? methodsHook : methodsProp
+
+  const handleSubmit = methods.handleSubmit(async values => {
+    const result = await onRawSubmit(values, {
+      reset: methods.reset,
+      resetFieldState: methods.resetField,
+      restart: () => methods.reset(initialValues),
+      change: methods.setValue,
+    })
```
DorianMaliszewski, line 105: Do we use somewhere the second parameter ?

DorianMaliszewski: May be it can be useful in the future but if it is useless actually remove it we will add it later when needed

packages/form/src/components/SelectInputField/index.tsx
```diff
@@ -96,84 +94,81 @@ type SelectInputOptionOrGroup = NonNullable<SelectInputOptions>[number]
...
-    | 'value'
-    | 'noTopLabel'
-    | 'noOptionsMessage'
-    | 'customStyle'
-    | 'data-testid'
+  TFieldValues extends FieldValues = FieldValues,
+> = BaseFieldProps<TFieldValues> &
+  Partial<
+    Pick<
+      SelectInputProps,
```
DorianMaliszewski, line 101: May be Omit is a better choice here ?

packages/form/src/components/FormSpy/index.tsx
```diff
@@ -0,0 +1,24 @@
+import { useEffect } from 'react'
```
DorianMaliszewski, line 1: Why we need this ?

packages/form/src/index.ts
```diff
@@ -24,8 +15,19 @@ export {
...
+export {
+  useFormStateDeprecated,
+  useFieldDeprecated,
+  useFormDeprecated,
+  useFieldArrayDeprecated,
+  useOnFieldChange,
+} from './hooks'
+
+// eslint-disable-next-line no-restricted-syntax
+export * from 'react-hook-form'
```
DorianMaliszewski, line 31: IMO You should be explicit to follow the code guildeline we actually have

packages/form/src/hooks/useFormState.ts
```diff
@@ -0,0 +1,29 @@
...
+import { useFormContext } from 'react-hook-form'
+
+type UseFormStateParams = {
+  subscription?: Record<string, boolean>
+}
+
+/**
+ * @deprecated
+ */
+export const useFormStateDeprecated = <TFieldValues extends FieldValues>(
```
DorianMaliszewski, line 11: We cant have useXDeprecated everywhere it should be retrocompatible with actual useFormState, yes I know it can be redundant but the migration will be easier

DorianMaliszewski (changes requested):

packages/form/src/components/Form/index.tsx
DorianMaliszewski: -> Need to be documented as deprecated :)

DorianMaliszewski: Need to be removed and refactor our forms that use it

PR #3011, opened 2023-10-11
feat(dateinput): update design and add range date

JulienSaguez:
## Summary
update design and add range date
## Type
- Feature

For the reference, check the figma https://www.figma.com/file/XHoO4V9LOFEFkAZsocYJ83/Ultraviolet%3A-Core-Components?type=design&node-id=7493%3A121027&mode=design&t=KHeUecaJBVlUWfK4-1 ant the ticket https://jira.infra.online.net/browse/DS-579

### DIFF
--- a/.changeset/fair-boxes-carry.md
+++ b/.changeset/fair-boxes-carry.md
@@ -2,4 +2,6 @@
 '@ultraviolet/ui': minor
 ---
 
-update design and add range date
+Component `<DateInput />` updates:
+- Design update
+- New prop available `selectRange` allowing the user to select a range of dates
--- a/.changeset/fair-boxes-carry.md
+++ b/.changeset/fair-boxes-carry.md
@@ -1,0 +1,5 @@
+---
+'@ultraviolet/ui': minor
+---
+
+update design and add range date
--- a/packages/ui/src/components/DateInput/__stories__/Exclude.stories.tsx
+++ b/packages/ui/src/components/DateInput/__stories__/Exclude.stories.tsx
@@ -1,0 +1,21 @@
+import { Template } from './Template'
+
+export const Exclude = Template.bind({})
+
+Exclude.parameters = {
+  docs: {
+    storyDescription:
+      'With `excludeDates` you can define a array of dates to exclude',
+  },
+}
+
+Exclude.args = {
+  label: 'Date',
+  value: new Date('December 13, 1995 03:24:00'),
+  excludeDates: [
+    new Date('December 1, 1995 03:24:00'),
+    new Date('December 14, 1995 03:24:00'),
+    new Date('December 22, 1995 03:24:00'),
+    new Date('December 28, 1995 03:24:00'),
+  ],
+}
--- a/packages/ui/src/components/DateInput/__stories__/Exclude.stories.tsx
+++ b/packages/ui/src/components/DateInput/__stories__/Exclude.stories.tsx
@@ -4,8 +4,9 @@ export const Exclude = Template.bind({})
 
 Exclude.parameters = {
   docs: {
-    storyDescription:
-      'With `excludeDates` you can define a array of dates to exclude',
+    description: {
+      story: 'With `excludeDates` you can define a array of dates to exclude',
+    },
   },
 }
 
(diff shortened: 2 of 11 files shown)


packages/ui/src/components/DateInput/index.tsx
```diff
@@ -204,6 +238,17 @@ export const DateInput = ({
...
+  const valueStart = `${
+    startDate !== undefined && startDate !== null ? format(startDate) : ''
+  }`
+  const valueEnd = `${
+    endDate !== undefined && endDate !== null ? format(endDate) : ''
+  }`
+
+  const valueFormat = selectRange
+    ? `${valueStart} - ${valueEnd}`
+    : format(value)
```
matthprost, line 250: Here you need to check if `valueStart` and `valueEnd` exists or if they don't we have an input containing only `-`

Screenshot 2023-10-12 at 10 19 29

packages/ui/src/components/DateInput/__stories__/Range.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import { Template } from './Template'
+
+export const Range = Template.bind({})
```
matthprost, line 3: Could you make this story like controlled so we can test the component on storybook? I know designers will want to test it haha

.changeset/fair-boxes-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ultraviolet/ui': minor
+---
+
+update design and add range date
```
matthprost, line 5: You can precise the prop that has been added and to what component, for example:

```suggestion
Component `<DateInput />` updates:
- Design update
- New prop available `selectRange` allowing the user to select a range of dates
```

packages/ui/src/components/DateInput/__stories__/Range.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+Range.parameters = {
+  docs: {
+    storyDescription:
+      'With `startDate` and `endDate` you can define a range of dates',
+  },
+}
```
matthprost, line 10: This is the old way to describe a story please use this instead:

```suggestion
Range.parameters = {
  docs: {
    description: {
          story: 'With `startDate` and `endDate` you can define a range of dates',
      }
  },
}
```

packages/ui/src/components/DateInput/__stories__/Exclude.stories.tsx
```diff
@@ -0,0 +1,21 @@
+import { Template } from './Template'
+
+export const Exclude = Template.bind({})
+
+Exclude.parameters = {
+  docs: {
+    storyDescription:
+      'With `excludeDates` you can define a array of dates to exclude',
+  },
+}
```
matthprost, line 10: Same goes for here:

```suggestion
Exclude.parameters = {
  docs: {
    description: {
        story: 'With `excludeDates` you can define a array of dates to exclude',
      }
  },
}
```

packages/ui/src/components/DateInput/__tests__/index.test.tsx
```diff
@@ -97,4 +97,15 @@ describe('DateInput', () => {
...
+  test('render correctly with a range of date', () =>
+    shouldMatchEmotionSnapshot(
+      <DateInput
+        label="Date"
+        selectRange
+        startDate={new Date('December 12, 1995 03:24:00')}
+        endDate={new Date('December 25, 1995 03:24:00')}
+        onChange={() => {}}
+      />,
+    ))
```
matthprost, line 110: You need to add another test for `excludeDates` prop too or the testing coverage will go down

matthprost (changes requested):

PetmePant (changes requested):
1) There is an issue when you swap months with the button navigations. It rearranges like small portion of the location and it changes the width of the picker. The month text shouldn't change the container width or change the button's location.

2) When range is enabled the selected dates looks a bit weird with some numeric numbers that we should hide.

3) need to change prop name from selectRange to range 

Notes:
*I couldn't check the range functionality 
*Overall styles look like design, perfect


Screenshot 2023-10-12 at 6 04 04 PM

https://github.com/scaleway/ultraviolet/assets/8331633/28324ff3-7c31-4c61-a32e-e46860ab41a4

PetmePant (approved):

Lawndlwd (approved):

PR #3018, opened 2023-10-12
feat(experimental-popup): move popup as an experimental component

BABAK0T0:
## Type

- Enhancement

### Summarise concisely:

Move Popup component as an experimental component

#### What is expected?

Make the Popup component available

1. I moved the component from an internal usage to an experimental one


packages/ui/src/components/Popup/index.tsx
```diff
@@ -157,6 +157,9 @@ type PopupProps = {
   needDebounce?: boolean
 }
 
+/**
+ * @experimental
```
matthprost, line 161:
```suggestion
 * @experimental This component is experimental and may be subject to breaking changes in the future.
```

matthprost (changes requested):

johnrazeur (approved):

ModuloM (approved):

PR #3075, opened 2023-10-24
fix(MenuV2): focus when opening error

QuiiBz:
## Type

- Bug

#### What is expected?

`requestIdleCallback` isn't defined yet in stable Safari (https://caniuse.com/requestidlecallback), but we have to wait for the `popupRef` to be defined before focusing the content when opening a menu. This PR replaces `requestIdleCallback(cb)` by `setTimeout(cb, 1)`

### DIFF
--- a/packages/ui/src/components/MenuV2/index.tsx
+++ b/packages/ui/src/components/MenuV2/index.tsx
@@ -114,12 +114,12 @@ const FwdMenu = forwardRef(
 
       // Focus the first item when the menu is opened
       if (!isVisible) {
-        requestIdleCallback(() => {
+        setTimeout(() => {
           // We have to wait for the popup to be inserted in the DOM
           if (popupRef.current?.firstChild?.firstChild instanceof HTMLElement) {
             popupRef.current.firstChild.firstChild.focus()
           }
-        })
+        }, 1)
       }
     }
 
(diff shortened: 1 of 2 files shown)


BABAK0T0 (approved):

packages/ui/src/components/MenuV2/index.tsx
```diff
@@ -114,12 +114,12 @@ const FwdMenu = forwardRef(
...
       // Focus the first item when the menu is opened
       if (!isVisible) {
-        requestIdleCallback(() => {
+        setTimeout(() => {
           // We have to wait for the popup to be inserted in the DOM
           if (popupRef.current?.firstChild?.firstChild instanceof HTMLElement) {
             popupRef.current.firstChild.firstChild.focus()
           }
-        })
+        }, 1)
```
matthprost, line 122: Why providing 1? Can't we leave it empty (so it will be 0 by default) and so it will be executed asap

matthprost (changes requested):

packages/ui/src/components/MenuV2/index.tsx
QuiiBz: It doesn't work with a timeout of `0`, not exactly sure why... Maybe due to how React schedules the update

matthprost (approved):

PR #3081, opened 2023-10-25
fix(Popover): set a default color in content to avoid inherit from parent

matthprost:
## Type

- Bug

#### What is expected?

The color inside the popover is inherited from parents which makes the next white. To avoid this I forced the color to be text neutral.

| url  | Screenshot 2023-10-25 at 14 10 38 | Screenshot 2023-10-25 at 14 10 51 |

### DIFF
--- a/.changeset/strong-pumpkins-buy.md
+++ b/.changeset/strong-pumpkins-buy.md
@@ -1,0 +1,5 @@
+---
+'@ultraviolet/ui': patch
+---
+
+Fix popover color inherited from parent
--- a/packages/ui/src/components/Popover/index.tsx
+++ b/packages/ui/src/components/Popover/index.tsx
@@ -55,6 +55,11 @@ const StyledPopup = styled(Popup, {
   }}
 `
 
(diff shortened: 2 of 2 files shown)


packages/ui/src/components/Popover/__stories__/Template.stories.tsx
```diff
@@ -7,7 +7,7 @@ export const Template: StoryFn<typeof Popover> = props => (
     visible
     {...props}
     title="Popover Title"
-    content="This is a simple text content inside the popover. You can customize it by passing text into content property."
```
fabienhebert, line 10: optional mb previous story was a more recommended usage especially for template story ? We could go back to previous story and then decide if a react node as content could have a dedicated story

fabienhebert (approved):

packages/ui/src/components/Popover/__stories__/Template.stories.tsx
matthprost: @fabienhebert my bad I forgot to remove the test

philibea (approved):

PR #3085, opened 2023-10-26
feat(stepper): update component stepper with prop size and labelposition

JulienSaguez:
## Type

- Feature

#### What is expected?

update component stepper with prop size and labelposition

For the reference, check the figma https://www.figma.com/file/XHoO4V9LOFEFkAZsocYJ83/%5BCore%5D-Core?type=design&node-id=9924%3A119162&mode=design&t=KFPgzkAXzhRvPvYS-1 and the ticket https://jira.infra.online.net/browse/DS-587


packages/ui/src/components/Stepper/index.tsx
```diff
@@ -8,6 +8,15 @@ import { Text } from '../Text'
 
 type Temporal = 'previous' | 'next' | 'current'
 
+const SIZES = {
+  small: {
+    height: 2,
```
DorianMaliszewski, line 13: We should not have this hardcoded but comes from theme no ?

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -8,6 +8,15 @@ import { Text } from '../Text'
 
 type Temporal = 'previous' | 'next' | 'current'
 
+const SIZES = {
```
DorianMaliszewski, line 11:
```suggestion
const LINE_HEIGHT_SIZES = {
```

better to understand the usage ?

matthprost:
@JulienSaguez can you rebase your branch it won't trigger the deploy action until it is done, thanks 🙏

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -111,6 +132,8 @@ type StepperProps = {
   selected?: number
   children: ReactNode[]
   className?: string
+  labelPosition?: 'column' | 'row'
+  size?: 'small' | 'medium'
```
ModuloM, line 136: I think you'll need two tests for this.

ModuloM (approved):

packages/ui/src/components/Stepper/index.tsx
JulienSaguez: Yes i can add test for the two props

PetmePant (changes requested):
1) Need to change the prop options for labelPosition to bottom and right

2) Need to fix a small bug on the bullet number on small with right orientation, the text is not aligned center

Screenshot 2023-11-02 at 1 01 41 PM
Screenshot 2023-11-02 at 1 01 37 PM

PR #3086, opened 2023-10-27
feat(node): update node version ci

philibea:
## Type

- Enhancement

#### What is expected?

Use node lts for our CI


ModuloM (approved):

.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        node: ['18']
+        node: ['18', '20']
```
matthprost, line 43: Why we wanna keep both? Want to check if it build correctly on both version? Is it worth it cause it's gonna make a lot of CI actions

matthprost (changes requested):

.github/workflows/ci.yml
philibea: As we keep node version to 18, i think it's usefull to add some test on node20 actually but if we want to drop node18 we can remove both version. I think it's too soon right now to remove node18 maybe ?
```
"engines": {
    "node": ">=18.x",
    "pnpm": ">=8.x"
  },
```

matthprost: @philibea I think if we move forward to a new version we should do it everywhere. Logically node 20 is backward compatible with node 18. 

Do you have an idea in mind to keep node 18?

matthprost (approved):

PR #3141, opened 2023-11-13
fix(Popup): attach to children and new prop portal target

matthprost:
## Type

- Enhancement

#### What is expected?

- Improved Popup (and so Popover, MenuV2 and Tooltip) to be attached to children instead of `document.body`
- Added new prop `portalTarget` on Popup component to allow attaching to a specific element

Overall the components are more accessible and the animation works better.


BABAK0T0 (approved):

ModuloM (approved):

packages/ui/src/components/Popup/index.tsx
```diff
@@ -422,6 +422,49 @@ export const Popup = forwardRef(
...
 
+    /**
+     * This event will occur only for dialog and will trap focus inside the dialog.
+     */
+    const handleFocusTrap: KeyboardEventHandler = useCallback(event => {
+      event.stopPropagation()
+      const isTabPressed = event.key === 'Tab'
+
+      if (!isTabPressed) {
+        return
```
ModuloM, line 409: Maybe stop propagation after testing if tab is pressed, so the event can properly bubble?

packages/ui/src/constants/zIndex.ts
```diff
@@ -0,0 +1,8 @@
+// Those values are used in each corresponding component
+// Those component needs z-index as the portal is attached to the children container to keep accessibility working properly.
+// Only tooltip is attached to document.body and do not need z-index
+export const Z_INDEX = {
+  menu: 1,
+  popover: 1,
+  modal: 1,
```
ModuloM: As Modal and Popover (and Menu) are in the same z-index, wont' it be problematic if you have, say, a popover into a modal?

matthprost: @ModuloM They are made the way that the portal created in order in the children. They all are same zindex so first in dom is shown first. You want to try you can check in Modal story and Popover story there is examples at the bottom

packages/ui/src/components/Popup/index.tsx
matthprost: @ModuloM good point I took this from Modal made by dorian without re-thinking the code, I will change it

PR #3172, opened 2023-11-20
fix(popup): to handle table and sub table elements when choosing parent

matthprost:
## Type

- Bug

#### What is expected?

There is an issue when the popup is used inside a `Table` due to the positioning of the table. To fix this issue I updated the way we get the parent to calculate positioning when the portal is attached to the children. I made a function `findOffsetParent` that will check if the parents is a table or a sub element and will search until it reaches `document.body`. If nothing is found `document.body` will be the one the portal will attach to.


BABAK0T0 (approved):

vincentaudebert (approved):

packages/ui/src/components/Popup/helpers.ts
```diff
@@ -73,6 +73,34 @@ const computePlacement = ({
...
 }
 
+/**
+ * This function will check if the offset parent is usable for popup positioning
+ * If not it will loop and search for a compatible parent until document.body is reached
+ */
+const findOffsetParent = (element: RefObject<HTMLDivElement>) => {
+  const offsetParent = element?.current?.offsetParent
+
+  if (offsetParent && !['TH', 'TD', 'TABLE'].includes(offsetParent.tagName)) {
```
philibea, line 83: Maybe add a commentaire why we are looking for this specific element ?

packages/ui/src/components/Popup/helpers.ts
```diff
@@ -73,6 +73,34 @@ const computePlacement = ({
...
+    return offsetParent
+  }
+
+  let currentElement = element?.current
+
+  while (currentElement && currentElement.tagName !== 'BODY') {
+    const { position } = window.getComputedStyle(currentElement)
```
philibea, line 90: check if window exist ? it's will not work server side ?

matthprost: @philibea should exists as this function is only executed in useEffect, the build of examples seems fine so I think it's okay. Wdyt?

PR #3183, opened 2023-11-22
fix(Tag): to display tooltip when tag content is too long

matthprost:
## Type

- Enhancement

#### What is expected?

When the tag is too long we are hiding the text but we never show it with a tooltip.

Before:
Screenshot 2023-11-22 at 16 01 31

After:
Screenshot 2023-11-22 at 16 01 35

### DIFF
--- a/packages/ui/src/components/TagList/__tests__/__snapshots__/index.test.tsx.snap
+++ b/packages/ui/src/components/TagList/__tests__/__snapshots__/index.test.tsx.snap
@@ -48,14 +48,19 @@ exports[`TagList renders correctly 1`] = `
   border: solid 1px #d9dadd;
 }
 
-.cache-t6vuyb-StyledTag {
+.cache-1l2yxbb-StyledText-StyledText {
   font-size: 12px;
+  font-family: Inter,Asap;
   font-weight: 500;
-  color: inherit;
-  max-width: 232px;
-  overflow: hidden;
+  letter-spacing: 0;
+  line-height: 16px;
+  text-transform: none;
+  -webkit-text-decoration: none;
+  text-decoration: none;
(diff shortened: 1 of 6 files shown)


JulienSaguez (approved):

Lawndlwd (approved):

packages/ui/src/components/TagList/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -94,16 +99,15 @@ exports[`TagList renders correctly 1`] = `
           class="cache-1anauza-StyledContainer el6733p2"
         >
           <span
-            aria-disabled="false"
```
philibea, line 97: i think Text component doesn't forward aria-disabled props :)

philibea: https://github.com/scaleway/ultraviolet/pull/3183/files#diff-f307eebe3b932f9e681fd1ceb69316f0bfb8b7929514a977fd7adcd1aba1beb4R108

BABAK0T0 (approved):

packages/ui/src/components/TagList/__tests__/__snapshots__/index.test.tsx.snap
matthprost: @philibea that's right but it was a mistake to put it on text only. Instead I will add it to the container so it will explicitly say that the container and it's children are disabled.

matthprost: Screenshot 2023-11-23 at 10 54 14

PR #3189, opened 2023-11-23
feat(verificationcode): add size prop and fix disabled

JulienSaguez:
## Type
- Feature

#### What is expected?

Add size prop and update state for disable


packages/ui/src/components/VerificationCode/index.tsx
```diff
@@ -45,6 +65,14 @@ const StyledInput = styled.input`
   &::placeholder {
     color: ${({ theme }) => theme.colors.neutral.textWeak};
   }
```
matthprost, line 67: Proto was not clear about placeholder so I asked for clarification. Here you need to add a condition in css to check if the input is disabled then the color of placeholder should be `neutral.textWeakDisabled`

matthprost (changes requested):

packages/ui/src/components/VerificationCode/__stories__/Size.stories.tsx
```diff
@@ -0,0 +1,12 @@
...
+export const Size = Template.bind({})
+Size.args = {
+  size: 'small',
+}
+
+Size.parameters = {
+  docs: {
+    description: { story: 'use `size` prop to change the size of the input' },
+  },
+}
```
matthprost, line 12: Can we get all the sizes in the storybook ? 

Screenshot 2023-11-28 at 10 44 05
 
That's my last feedback im approving in advance!

matthprost (approved):

packages/ui/src/components/VerificationCode/__stories__/Size.stories.tsx
JulienSaguez: Yes i add all sizes

BABAK0T0 (approved):

PR #3199, opened 2023-11-28
feat: new component TextArea

fabienhebert:
## Type

- Feature

### Summarise concisely:

Add a new `TextArea` component


packages/form/src/components/TextAreaField/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+        }}
+      >
+        {ChildStory()}
+      </Form>
+    ),
+  ],
+  parameters: {
+    docs: {
+      description: {
+        component: 'A tags field',
```
matthprost, line 24: Wrong description here, by the way we don't need description here anymore. You can add the description of the component with JSDoc above the component declaration in your code.

packages/form/src/components/TextAreaField/__stories__/Template.stories.tsx
```diff
@@ -0,0 +1,18 @@
+import type { StoryFn } from '@storybook/react'
+import type { TextAreaFieldProps } from '..'
+import { TextAreaField } from '..'
+import { Submit, TextInputField } from '../..'
+
+export const Template: StoryFn<TextAreaFieldProps> = args => (
+  <div>
+    <TextAreaField {...args} />
+    <TextInputField name="osef" required regex={[/[a-z]+/]} />
```
matthprost, line 9:
```suggestion
    <TextInputField name="example" required regex={[/[a-z]+/]} />
```

packages/form/src/components/TextAreaField/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+describe('TextAreaField', () => {
+  test('should render correctly', () =>
+    shouldMatchEmotionSnapshotFormWrapper(
+      <TextAreaField label="Test" name="test" />,
+    ))
+
+  test('should render correctly generated', async () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const onSubmit = jest.fn<any>()
```
matthprost, line 21: Doesn't it work like this ?

```suggestion
    const onSubmit = jest.fn(() => {})
```

So we can remove the eslint disable

packages/form/src/components/TextAreaField/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onFocus,
+  onBlur,
+  placeholder,
+  readOnly,
+  required,
+  rows,
+  success,
+  tooltip,
+  validate,
+  regex,
```
matthprost, line 40: `className` is missing among the props

matthprost: also it can be useful to integrate `tabIndex`

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -0,0 +1,288 @@
...
+  minLength?: number
+  maxLength?: number
+  tooltip?: string
+  required?: boolean
+  'data-testid'?: string
+  name?: string
+  onFocus?: DOMAttributes<HTMLTextAreaElement>['onFocus']
+  onBlur?: DOMAttributes<HTMLTextAreaElement>['onBlur']
+  clearable?: boolean
+  labelDescription?: ReactNode
```
matthprost, line 135: Here too `className` and `tabIndex` are missing

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -0,0 +1,288 @@
...
+}
+
+/**
+ * Descriptionnnnnn
```
matthprost, line 139: Needs a real description 😄

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+  ${({ theme, isSuccess, isError, readOnly, disabled }) => {
+    if (disabled) {
+      return `
+            background : ${theme.colors.neutral.backgroundDisabled};
+            border-color : ${theme.colors.neutral.borderDisabled};
+            color : ${theme.colors.neutral.textDisabled};
+            &::placeholder {
+              color: ${theme.colors.neutral.textWeakDisabled};
+            }
```
matthprost, line 71: `cursor: not-allowed` is missing here

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+    return ''
+  }}
+`
+
+type TextAreaProps = {
+  id?: string
+  autoFocus?: boolean
+  label: string
+  value: string
```
matthprost, line 99: Why value is required? If it is required the placeholder has no utility

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -0,0 +1,288 @@
...
+              <Text
+                as="div"
+                sentiment="neutral"
+                prominence="weak"
+                variant="caption"
+              >
+                {value.length}/{maxLength}
+              </Text>
+            ) : null}
+          </Row>
```
matthprost, line 283: You need to condition this to be rendered or not, or we will have empty divs and spacing when not bottom text is displayed

Screenshot 2023-12-05 at 14 37 44

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -0,0 +1,288 @@
...
+              sentiment="neutral"
+              htmlFor={id ?? localId}
+            >
+              {label}
+            </Text>
+            {required ? <Icon name="asterisk" color="danger" size={8} /> : null}
+          </Stack>
+          {labelDescription ?? null}
+        </Stack>
+        <div>
```
matthprost, line 193: Does this div have a purpose? If not can we make it an empty snippet?

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -0,0 +1,288 @@
...
+                autoFocus={autoFocus}
+                disabled={disabled}
+                rows={rows}
+                ref={ref}
+                value={value}
+                onChange={event => {
+                  onChange(event.currentTarget.value)
+                }}
+                isSuccess={!!computedSuccess}
+                isError={!!computedError}
```
matthprost, line 207: I just read that `aria-invalid` can be a good addition but not required, so maybe we should add it too

> When visible text is used to programmatically identify a failed field and / or convey how the error can be corrected, setting aria-invalid to "true" is not required from a strict compliance standpoint but may still provide helpful information for users.

Source: https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA21.html#:~:text=When%20visible%20text%20is%20used,provide%20helpful%20information%20for%20users.

matthprost (changes requested):

PetmePant (changes requested):
1) The clear button should be neutral and not primary
Screenshot 2023-12-06 at 10 35 06 AM When you resize is it normal to look like the screenshot or we should make it minimum 1 row?

Screenshot 2023-12-06 at 10 36 08 AM

packages/ui/src/components/TextArea/index.tsx
fabienhebert: placeholder is displayed if value is an empty string, however form might provide undefined indeed

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -0,0 +1,297 @@
...
+              onBlur={onBlur}
+              readOnly={computedReadOnly}
+            />
+            <StyledTextAreaAbsoluteStack
+              direction="row"
+              alignItems="center"
+              gap="1"
+            >
+              {clearable ? (
+                <Button
```
matthprost, line 231: A bit miss aligned 
Screenshot 2023-12-11 at 10 17 37

matthprost (changes requested):

packages/ui/src/components/TextArea/index.tsx
mhervouet: Yes, should be 12px and not 16

vincentaudebert (approved):

PetmePant (approved):

matthprost (approved):

PR #3244, opened 2023-12-15
feat(NumberInputV2): new component

matthprost:
## Type

- Feature

#### What is expected?

New component NumberInputV2, an improved version of the NumberInput with similar design.

Screenshot 2024-01-03 at 16 29 16


packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,373 @@
...
+                />
+              </SideContainer>
+            </Container>
+          </Tooltip>
+        </div>
+        {error || success || helper ? (
+          <Text
+            variant="caption"
+            as="span"
+            prominence="weak"
```
fabienhebert, line 364: weak is only for "normal"/helper case

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,373 @@
...
+  className?: string
+  'data-testid'?: string
+  label?: string
+  /**
+   * Label description displayed right next to the label. It allows you to customize the label content.
+   */
+  labelDescription?: ReactNode
+  error?: string
+  success?: string
+  helper?: string
```
fabienhebert, line 183: could be a ReactNode

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,373 @@
...
+  useImperativeHandle,
+  useMemo,
+} from 'react'
+import { Button } from '../Button'
+import { Stack } from '../Stack'
+import { Text } from '../Text'
+import { Tooltip } from '../Tooltip'
+
+const SIZES = {
+  small: '34px',
```
fabienhebert, line 18: not 32px ?

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,373 @@
...
+          {labelDescription ?? null}
+        </Stack>
+        <div>
+          <Tooltip text={tooltip}>
+            <Container
+              data-disabled={disabled}
+              data-readOnly={readOnly}
+              data-error={!!error}
+              data-success={!!success}
+              className={className}
```
fabienhebert, line 296: It may should be applied on the main container (Stack l274), otherwhile there is no way to apply width/position  or placement properties like that on the all component ?

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,373 @@
...
+  large: '48px',
+}
+
+type Sizes = keyof typeof SIZES
+
+const SideContainer = styled('div', {
+  shouldForwardProp: prop => !['size'].includes(prop),
+})<{ size: Sizes }>`
+  &[data-position='left'] {
+    border-right: 1px solid ${({ theme }) => theme.colors.neutral.border};
```
fabienhebert, line 29: May be you could do, this way you are sure SideContainer follow parent border-color including hover/focus cases
```css
   border-right: 1px solid;
   border-color: inherit;
```

PS : I didn't find a way to one line this `border-right: 1px solid "inherit";`

.changeset/itchy-comics-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ultraviolet/ui': minor
```
fabienhebert, line 2: reminder: if you add later a NumberInputFieldV2, you'll probably need to generate another changeset

packages/ui/src/components/NumberInputV2/index.tsx
matthprost: @fabienhebert good idea im gonna do this

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,424 @@
...
+} from 'react'
+import { Button } from '../Button'
+import { Stack } from '../Stack'
+import { Text } from '../Text'
+import { Tooltip } from '../Tooltip'
+
+const SIZES = {
+  small: '30px',
+  medium: '38px',
+  large: '46px',
```
fabienhebert, line 25: I didn't check so it's may be fine but if they are aligned with the new TextInput will it look ok ? height won't be different ?

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,424 @@
...
+  border-color: inherit;
+  background: inherit;
+`
+
+const Unit = styled(Text, {
+  shouldForwardProp: prop => !['size'].includes(prop),
+})<{ size: Sizes }>`
+  display: flex;
+  align-items: center;
+  padding: ${({ theme }) =>
```
fabienhebert, line 59: `padding: ${({ theme }) => theme.space['1]}` should have the same result since for left  (missing value) it will take  the first value (like a loop)

packages/ui/src/components/NumberInputV2/index.tsx
fabienhebert: may be more like `prominence={!error && !success ? 'weak' : 'default'}`

fabienhebert (approved):

fabienhebert:
Most of custom unit/suffix has been solved so it's probably fine, just had this case but not sure it's a relevent one

Capture d’écran 2024-01-03 à 18 48 45

fabienhebert (approved):

packages/ui/src/components/NumberInputV2/index.tsx
matthprost: @fabienhebert the height is aligned, there is only 2 px difference compared to design specs due to border

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,424 @@
...
+      `${theme.space['1']} 0 ${theme.space['1']} ${theme.space['1']}`};
+  }
+
+  // Remove native arrows from input[type=number]
+
+  &::-webkit-outer-spin-button,
+  &::-webkit-inner-spin-button {
+    -webkit-appearance: none;
+    margin: 0;
+  }
```
DorianMaliszewski, line 91: Is this enough for all browsers ?

matthprost: Under you also have for firefox 

```css
& {
    -moz-appearance: textfield;
  }
```

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,424 @@
...
+      if (error) {
+        return 'danger'
+      }
+
+      if (success) {
+        return 'success'
+      }
+
+      return 'neutral'
+    }, [error, success])
```
DorianMaliszewski, line 285:
```suggestion
    const helperSentiment: Sentiment = useMemo(() => {
      if (error) {
        return 'danger'
      }

      if (success) {
        return 'success'
      }

      return 'neutral'
    }, [error, success])
```

?

DorianMaliszewski: May be we can add some types

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -0,0 +1,424 @@
...
+                    ) {
+                      onChange?.(createChangeEvent(min.toString()))
+                    }
+
+                    if (
+                      Number.isNaN(numericValue) ||
+                      (!Number.isNaN(maxValue) && numericValue > maxValue)
+                    ) {
+                      onChange?.(createChangeEvent(max.toString()))
+                    }
```
DorianMaliszewski, line 361: how do you handle min value but not required ?

packages/ui/src/components/NumberInputV2/index.tsx
matthprost: @DorianMaliszewski Im not sure what you want to do, the typing is quite good already and there is no sentiments props on that component. Only success, error and helper as direct prop.

Screenshot 2024-01-08 at 14 10 23

mhervouet:
When the value is 0. Minus button should be disabled.

philibea (approved):

packages/form/src/components/NumberInputFieldV2/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+import type { FieldPath, FieldValues, Path, PathValue } from 'react-hook-form'
+import { useErrors } from '../../providers'
+import type { BaseFieldProps } from '../../types'
+
+type NumberInputV2Props<
+  TFieldValues extends FieldValues,
+  TName extends FieldPath<TFieldValues>,
+> = BaseFieldProps<TFieldValues, TName> &
+  Partial<
+    Pick<
```
DorianMaliszewski, line 13: Not better to use omit instead of that big list ?

DorianMaliszewski (approved):

mhervouet (approved):

packages/form/src/components/NumberInputFieldV2/index.tsx
matthprost: @DorianMaliszewski Pick was more explicit for this case as we add BaseFieldProps and then we pick from the component.

I will ad name and required are part of BaseFieldProps already but I will add className

PetmePant (approved):

PR #3256, opened 2023-12-19
refactor(Icons): change file structure to optimize svgs and reduce bundle size

matthprost:
## Type

- Refactor

#### What is expected?

Since now we use to have icons inlined into a `.ts` file which was painful to maintain and impossible to optimize.

Today we have each `.svg` file who will be in `assets` folder of `CategoryIcon` and `ProductIcon`. Those individual files can be optimized thanks to svgo and so reduce the bundle size. This is also a good way to cleanup the `.svg` files coming from design and optimizing it while pushing on github thanks to husky.

To achieve this I needed to create special config for `svgo` as we need to keep `className` on svgs as they are useful for setting a color and allows those component to be theme compliant.


vincentaudebert (approved):

vincentaudebert:
CI to be fixed 👌

fabienhebert (approved):

PR #3262, opened 2023-12-20
fix(TextInput): input padding to follow unit length

matthprost:
## Type

- Bug

#### What is expected?

Current `<TextInput />` has an issue when we have too long unit lenght. The padding inside the input follows the size of the elements on the right it includes the unit length. So far we were just checking how many element there was on the right without counting the number of character in unit creating overflow issue as you can see on screenshots. 

It has been solve to take in account the unit size. It also fixe `<EstimateCost.Unit />` and I increased slightly the size of the unit.

| TextInput  | Screenshot 2023-12-20 at 11 15 24 | Screenshot 2023-12-20 at 11 15 17 |
| EstimateCost.Unit  | Screenshot 2023-12-20 at 11 17 02 | Screenshot 2023-12-20 at 11 17 10 |

### DIFF
--- a/.changeset/chilly-cobras-bathe.md
+++ b/.changeset/chilly-cobras-bathe.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/form": patch
+---
+
+Fix `<TextInputField />` to change input padding according to unit size avoiding overflow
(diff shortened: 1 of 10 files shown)


Lawndlwd (approved):
💜

JulienSaguez (approved):

PR #3265, opened 2023-12-20
feat: add TextInputV2 and TextInputFieldV2

fabienhebert:
## Type

- Feature

### Summarise concisely:

Add a new version of `TextInput` and `TextInputFieldV2`


packages/form/src/components/TextInputFieldV2/__stories__/Template.stories.tsx
```diff
@@ -0,0 +1,19 @@
+import type { StoryFn } from '@storybook/react'
+import { Stack } from '@ultraviolet/ui'
+import type { ComponentProps } from 'react'
+import { TextInputField } from '..'
+import { Submit } from '../..'
+
+export const Template: StoryFn<
+  ComponentProps<typeof TextInputField>
+> = args => (
+  <Stack gap="1">
```
vincentaudebert, line 10:
```suggestion
  <Stack gap={1}>
```

?

packages/ui/src/components/TextInputV2/__stories__/Examples.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { TextInput } from '..'
+import { Link } from '../../Link'
+import { Stack } from '../../Stack'
+import { Text } from '../../Text'
+
+export const Examples: StoryFn<typeof TextInput> = args => {
+  const [value, setValue] = useState<string>('Text')
+
+  return (
+    <Stack gap="2">
```
vincentaudebert, line 12:
```suggestion
    <Stack gap={2}>
```

?

vincentaudebert:
Screenshot 2024-01-08 at 10 35 09

it seems magic wand doesn't work in here?

matthprost:
> Screenshot 2024-01-08 at 10 35 09
> it seems magic wand doesn't work in here?

This is because magic wand needs a function to be working using prop `onRandomize`, I will add it into the story so it actually works

packages/form/src/components/TextInputFieldV2/__stories__/Template.stories.tsx
Loriick: @vincentaudebert Look like gap can be a `number` or a key of space in the `lightTheme`

Loriick (approved):

packages/ui/src/components/TextInputV2/index.tsx
```diff
@@ -0,0 +1,374 @@
...
+  helper?: ReactNode
+  iconName?: ComponentProps<typeof Icon>['name']
+  id?: string
+  label: string
+  labelDescription?: ReactNode
+  loading?: boolean
+  minLength?: number
+  maxLength?: number
+  name?: string
+  onBlur?: DOMAttributes<HTMLInputElement>['onBlur']
```
DorianMaliszewski, line 134:
```suggestion
  onBlur?: DOMAttributes<HTMLTextAreaElement>['onBlur']
```

?

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -22,13 +22,12 @@ const StyledTextAreaAbsoluteStack = styled(Stack)`
...
 type StyledTextAreaProps = {
-  isSuccess: boolean
-  isError: boolean
+  hasSentimentIcon: boolean
   isClearable: boolean
 }
 const StyledTextArea = styled('textarea', {
   shouldForwardProp: prop =>
-    !['isSuccess', 'isError', 'isClearable'].includes(prop),
+    !['hasSentimentIcon', 'isClearable'].includes(prop),
```
DorianMaliszewski, line 30: May be use data attirbutes to don't recompute style ?

DorianMaliszewski (approved):

packages/form/src/components/TextInputFieldV2/__stories__/Template.stories.tsx
matthprost: @vincentaudebert we don't have restrictions onto that both are being used

packages/ui/src/components/TextInputV2/index.tsx
matthprost: @DorianMaliszewski text input is indeed an HTML input element, we have a TextArea component which is different. TextInput doesn't support text area 😄

mhervouet:
On read-only state, better to not be able to focus on the field (same for hover) 
Capture d’écran 2024-01-10 à 15 03 56

matthprost:
> On read-only state, better to not be able to focus on the field (same for hover) Capture d’écran 2024-01-10 à 15 03 56

The native behavior allows focus on readonly state but not hovering, do we want to sync onto that or remove any effect as you said? Cf: https://developer.mozilla.org/fr/docs/Web/HTML/Attributes/readonly#r%C3%A9sultat

mhervouet:
> > On read-only state, better to not be able to focus on the field (same for hover) Capture d’écran 2024-01-10 à 15 03 56
> 
> The native behavior allows focus on readonly state but not hovering, do we want to sync onto that or remove any effect as you said? Cf: https://developer.mozilla.org/fr/docs/Web/HTML/Attributes/readonly#r%C3%A9sultat

Ok to keep this behavior then it will be easier to identify read-only than disabled. Everything is good then :)

mhervouet (approved):

PR #3270, opened 2023-12-21
feat(plus): add customer satisfaction components

philibea:
## Type

- Feature


packages/plus/src/components/CustomerSatisfaction/index.tsx
```diff
@@ -0,0 +1,159 @@
...
+        )
+
+        return (
+          <Image
+            key={rating.value}
+            isHappy={isHappy}
+            isHovered={isHovered}
+            isSelected={isSelected}
+            onClick={() => {
+              onChange(rating.value)
```
vincentaudebert, line 155: il semblerait qu'on perd un peu de code coverage ici?

matthprost: @vincentaudebert I just updated the tests 😄

packages/plus/src/components/CustomerSatisfaction/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+    imgNotSelected: rating4NS,
+  },
+  {
+    value: 5,
+    imgSelected: rating5,
+    imgNotSelected: rating5NS,
+  },
+] as const
+
+type Value = (typeof ratings)[number]['value']
```
DorianMaliszewski, line 44: use number because in DX we will always need to cast number to Value type. We should handle number <0 and >5 and transform them to min and max ?

vincentaudebert (approved):

packages/plus/src/components/CustomerSatisfaction/index.tsx
matthprost: @DorianMaliszewski rating is hard coded number of values, if I type it as number it means developer will be able to put a value higher or lower than the number of rating. What is the purpose of doing so? Do you have an example?

PR #3315, opened 2024-01-10
refactor: fix svgs in all repo with svgr webpack for storybook and rollup for bundle

matthprost:
## Type

- Refactor

#### What is expected?

Using svgr was making all svg as react component while it is a miss configuration from our part. I fixed it the way that those two works fine in storybook:

```ts
import React from 'react';

// default export should return the URL to the asset
import svgUrl from './img/test.svg';

// named export should return a react component
import { ReactComponent as Svg } from './img/test.svg';

const Test: React.FC = () => (
  <>
    {/* use default import as URL */}
    <img src={svgUrl} />

    {/* use named import as react component */
    <Svg />
  </>
);

```


matthprost:
We can ignore the coverage error as the old snapshots were uselessly empty (you can see it in changes)

Lawndlwd (approved):

vincentaudebert (approved):

Loriick (approved):

.storybook/main.ts
```diff
@@ -19,13 +19,33 @@ export default {
   docs: {
     autodocs: true,
   },
+  core: {
+    builder: '@storybook/builder-webpack5',
+  },
   webpackFinal: async config => {
-    config?.module?.rules?.push({
+    const imageRule = config.module?.rules?.find(rule => {
+      const test = (rule as { test: RegExp }).test
```
alexandre-combemorel, line 27: if you can add a small comment about the reason why excluding the svg from imageLoader, not blocking, nice to have

alexandre-combemorel (approved):

.storybook/main.ts
matthprost: @alexandre-combemorel I will add it here: https://github.com/scaleway/ultraviolet/pull/3309

To answer here: we want to remove any default configuration that storybook has with webpack and apply our own rule with svgr

PR #3318, opened 2024-01-11
feat(banner): add disabled button prop

johnrazeur:
## Type

- Feature

Add disabled button prop to `Banner`.


Loriick (approved):

matthprost (changes requested):
As seen with design we will avoid using disabled on banner / alert as the goal of those component is to have an available action or not to be dispayed. We will hide banner or alert if the button action is not available. You can close this PR 😄

philibea (approved):

PR #3409, opened 2024-02-05
fix(CheckboxField): remove array value

johnrazeur:
## Type

- Bug

- Remove array value (use `CheckboxGroupField` for that).
- Add id prop.


.changeset/ninety-ghosts-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/form": patch
+---
+
+CheckboxField now support only key/value
```
matthprost, line 5:
```suggestion
`CheckboxField` now support only key/value
```

matthprost (changes requested):

packages/form/src/components/CheckboxField/index.tsx
```diff
@@ -45,7 +47,6 @@ export const CheckboxField = <
   helper,
   tooltip,
   'data-testid': dataTestId,
-  value,
```
matthprost, line 48: You can't remove a prop like that or it will be breaking change. Instead you can mark it as deprecated using JSDoc

matthprost (changes requested):

vincentaudebert:
Apart from matthias comment, all good to me. If it gets deprecated, what will it do though?

matthprost (approved):

vincentaudebert (approved):

fabienhebert (approved):

fabienhebert:
> Apart from matthias comment, all good to me. If it gets deprecated, what will it do though?

You can still use initialValues then `setValue`, if you need to change value in code

PR #3415, opened 2024-02-05
fix(TextInputV2): remove some props and add some others

matthprost:
## Type

- Enhancement
- Documentation

#### What is expected?

- Fix `TextInputV2` to implement more props and remove `iconName`
- Add documentation to migrate `TextInput` to `TextInputV2`

### DIFF
--- a/packages/ui/src/components/TextInputV2/index.tsx
+++ b/packages/ui/src/components/TextInputV2/index.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import { Icon } from '@ultraviolet/icons'
-import type { ComponentProps, InputHTMLAttributes, ReactNode } from 'react'
+import type { InputHTMLAttributes, ReactNode } from 'react'
 import { forwardRef, useId, useMemo, useState } from 'react'
 import { Button } from '../Button'
 import { Loader } from '../Loader'
@@ -123,7 +123,6 @@ type TextInputProps = {
   'data-testid'?: string
   error?: string
   helper?: ReactNode
-  iconName?: ComponentProps<typeof Icon>['name']
   label?: string
   labelDescription?: ReactNode
   loading?: boolean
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/TextInputV2/index.tsx
```diff
@@ -133,9 +132,8 @@ type TextInputProps = {
   onChange?: (newValue: string) => void
   prefix?: ReactNode
   size?: TextInputSize
-  success?: string
+  success?: string | boolean
```
fabienhebert, line 135: new spec ? not checking the type could lead to not wanted TextComponent as helper

matthprost: Yep new specs, I will add a verification for the text part

Lawndlwd (approved):

BABAK0T0 (approved):

JulienSaguez (approved):

packages/ui/src/components/TextInputV2/index.tsx
```diff
@@ -133,9 +132,8 @@ type TextInputProps = {
...
   suffix?: ReactNode
-  tabIndex?: number
```
DorianMaliszewski, line 138: We should keep that its more html compliant ?

packages/ui/src/components/TextInputV2/index.tsx
```diff
@@ -369,7 +371,7 @@ export const TextInputV2 = forwardRef<HTMLInputElement, TextInputProps>(
             </StyledInputWrapper>
           </Tooltip>
         </div>
-        {error || success || typeof helper === 'string' ? (
+        {error || typeof success === 'string' || typeof helper === 'string' ? (
```
DorianMaliszewski, line 374: Why not use a prop state = {variant: "", text: ""} ? Like this we can handle success, warning, error ? :thinking:

DorianMaliszewski: Here it can lead to weird behaviour if you pass error and success

matthprost: It won't be consistent this way as depending on which one you choose the text might be facultative or mandatory, also there will be no use case where the error and success will be passed. But let's imagine you do it we defined a specific order: loading > error > success > neutral

PR #3425, opened 2024-02-07
feat: update tag input component

fabienhebert:
## Type

- Enhancement

### Summarise concisely:

Update `TagInput` & `TagInputField` to follow new Ultraviolet specifications

### DIFF
--- a/packages/ui/src/components/TagInput/index.tsx
+++ b/packages/ui/src/components/TagInput/index.tsx
@@ -131,7 +131,11 @@ type TagInputProps = {
   // eslint-disable-next-line react/no-unused-prop-types
   onChangeError?: (error: Error | string) => void
   placeholder?: string
+  /**
+   * @deprecated use `value` property instead, both properties work the same way
+   */
   tags?: TagInputProp
+  value?: TagInputProp
   /**
    * @deprecated there is only one variant now, this prop has no more effect
    */
@@ -156,7 +160,6 @@ type TagInputProps = {
 
 /**
  * TagInput is a component that allows users to input tags.
- * @experimental This component is experimental and may be subject to breaking changes in the future.
  */
 export const TagInput = ({
   disabled = false,
@@ -165,6 +168,7 @@ export const TagInput = ({
   onChange,
   placeholder,
   tags,
+  value,
   className,
   'data-testid': dataTestId,
   label,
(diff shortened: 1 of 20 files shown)


mhervouet (approved):

packages/ui/src/components/TagInput/index.tsx
```diff
@@ -220,48 +276,139 @@ export const TagInput = ({
...
+            sentiment="neutral"
+            htmlFor={id ?? localId}
+          >
+            {label}
+          </Text>
+          {required ? <Icon name="asterisk" color="danger" size={8} /> : null}
+        </Stack>
+        {labelDescription ?? null}
+      </Stack>
+
```
matthprost, line 320: This is a bit fucked up but we need to encapsulate tooltip into a div so it inherit from display block in this case, I had the same issue with other components if we don't do it the width is broken when setting a tooltip

Screenshot 2024-02-07 at 18 00 43

packages/ui/src/components/TagInput/index.tsx
```diff
@@ -17,79 +37,82 @@ const STATUS = {
...
+    border-color: ${({ theme }) => theme.colors.primary.borderHover};
+  }
 
-type TagInputContainersProps = {
-  variant: Variant
-}
+  &[data-readonly='true'] {
+    border-color: ${({ theme }) => theme.colors.neutral.border};
+    background: ${({ theme }) => theme.colors.neutral.backgroundWeak};
+    cursor: not-allowed;
```
matthprost, line 79: I think we should allow cursor in ready only as you should be able to select and copy the content

packages/ui/src/components/TagInput/index.tsx
```diff
@@ -1,13 +1,33 @@
...
+import { Stack } from '../Stack'
 import { Tag } from '../Tag'
+import { Text } from '../Text'
+import { Tooltip } from '../Tooltip'
+
+// SIZE
+export const TAGINPUT_SIZE_HEIGHT = {
+  large: 48,
+  medium: 40,
+  small: 32,
```
matthprost, line 21: There is a little issue with the sizes, I think it should subtract the border so it will match other inputs

TagInput:
Screenshot 2024-02-07 at 18 06 42

TextInput:
Screenshot 2024-02-07 at 18 06 57

So large should be 46, medium 38 and small 30

matthprost (changes requested):
Also I think we can remove the experimental on the component as it tend to be stable now 😄 
Screenshot 2024-02-08 at 10 14 55

packages/ui/src/components/TagInput/index.tsx
```diff
@@ -220,48 +267,140 @@ export const TagInput = ({
...
+                        }
+                      : undefined
+                  }
+                >
+                  {tag.label}
+                </Tag>
+              ))}
+              {!disabled ? (
+                <StyledInput
+                  id={id}
```
matthprost, line 344: Here you should pass `id ?? localId` or it won't work when we click on label

```suggestion
                  id={id ?? localId}
```

matthprost (changes requested):

matthprost (approved):

BABAK0T0 (approved):

matthprost (approved):

PR #3426, opened 2024-02-08
feat(icons): add export to IconName props

Slashgear:
## Type

- Feature

### Summarise concisely:

We need to get the IconName easily for wrapping purpose, extract this props from the component props is quite complex due to forwardRef, having it directly extracted would be really useful

(Describe what you did)

1. Add missing export type in package ultraviolet/icon

### DIFF
--- a/.changeset/fresh-toes-jam.md
+++ b/.changeset/fresh-toes-jam.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/icons": minor
+---
+
+add export to IconName type
--- a/packages/icons/src/components/index.ts
+++ b/packages/icons/src/components/index.ts
@@ -1,3 +1,4 @@
+export type { IconName } from './Icon'
 export { Icon } from './Icon'
 export { CategoryIcon } from './CategoryIcon'
 export { ProductIcon } from './ProductIcon'


matthprost (changes requested):
We don't usually export type that are already available through props, why not using this instead?

```ts
ComponentProps<typeof Icon>['name']
```

You will get the exact same type as `IconName`

Edit: I've re-read your message, what's the issue with the forwardRef and componentProps?

philibea:
> > We don't usually export type that are already available through props, why not using this instead?
> > ```ts
> > ComponentProps<typeof Icon>['name']
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You will get the exact same type as `IconName`
> > Edit: I've re-read your message, what's the issue with the forwardRef and componentProps?
> 
> Sadly due to `forwardRef` usage, this does not work 😢

You can do this one in this case :)
```ts
type IconName =  ComponentPropsWithoutRef<typeof Icon>['name']
```

Slashgear:
> > > We don't usually export type that are already available through props, why not using this instead?
> > > ```ts
> > > ComponentProps<typeof Icon>['name']
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > You will get the exact same type as `IconName`
> > > Edit: I've re-read your message, what's the issue with the forwardRef and componentProps?
> > 
> > 
> > Sadly due to `forwardRef` usage, this does not work 😢
> 
> You can do this one in this case :)
> 
> ```ts
> type IconName =  ComponentPropsWithoutRef<typeof Icon>['name']
> ```

We will try it, we already did and it did not work but let's retry it

dcerbah:
I tested the suggestion and it doesn't work, here is a screen of the error :

Capture d’écran 2024-02-09 à 09 54 23

matthprost:
> I tested the suggestion and it doesn't work, here is a screen of the error :
> 
> Capture d’écran 2024-02-09 à 09 54 23

I tried locally on another project than ultraviolet, it seems to work just fine:
Screenshot 2024-02-09 at 11 45 49

Is there any ways you could provide me a way to reproduce or try this error locally? So I can help you search for a solution.

dcerbah:
Following your investigations this problem came from our yarn lock file, thanks for your help.

PR #3455, opened 2024-02-15
feat(Table): add highlightAnimation on table row

matthprost:
## Type

- Enhancement

#### What is expected?

We want to be able to add an animation on one specific row of a table to show to the user that a new data has been added into it. To do so we created a new prop `highlightAnimation` on `Table.Row` that will show an animation on the row.

https://github.com/scaleway/ultraviolet/assets/15812968/9cd26c81-c641-4097-a988-68db66471b9f


PetmePant (approved):

BlandinePelat (approved):
Hello! The animation may be a little long, but for now we can get away with it.

PR #3459, opened 2024-02-15
fix(TextInputFieldV2): fix type prop

lisalupi:
## Type

- Bug

### Summarise concisely:

The prop "type" in `TextInputField` is now working.
#### What is expected?

When choosing a type for an text input field, it can be a password (with the style of a TextInputV2 field), email, url or text. 

(Describe what you did)

1. Edit the component (add type)

### DIFF
--- a/.changeset/real-spoons-cough.md
+++ b/.changeset/real-spoons-cough.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/form": minor
+---
+
+Prop `type` working properly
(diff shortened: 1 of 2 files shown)


.changeset/real-spoons-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/form": minor
+---
+
+Prop `type` working properly
```
matthprost, line 5:
```suggestion
Fix prop `type` on component `<TextInputFieldV2 />` that wasn't working properly
```

We try to add the component name so people will understand what change and where 😄

matthprost (changes requested):

matthprost (approved):

johnrazeur (approved):

philibea (approved):

PR #3466, opened 2024-02-16
refactor: new component FAQ

lisalupi:
## Type

- Refactor

#### What is expected?

`FAQ` component added to ultraviolet/plus

1. New component 

2. Add component to storybook


.changeset/real-spoons-cough.md
```diff
@@ -1,5 +0,0 @@
----
-"@ultraviolet/form": minor
----
-
-Fix prop `type` on component `<TextInputFieldV2 />` that wasn't working properly
```
matthprost, line 5: This is weird it shouldn't be here, maybe if you rebase and push it will disappear?

matthprost (changes requested):

matthprost (changes requested):
You need to export your component in `packages/plus/src/components/index.ts`, I think I forgot to tell you for Conversation component also, so you can do both in that PR

packages/plus/src/components/FAQ/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,9 @@
+import type { Meta } from '@storybook/react'
+import { FAQ } from '..'
+
+export default {
+  component: FAQ,
+  title: 'Plus/Compositions/FAQ',
+} as Meta
```
alexandre-combemorel, line 7: prefere`satisfies` instead of `as`, like that you get typescript error when you change the type :)

alexandre-combemorel (approved):

.changeset/real-spoons-cough.md
matthprost: @lisalupi this is still an issue, did you try rebase on main?

matthprost (approved):

PR #3493, opened 2024-02-23
fix(Skeleton): add aria-busy property

matthprost:
## Type

- Enhancement

#### What is expected?

Fix `<Skeleton />` component to have `aria-busy`

### DIFF
--- a/packages/ui/src/components/Skeleton/index.tsx
+++ b/packages/ui/src/components/Skeleton/index.tsx
@@ -84,7 +84,7 @@ export const Skeleton = ({
   const Component = variants[variant]
 
   return (
-    <StyledContainer className={className} aria-busy>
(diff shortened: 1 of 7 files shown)


Lawndlwd (approved):

packages/ui/src/components/Skeleton/index.tsx
```diff
@@ -84,7 +84,7 @@ export const Skeleton = ({
   const Component = variants[variant]
 
   return (
-    <StyledContainer className={className}>
+    <StyledContainer className={className} aria-busy>
```
alexandre-combemorel, line 87: I think we need to associate it with `aria-live="polite"`

johnrazeur (approved):

BABAK0T0 (approved):

PR #3508, opened 2024-02-28
feat(Popup): popup is now capable to detect overflow and re-adjust

matthprost:
## Type

- Enhancement

### Popup Enhancement 

Our home made popup is used by quite a lot of components (`Tooltip`, `Popover`, `Menu`), it works well but it was lacking some more smart behaviours. In fact the popup was quite often going out of screen. In the best case we have a scroll bar but in the worst you can't see half of the content.

gif_before

#### Introducing: Smart Positionning ✨

A new function has been created `getPopupOverflowFromParent()`. This function has only one purpose, calculate how much the popup overflow from it's parent and so from screen. It is capable to calculate for any placements (`top`, `bottom`, `right`, `left`).

Thanks to this function we are now capable to re-adjust the position of the popup. We also re-adjust the position of the arrow on the popup to follow the dynamic position.

Here is the result:
popup_after

No more overflow, no more hidden values, popup is now smarter! 🚀


Loriick (approved):

lisalupi (approved):

vincentaudebert:
j'ai une blague! j'ai une blague!

1+1 = toujours 2! C'est Matthoumatique! 

🤣

PR #3509, opened 2024-02-28
feat: New component SteppedListCard

lisalupi:
## Type

- Feature

#### What is expected?

Created a new component "SteppedListCard" for ultraviolet/plus

1. New component, story and tests

2.

3.

Capture d’écran 2024-02-28 à 17 35 45

### DIFF
--- a/packages/plus/package.json
+++ b/packages/plus/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ultraviolet/plus",
-  "version": "0.8.3",
+  "version": "0.8.4",
   "description": "Ultraviolet Plus",
   "homepage": "https://github.com/scaleway/ultraviolet#readme",
   "repository": {
--- a/.changeset/gorgeous-insects-build.md
+++ b/.changeset/gorgeous-insects-build.md
@@ -1,6 +1,0 @@
----
-"@ultraviolet/form": minor
-"@ultraviolet/ui": minor
----
-
-`<SelectableCard />` component now implement natively indentation for childrens and spacing between label and childrens
--- a/.changeset/popular-spies-dream.md
+++ b/.changeset/popular-spies-dream.md
@@ -1,6 +1,0 @@
----
-"@ultraviolet/illustrations": minor
----
-
-- 11 new illustrations (danger, success, warning, empty, failed, generic, loading, lock, plus, resources, search)
-- New component `DynamicIllustration` to swap automatically to the appropriate version of the illustration
--- a/.changeset/renovate-bfe6362.md
+++ b/.changeset/renovate-bfe6362.md
@@ -1,7 +1,0 @@
----
-'@ultraviolet/plus': patch
----
-
-Updated dependency `@uiw/codemirror-extensions-langs` to `4.21.24`.
-Updated dependency `@uiw/codemirror-theme-material` to `4.21.24`.
-Updated dependency `@uiw/react-codemirror` to `4.21.24`.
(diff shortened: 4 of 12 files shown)


matthprost:
@lisalupi can you please rebase and fix the merge conflict or the pipeline won't run and we cannot check the deploy 😅

packages/plus/package.json
```diff
@@ -59,6 +59,7 @@
     "@uiw/react-codemirror": "4.21.22",
     "@ultraviolet/themes": "workspace:*",
     "@ultraviolet/ui": "workspace:*",
+    "@ultraviolet/illustrations": "workspace:*",
```
matthprost, line 62: You can add this in dev dependencies as you are only using it in storybook

packages/plus/src/components/SteppedListCard/SteppedListContainer.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { Data } from './helper'
+
+const StyledHeader = styled(Row)`
+  padding-bottom: ${({ theme }) => theme.space['3']};
+`
+
+const Content = styled(Row)`
+  border: solid ${({ theme }) => theme.colors.neutral.border};
+  width: 100%;
+`
```
matthprost, line 16: This can probably be replace by a Card and a Row inside of it without any styled 😄 

If you can try to avoid using styled

packages/plus/src/components/SteppedListCard/SteppedListContainer.tsx
```diff
@@ -0,0 +1,125 @@
...
+      setDone,
+      setHidden,
+    }),
+    [currentStep, setCurrentStep, numberOfSteps, done, setDone, setHidden],
+  )
+
+  return (
+    <>
+      {hidden ? (
+        showComponent(setHidden)
```
matthprost, line 81: I think we can remove this part, we can say that when the hide button is pressed only the header + hide/show button is displayed. We can add a prop `onClickHide` function that will be triggered on click on hide.

packages/plus/src/components/SteppedListCard/helper.ts
```diff
@@ -0,0 +1,52 @@
...
+  completed,
+  setCompleted,
+  stepNumber,
+  setCurrentStep,
+  numberOfSteps,
+  setHidden,
+  done,
+}: NextStepProps) => {
+  const tempDone = done
+  tempDone[stepNumber - 1] = completed
```
matthprost, line 45: There is a mutation here we try to avoid doing them as there is a risk of side effects, better do the following:

```ts
const tempDone = done.map((item, index) => index === stepNumber - 1 ? completed : item);
```

If you want to check about it here is a good article: https://dev.to/amissah17/ultimate-guide-to-mutations-in-javascript-2a80#:~:text=Mutation%20means%20a%20change%20in%20the%20original%20code%20or%20data.&text=In%20the%20example%2C%20notice%20that,at%20another%20example%20involving%20strings.

packages/plus/src/components/SteppedListCard/Step.tsx
```diff
@@ -0,0 +1,72 @@
...
+          bulletText={String(stepNumber)}
+          prominence={active ? 'strong' : undefined}
+          sentiment={active ? 'primary' : undefined}
+          onClick={() => containerData.setCurrentStep(stepNumber)}
+        >
+          <CustomText as="h3" variant={active ? 'bodyStrong' : 'body'}>
+            {stepTitle}
+          </CustomText>
+        </StepList.Item>
+      )}
```
matthprost, line 69: You can simplify it I think and add `completed` condition inside of the props

matthprost (changes requested):

PR #3526, opened 2024-03-05
feat(List): expandCTA property

fabienhebert:
## Type

- Enhancement

### Summarise concisely:

Row supports expandable content with the prop `expandable`, row becomes then clickable to expand, but expand feature is hard to discover, adding `expandCTA` to the List component, add a dedicated column and a dedicated button to expand row


packages/ui/src/components/List/ListContext.tsx
```diff
@@ -39,12 +40,14 @@ type ListProviderProps = {
   children: ReactNode
   autoCollapse: boolean
   selectable: boolean
+  expandCTA: boolean
```
matthprost, line 43: Im not a big fan of the wording for this variable, `expandButton` might be more in accordance to what we use among components

packages/ui/src/components/List/Row.tsx
```diff
@@ -211,6 +213,17 @@ export const Row = forwardRef(
...
         ) : null}
+        {expandCTA ? (
+          <Cell preventClick>
+            <Button
+              disabled={disabled || !expandable}
+              icon={expandedRowIds[id] ? 'arrow-up' : 'arrow-down'}
+              onClick={toggleRowExpand}
+              size="xsmall"
+              sentiment="neutral"
+            />
```
matthprost, line 224: It should be a variant `ghost`

matthprost (changes requested):

vincentaudebert (approved):

lisalupi (approved):

packages/ui/src/components/List/constants.tsx
```diff
@@ -1 +1,2 @@
 export const SELECTABLE_CHECKBOX_SIZE = 24
+export const EXPAND_CTA_SIZE = 24
```
matthprost, line 2: You can rename this one too 

```suggestion
export const EXPAND_BUTTON_SIZE = 24
```

matthprost (approved):

PR #3532, opened 2024-03-07
refactor(Icon): system icons refactor

lisalupi:
## Type

- Refactor

#### What is expected?

New system icons with 4 variants (outlined/filled and small/large).

1. Imported the new icons

2. Added the variants

3. Updates every component & snapshot using the icons

Currently no test for the component Icon from @ultraviolet/icons


mhervouet:
[system-icon-fix.zip](https://github.com/scaleway/ultraviolet/files/14527362/system-icon-fix.zip)

Some icons were not exported right : folder, profile, arrow-right-bottom. Here's a .zip with the correct icons.

mhervouet:
`Delete` icon seems smaller than others.

packages/icons/src/components/Icon/index.tsx
```diff
@@ -468,23 +136,53 @@ export const Icon = forwardRef<SVGSVGElement, IconProps>(
...
 
+    if (size === 'small' || size === 16) {
+      SystemIcon = StyledIcon(
+        SMALL_ICONS[
+          variant === 'outlined' && name in SMALL_ICONS.default
+            ? 'outline'
+            : 'default'
+        ][name] || SMALL_ICONS.default.alert,
+      )
+    }
```
matthprost, line 160: Best is try to avoid using `let` and use only `const` as it encourage immutability, optimization and prevent from doing errors.

Maybe using a `useMemo` and adding if/else condition into it might work

matthprost (changes requested):

vincentaudebert (approved):

matthprost (approved):

mhervouet (approved):

PetmePant (approved):

mhervouet (approved):

PR #3537, opened 2024-03-08
feat: new navigation

matthprost:
## Type

- Feature

### Introducing: Navigation ✨

Screen Recording 2024-03-25 at 15 45 21

Navigation is a complex component needed for many plateforme and flexible enough to match all their needs. The Navigation implement multiple features such as an expanded version and smaller collapsed version.

It also implement Items and that can be nested into each others and will be shown as category items.
Screenshot 2024-03-25 at 15 47 33

The items are split in two: button when expanded and menu when collapsed.


PetmePant (changes requested):
Some quick issues that I found from the first scanning:

- Spacing between elements should be 2px (now its 0px)
- Pressed background color is different (primary), it should be background neutral weak hover and the initial one background neutral weak
- Similar states are missed when there is an active page
- I cant review the states of an active page and the functionality of pin
- When there is an active item in a group, there is no identification in the group main item (the background should change, check subnav states)
- The transition of the pin button appearance should have more delay (i cant find the current set but it should be smoother and not that instance and definitely a bit more delay)
- Typography for subnav item should be bodySmallStrong and neutralStong text and accordingly the hover color to change.
- Nav items dont have color change in the text from text weak on hover to text weak hover
- In the collapsed version the menu should be triggered on hover and not on click
- The top elements that have no sub items when they are collapsed they should have a tooltip when user hover them to understand the name of the group
- Menu items should have no borders 
- Menu item have no hover/pressed states (not sure if this is part of menu update)
- Collapse/Expand button have no tooltip when you hover it so users can understand what it does (its different depending the state)
- Shadow is missing from the footer nav when there is overflow of elements. 
- Skeleton is missing when loading (for initial page loadings)
- Pinned item on the top section should be always visible and not after a user pins an element. (it should have all the states)
- When pinning/unpinning an element i see no toaster being triggered to indicate that is added
- It seems that nav group header has 16px padding top but it looks that the space is more than this. (mb its an issue with the divider on top)
- Counter is no needed next to the pinned elements (we have other functionality with toaster, we dont want to showcase this limitation since it will be also edge case to have more than that)
- Scaleway logo should also support link (home return)
- I cant review mobile version
- The border in the footer should appear only when there is an overflow along with the shadow (previously repored)

Some images but majority of those deigns are available in figma specs:

Screenshot 2024-03-21 at 7 42 22 PM
Screenshot 2024-03-21 at 7 48 03 PM

packages/plus/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,333 @@
...
+  className,
+}: NavigationContentProps) => {
+  const sliderRef = useRef<HTMLDivElement>(null)
+  const navigationRef = useRef<HTMLDivElement>(null)
+
+  const { expanded, setExpanded, animation, setAnimation, locales } =
+    useNavigation()
+
+  // This function will be triggered when expand/collapse button is clicked
+  const triggerExpand = useCallback(() => {
```
fabienhebert, line 129: Really optional: just naming, mb `toggle` is more accurate than `trigger` ?

packages/ui/src/components/Expandable/__stories__/Controlled.stories.tsx
```diff
@@ -9,14 +9,23 @@ export const Controlled: StoryFn<typeof Expandable> = args => {
...
+      }}
+    >
       <Button icon={toggled ? 'minus' : 'plus'} onClick={toggle}>
         Click me to {toggled ? 'hide' : 'show'} content
       </Button>
       <Expandable {...args} opened={toggled}>
-        I&lsquo;m a visible Expandable content
+        <Stack gap={1}>
+          <div>I&lsquo;m a visible Expandable content</div>
+          <div>feofpwefok weopkf powekf </div>
```
fabienhebert, line 25: I knew you more corporate @matthprost 😂

fabienhebert (approved):

packages/plus/src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+const Container = styled.div`
+  background: ${({ theme }) => theme.colors.neutral.background};
+  display: flex;
+  flex-direction: column;
+
+  width: ${NAVIGATION_WIDTH}px;
+
+  &[data-expanded='true'][data-animation='false'] {
+    max-width: 320px;
```
vincentaudebert, line 32: NAVIGATION_MAX_WIDTH ?

packages/ui/src/components/Expandable/__stories__/Controlled.stories.tsx
matthprost: Oups 😅

packages/plus/src/components/Navigation/constants.ts
```diff
@@ -0,0 +1,35 @@
+import { keyframes } from '@emotion/react'
+
+export const NAVIGATION_MIN_WIDTH = 220 // in px
```
philibea, line 3: should we add the possibility to customize theses constants ?

packages/plus/src/components/Navigation/components/PinnedItems.tsx
```diff
@@ -0,0 +1,31 @@
...
+  if (pinnedItems.length > 0 && pinnedFeature) {
+    return (
+      <Item
+        label={locales['navigation.pinned.item.group.label']}
+        categoryIcon="pin"
+        toggle={false}
+        type="pinnedGroup"
+      >
+        {pinnedItems.map(item => (
+          <Item key={item} label={item} type="pinned" />
```
philibea, line 24: missing key ?

packages/plus/src/components/Navigation/constants.ts
matthprost: I fixed this by adding a prop `width` and `onWidthChange` that will allow the dev to save and manipulate the width as they need. There is also a complete example in the story

PetmePant (approved):
All cases seems to work functionally as it was expected in the initial release.

Next minor things for this first release are more related to some category icons colors for the top links and the menu style updates

PR #3542, opened 2024-03-11
fix: add container id to toaster and notification

lisalupi:
## Type

- Enhancement

#### What is expected?

Added a prop `containerId` to toaster and notification components to be able to use multiple Containers simultaneously. 

1. new prop to `NotificationContainer`, `notification`, `ToastContainer`and `toast` 

2. Stories to showcase the prop

### DIFF
--- a/packages/ui/src/components/Notification/index.tsx
+++ b/packages/ui/src/components/Notification/index.tsx
@@ -47,6 +47,7 @@ const styles = {
     ${PREFIX}__toast-body {
       margin: 0;
       padding: 0;
+      display: none;
     }
   `,
 }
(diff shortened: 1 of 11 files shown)


packages/ui/src/components/Notification/index.tsx
```diff
@@ -47,6 +47,7 @@ const styles = {
     ${PREFIX}__toast-body {
       margin: 0;
       padding: 0;
+      display: none;
```
matthprost, line 50: Why do we need this?

matthprost (changes requested):

packages/ui/src/components/Notification/index.tsx
lisalupi: I saw an empty vertical space when using the notification

matthprost (approved):

BABAK0T0 (approved):

JulienSaguez (approved):

PR #3555, opened 2024-03-13
fix(NumberInputV2): only use number as value type

fabienhebert:
## Type

- Enhancement

### Summarise concisely:

`NumberInputV2` : value is now typed as `number` and not as `string`
`NumberInputV2` : if `max` value is not provided, it will use `Number.MAX_SAFE_INTEGER` as max value
`NumberInputV2` : support keydown `Enter` as `onChange`, in addition of onBlur


vincentaudebert (approved):

packages/ui/src/components/NumberInputV2/index.tsx
```diff
@@ -372,36 +379,26 @@ export const NumberInputV2 = forwardRef(
                   name={name}
                   id={localId}
                   placeholder={placeholder}
+                  onKeyDown={event => {
+                    if (event.key === 'Enter') {
+                      onChangeValue(localRef.current?.value ?? '')
```
johnrazeur, line 384: We are gonna get a `NaN` by the `parseInt` (L.261) if we pass `''`.

alexandre-combemorel (approved):

johnrazeur (approved):

JulienSaguez (approved):

matthprost (approved):

PR #3599, opened 2024-03-25
fix(CheckboxGroup): to allow partial and total required

matthprost:
## Type

- Enhancement

#### What is expected?

- Fix `<CheckboxGroupField />` to make all checkbox of the group as required when set `required` prop
- Fix `<CheckboxGroupField.Checkbox />` to have prop `required` and allow partial requirement for the group to be valid
- Fix `<CheckboxGroup.Checkbox />` to have prop `required`

Screen Recording 2024-03-25 at 17 38 00

### DIFF
--- a/packages/form/src/components/CheckboxGroupField/index.tsx
+++ b/packages/form/src/components/CheckboxGroupField/index.tsx
@@ -11,7 +11,7 @@ const arraysContainSameValues = (array1: string[], array2: string[]) => {
     return false
   }
 
-  return array2.map(value => array1.includes(value)).every(Boolean)
+  return array2.every(value => array1.includes(value))
 }
 
 type CheckboxGroupFieldProps<
--- a/packages/form/src/components/CheckboxGroupField/index.tsx
+++ b/packages/form/src/components/CheckboxGroupField/index.tsx
@@ -7,14 +7,11 @@ import { useErrors } from '../../providers'
 import type { BaseFieldProps } from '../../types'
 
 const arraysContainSameValues = (array1: string[], array2: string[]) => {
-  if (array1.length !== array2.length) {
+  if (array1.length === 0) {
     return false
   }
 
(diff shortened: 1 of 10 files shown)


packages/form/src/components/CheckboxGroupField/index.tsx
```diff
@@ -38,14 +56,47 @@ export const CheckboxGroupField = <
...
+          }
+
+          return null
+        })?.filter(Boolean) ?? []
+
+      if (!required && arraysContainSameValues(value, requiredChildren)) {
+        return undefined
+      }
+
+      if (value.length >= Children.count(children)) {
```
johnrazeur, line 81: With 2 not required checkbox and 1 required checkbox.
If we check the 2 not required checkbox, the field gonna be valid.

matthprost: It should work just fine this condition is invalid in that case too as you have only 2 checkboxes out of 3 checked


https://github.com/scaleway/ultraviolet/assets/15812968/df239efe-4c3e-4352-8527-97b5e73d67a1

packages/form/src/components/CheckboxGroupField/index.tsx
```diff
@@ -7,14 +7,11 @@ import { useErrors } from '../../providers'
...
-  if (array1.length !== array2.length) {
+  if (array1.length === 0) {
     return false
   }
 
-  const sortedArray1 = array1.slice().sort()
-  const sortedArray2 = array2.slice().sort()
-
-  return sortedArray1.every((element, index) => element === sortedArray2[index])
+  return array2.map(value => array1.includes(value)).every(Boolean)
```
johnrazeur, line 14: `return array2.every(value => array1.includes(value))`

johnrazeur (approved):

JulienSaguez (approved):

PR #3620, opened 2024-04-02
fix(Navigation): slider and behavior fixes

matthprost:
## Type

- Enhancement

#### What is expected?

- Fix slider to have better border and a 8px width to leave margin for user to resize the navigation
- Fix menu on collapsed version: when clicking on a item it should close the menu
- Fix tab order on collapsed version on standalone items like "Project dashboard" in story. Tooltip is adding an tab on each item it wrap, in that case the items is a button so no tab index is needed.


johnrazeur:
Not directly link to this PR but hidden item can have focus:

https://github.com/scaleway/ultraviolet/assets/1339931/e9c0dfbc-b50b-4f2c-9628-34f57d0f65a7

matthprost:
> Not directly link to this PR but hidden item can have focus:
> 
>  Screen.Recording.2024-04-03.at.13.40.46.mov

Good catch! I fixed it, it was due to expandable allowing tabulation for its hidden children.

johnrazeur (approved):

PR #3629, opened 2024-04-05
fix(SelectInput): null child type error

johnrazeur:
## Type

- Bug

#### What is expected?

Fix type error when child is null or undefined.

### DIFF
--- a/packages/ui/src/components/SelectInput/index.tsx
+++ b/packages/ui/src/components/SelectInput/index.tsx
@@ -693,7 +693,7 @@ const FwdSelectInput = ({
             {
               ...child.props,
               label: child.props.children,
-            } as SelectOption,
+            },
           ]
         }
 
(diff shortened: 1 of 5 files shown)


matthprost (approved):

packages/ui/src/components/SelectInput/index.tsx
```diff
@@ -680,13 +686,19 @@ const FwdSelectInput = ({
...
-          }) as SelectOption,
-      ),
+      Children.toArray(children).reduce<SelectOption[]>((acc, child) => {
+        if (React.isValidElement<{ children: string; value: string }>(child)) {
+          return [
+            ...acc,
+            {
+              ...child.props,
+              label: child.props.children,
+            } as SelectOption,
```
alexandre-combemorel, line 696: satisfy ?

alexandre-combemorel (approved):

packages/ui/src/components/SelectInput/index.tsx
johnrazeur: Fixed

PR #3651, opened 2024-04-09
feat(SelectInputV2): new select input

lisalupi:
## Type

- Feature
- 
#### What is expected?

New select input to replace the old one. Do not use react-select.

(Describe what you did)

1. New component 

2. Tests & storybook

3. Deprecated v1
4. New prop "onKeyDown" for `<SelectInputV2 />` (because it was needed for selectInputV2)

| url  | Capture d’écran 2024-04-09 à 15 44 54 | Capture d’écran 2024-04-09 à 15 44 05 |


mhervouet:
Capture d’écran 2024-04-09 à 16 03 14

Need to have a 4px spacing between field and helper text.

mhervouet:
Capture d’écran 2024-04-09 à 16 04 52

Badges looks to have a too much padding on sides (16px instead of 8px)

mhervouet:
Capture d’écran 2024-04-09 à 16 05 33

mhervouet:
Capture d’écran 2024-04-09 à 16 06 15

On some versions the dropdown is not aligned with the field

mhervouet:
Capture d’écran 2024-04-09 à 16 06 21

In empty state the link should be center aligned

mhervouet:
Capture d’écran 2024-04-09 à 16 06 44

Issue with the tags in multiselect inputs

mhervouet:
Capture d’écran 2024-04-09 à 16 09 06

There was an update on spacing to match Menu. Check the new spacing here : https://ultraviolet.scaleway.com/6dd9b5c45/p/3882cf-selectinput/t/29563b

mhervouet:
Capture d’écran 2024-04-09 à 16 09 29

Need to use a shadow for fixed background instead of big border.

mhervouet:
Capture d’écran 2024-04-09 à 16 10 23
Capture d’écran 2024-04-09 à 16 10 44

Need to fix the vertical alignment icon / text

mhervouet:
Capture d’écran 2024-04-09 à 16 10 32

Is it possible to align the text and left blank under the icon ?

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -0,0 +1,337 @@
...
+      gap="1"
+      wrap="nowrap"
+      ref={refTag}
+      width={width}
+      alignItems="center"
+    >
+      {nonOverflowedValues.map((option, index) => (
+        <CustomTag
+          data-testid="selected-options-tags"
+          key={option ? option.value : index}
```
matthprost, line 147: Won't it be risky that two options have the same value?

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -0,0 +1,337 @@
...
+            !readOnly
+              ? event => {
+                  event.stopPropagation()
+                  const newSelectedValues = selectedValues
+                    ?.filter(val => val !== option)
+                    .map(val => val?.value)
+                  setSelectedValues(
+                    selectedValues?.filter(val => val !== option),
+                  )
+                  onChange?.(newSelectedValues)
```
matthprost, line 160: I think it can be simplified to avoid doing the same filter twice, wdyt?

```tsx
                  const newSelected = selectedValues
                    ?.filter(val => val !== option)
                  setSelectedValues(newSelected)
                  onChange?.(newSelectedValues.map(val => val?.value))
```

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -0,0 +1,337 @@
...
+    if (error) {
+      return 'danger'
+    }
+    if (success) {
+      return 'success'
+    }
+
+    return 'neutral'
+  }, [error, success])
+  useEffect(() => {
```
matthprost, line 222:
```suggestion
  }, [error, success])
  
  useEffect(() => {
```

matthprost (changes requested):

packages/ui/src/components/SelectInputV2/SelectBar.tsx
lisalupi: Shouldn't the values be unique ?

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,585 @@
...
+  overflow: scroll;
+  padding: 0px;
+`
+const DropdownGroup = styled(Stack)`
+  height: ${({ theme }) => theme.space['4']};
+  background-color: ${({ theme }) => theme.colors.neutral.backgroundWeak};
+  position: sticky;
+  top: 0px;
+  height: 32px;
+  z-index: 1;
```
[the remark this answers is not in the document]
lisalupi: The checkbox goes on top of the group name without it

packages/ui/src/components/SelectInputV2/SelectBar.tsx
matthprost: good question haha I'd say yes but technically the user could set 2 same values

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,896 @@
...
+    ${({ theme }) => theme.space['2']} ${({ theme }) => theme.space['1.5']}
+    ${({ theme }) => theme.space['2']};
+  margin-left: ${({ theme }) => theme.space['0.5']};
+  margin-right: ${({ theme }) => theme.space['0.5']};
+  color:  ${({ theme, disabled }) => (disabled ? `${theme.colors.neutral.textDisabled}` : `${theme.colors.neutral.text}`)};
+
+  &:hover, :focus {
+    background-color: ${({ theme, disabled }) => (disabled ? `${theme.colors.neutral.backgroundStrongDisabled};` : `${theme.colors.primary.background};`)}
+    color: ${({ theme, disabled }) => (disabled ? `${theme.colors.neutral.textStrongDisabled}` : `${theme.colors.primary.text}`)};
+    cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
```
matthprost, line 112: Disabled state shows a pointer on the input there is something wrong, it should show `not-allows` instead

matthprost (changes requested):

packages/ui/src/components/SelectInputV2/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,181 @@
...
+    } else {
+      const possibleOptions = [...options].filter(option => !option.disabled)
+
+      if (possibleOptions.length > 0) {
+        return possibleOptions[0]
+      }
+    }
+  }
+
+  return 'NO_MATCH'
```
fabienhebert, line 65: is there a way to customize/localize text providing `locales` prop like some component in `ultraviolet/plus ?

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -0,0 +1,388 @@
...
+  selectedValues: OptionType[]
+  setSelectedValues: Dispatch<SetStateAction<OptionType[]>>
+  onChange?: (value: string[]) => void
+  overflowed: boolean
+  overflowAmount: number
+  setAllSelected: Dispatch<SetStateAction<boolean>>
+}
+const StateStack = styled(Stack)`
+  padding-right: ${({ theme }) => theme.space['2']};
+  right: 0px;
```
fabienhebert, line 58: since Stack is not positionned (absolute or fixed), this `right` has no effect no ? (if it does mb a comment could help)

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -0,0 +1,388 @@
...
+  display: flex;
+`
+
+const StackTags = styled(Stack, {
+  shouldForwardProp: prop => !['width'].includes(prop),
+})<{ width: number | undefined }>`
+  max-width: ${({ width }) => width}px;
```
fabienhebert, line 65: May be call the prop `maxWidth` and since `width` prop seems optionnal, mb it must be handled

mhervouet:
Capture d’écran 2024-04-18 à 10 01 18
Items must have a 4px radius.

mhervouet:
Capture d’écran 2024-04-18 à 10 02 56
In this "Left" example, should the icon be at the left of the text ?

mhervouet:
Capture d’écran 2024-04-18 à 10 03 13
The button "Load more" must have the same padding than items (2px)

packages/ui/src/components/SelectInputV2/SearchBarDropdown.tsx
lisalupi: this text will never be shown to the user, it is just a way to know when there is no match and thus do nothing in that case, so I don't think having locales is important here?

lisalupi: this text will never be shown to the user, it is just a way to know when there is no match and thus do nothing in that case, so I don't think having locales is important here?

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,916 @@
...
+  position: static;
+  text-align: left;
+`
+const EmptyState = styled(Stack)`
+  padding: ${({ theme }) => theme.space[2]};
+`
+
+const moveFocusDown = () => {
+  const listOptions: ChildNode[] = []
+  const options = document.querySelectorAll('#items')
```
DorianMaliszewski, line 135: Not sure that this is the good may be its better to focus focusable elements and not the entire :
"#items  button:not(disabled),input:not(disabled)" ? :thinking:

DorianMaliszewski: And for multiselect place a tabIndex={-1} on checkbox ?

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,916 @@
...
+          setSelectedGroups([...selectedGroups, group])
+        }
+      }
+    } else {
+      setSelectedValues([clickedOption])
+      onChange?.([clickedOption.value])
+    }
+    setIsDropdownVisible(multiselect) // hide the dropdown on click when single select only
+  }
+  const displayOption = (option: OptionType) => {
```
DorianMaliszewski, line 380: Is It possible to export this in a different file ? Better for maintenance and readability

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -0,0 +1,388 @@
...
+`
+const StyledInputWrapper = styled(Stack, {
+  shouldForwardProp: prop =>
+    !['size', 'isDropdownVisible', 'state', 'disabled', 'readOnly'].includes(
+      prop,
+    ),
```
DorianMaliszewski, line 71: can we use data/aria attributes ?

DorianMaliszewski: Or use better selectors like :has(*:disabled)

packages/ui/src/components/SelectInputV2/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,181 @@
...
+  setAllSelected,
+  selectAllGroup,
+  selectedGroups,
+  setSelectedGroups,
+  numberOfOptions,
+}: SearchBarProps) => {
+  const handleChange = (search: string) => {
+    if (search.length > 0) {
+      // case insensitive search
+      const regex = RegExp(search, 'gi')
```
DorianMaliszewski, line 89: Not sure that we need 'g' only one match is enough ? If we want to match on multiline maybe 'm' flags is better ?

packages/ui/src/components/SelectInputV2/Dropdown.tsx
DorianMaliszewski: Moreover you can cast querySelectorAll<HTMLButtonElement | HTMLInputElement> and drop downcasting with 'as' below

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,916 @@
...
+}
+const handleKeyDownSelect = (key: string) => {
+  if (key === 'ArrowDown') {
+    moveFocusDown()
+  }
+  if (key === 'ArrowUp') {
+    moveFocusUp()
+  }
+}
+const HandleDropdown = (
```
DorianMaliszewski, line 180: Its not a component better to call this handleDropdownRef or something lowerCamelCase

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,916 @@
...
+      focusedItemRef.current.focus()
+    }
+  }, [defaultSearchValue])
+
+  if (isEmpty) {
+    return (
+      <EmptyState gap={2} alignItems="center">
+        {emptyState ?? (
+          <Text variant="bodyStrong" as="p">
+            No options
```
DorianMaliszewski, line 333: Text is not customizable ?

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,916 @@
...
+    selectAllGroup ? '20px' : theme.space[2]};
+  height: ${({ theme }) => theme.space[4]};
+  text-align: left;
+  margin-bottom: ${({ theme }) => theme.space['0.25']};
+`
+const DropdownItem = styled('button', {
+  shouldForwardProp: prop => !['disabled', 'selected'].includes(prop),
+})<{
+  disabled: boolean
+  selected: boolean
```
DorianMaliszewski, line 90: use aria-selected and you can forward props :)

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,916 @@
...
+`
+
+const DropdownContainer = styled(Stack)`
+  max-height: 256px;
+  overflow-y: scroll;
+  padding: ${({ theme }) => theme.space[0]};
+  padding-bottom: ${({ theme }) => theme.space[0.5]};
+`
+const DropdownGroup = styled(Stack, {
+  shouldForwardProp: prop => !['selectAllGroup'].includes(prop),
```
DorianMaliszewski, line 74: data attributes

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,916 @@
...
+            <DropdownItem
+              key={option.value}
+              disabled={option.disabled}
+              selected={selectedValues.includes(option) && !option.disabled}
+              onClick={() => {
+                if (!option.disabled) {
+                  handleClick(option)
+                }
+              }}
+              aria-label={option.value}
```
DorianMaliszewski, line 753:
```suggestion
              onKeyDown={!option.disabled ? 
                  event =>
                      [' ', 'Enter'].includes(event.key) ? handleClick(option) : null
                   : undefined
               }
              aria-label={option.value}
```

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,916 @@
...
+              </DropdownItem>
+            </Stack>
+          ) : null}
+          {Object.keys(displayedOptions).map(group => (
+            <Stack key={group} gap={0.25}>
+              {displayedOptions[group].length > 0 ? (
+                <DropdownGroup
+                  key={group}
+                  selectAllGroup={selectAllGroup}
+                  justifyContent="center"
```
DorianMaliszewski, line 623:
```suggestion
                  justifyContent="center"
                  role="group"
```

DorianMaliszewski (changes requested):
May be using a Context/Reducer can be better to remove prop drilling

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -0,0 +1,916 @@
...
+                  checked={selectedValues.includes(option) && !option.disabled}
+                  disabled={option.disabled}
+                  value={option.value}
+                  onChange={() => {
+                    if (!option.disabled) {
+                      handleClick(option)
+                    }
+                  }}
+                >
+                  <Stack>{displayOption(option)}</Stack>
```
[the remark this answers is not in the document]
lisalupi: it's useless, I simply forgot to delete it ^^

packages/ui/src/components/SelectInputV2/Dropdown.tsx
lisalupi: When you don't pass an emptyState, it will have a default non-customizable text yes

mhervouet:
Capture d’écran 2024-04-22 à 11 51 17
Issue on tag with badge.

mhervouet (approved):

matthprost (approved):

PR #3672, opened 2024-04-16
fix: workaround for autofocus problem

lisalupi:
## Type

- Bug

#### What is expected?

`autoFocus` doesn't work with modals. For now, we have a workaround

1. Added a story to show how to recreate manually autofocus


.changeset/itchy-mirrors-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/ui": patch
+---
+
+Autofocus problem on `<Modal />`: found a workaround (see storybook)
```
matthprost, line 5: As this is only a story and there is no functional changes, you don't need to push a changeset file 😄

matthprost (changes requested):

BABAK0T0 (approved):

PR #3680, opened 2024-04-17
fix: various illustration bug

lisalupi:
## Type

- Bug

#### What is expected?

Illustrations `generic` and `accountBilling` should correctly render


packages/illustrations/src/various/documentation/generic.svg
matthprost, line 1: Seems like we loose the padding around the image, might be worth the shot to ask designers if it's what is wanted

matthprost (changes requested):

matthprost (approved):

BABAK0T0 (approved):

PR #3693, opened 2024-04-22
fix: change navitem text color

lisalupi:
## Type

- Bug

#### What is expected?

Text in `Navigation.Item` should be textDefault and not textWeak

### DIFF
--- a/.changeset/tall-seahorses-grow.md
+++ b/.changeset/tall-seahorses-grow.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/plus": patch
+---
+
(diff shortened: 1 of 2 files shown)


BABAK0T0 (approved):

PetmePant:
Seems good :)

matthprost (approved):

PetmePant (changes requested):
Text seems as it should but the icons are still in weak 

April 23 Screenshot from Matthias Prost (1)

PetmePant (approved):

PR #3695, opened 2024-04-24
fix(Navigation): export navigation provider and use navigation

matthprost:
## Type

- Enhancement

#### What is expected?

I exported `NavigationProvider` and `useNavigation()` so it can be easy to access to navigation info or change the state of the navigation with this hook.


packages/plus/src/components/Navigation/NavigationProvider.tsx
```diff
@@ -1,66 +1,123 @@
...
+    onClickExpand?.(!expanded)
+    if (navigationRef.current) {
+      navigationRef.current.style.width = ''
+    }
+
+    setAnimation(expanded ? 'collapse' : 'expand')
+
+    setTimeout(() => {
+      setExpanded()
+      // setFooterHasOverflowStyle(isScrollAtBottom())
```
BABAK0T0, line 117: Necessary ?

BABAK0T0 (approved):

packages/plus/src/components/Navigation/NavigationProvider.tsx
matthprost: @BABAK0T0 right I forgot to remove it, thanks!

PR #3717, opened 2024-04-26
feat(Navigation): new pinned feature and reorder

matthprost:
## Type

- Enhancement

### Introducing pin feature on Navigation ✨

Navigation is nice for traveling around in your website but it lacked some customization. Pinned feature is here to overcome this issue and provide a new way to customize your navigation. This feature can be enabled adding `pinnedFeature` on your `<NavigationProvider />`

Screen Recording 2024-04-26 at 17 55 32

Pinned feature allows you to pin / unpin your favorite items until it reach the limit you've set using `pinLimit` on your `<NavigationProvider />`.
It also allows you to re-order your pinned items by using drag n drop and placing it in the order you like.

All your pinned items can be accessed with the hook `useNavigation()`:

```tsx
// pinnedItems = ['myitem1', 'myitem2']
const { pinnedItems } = useNavigation()

// Save in local storage to keep the pinned items on refresh
useEffect(() => {
   localStorage.setItem('pinnedItems', pinnedItems.toString())
}, [pinnedItems])
```

You can of course provide your initially pinned items to the `<NavigationProvider />` by using the prop `initialPinned`.

### DIFF
--- a/packages/plus/src/components/Navigation/components/Item.tsx
+++ b/packages/plus/src/components/Navigation/components/Item.tsx
@@ -58,6 +58,7 @@ const ExpandedPinnedButton = styled(Button)`
 const GrabIcon = styled(Icon)`
   opacity: 0;
   margin: 0 ${({ theme }) => theme.space['0.25']};
+  cursor: grab;
 `
 
 // Pin button when the navigation is collapsed
@@ -362,7 +363,9 @@ export const Item = ({
 
   useEffect(
     () => {
-      registerItem({ [id]: { label, active, onClick } })
+      if (type !== 'pinnedGroup') {
+        registerItem({ [id]: { label, active, onClick } })
+      }
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
     [active, id, label, registerItem],
@@ -731,7 +734,6 @@ export const Item = ({
       <StyledMenuItem
         href={href}
         onClick={() => {
-          console.log('clicked', onClick)
           onClick?.()
         }}
         borderless
--- a/packages/plus/src/components/Navigation/components/Item.tsx
+++ b/packages/plus/src/components/Navigation/components/Item.tsx
@@ -57,6 +57,7 @@ const ExpandedPinnedButton = styled(Button)`
 
 const GrabIcon = styled(Icon)`
   opacity: 0;
+  margin: 0 ${({ theme }) => theme.space['0.25']};
 `
 
 // Pin button when the navigation is collapsed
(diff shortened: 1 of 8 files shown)


PetmePant (changes requested):
1) Issue with menu and active item while pinned. The pin item is not getting hidden when and overlaps with badge. (check video)

https://github.com/scaleway/ultraviolet/assets/8331633/587a3393-160e-44fc-9bfd-248655290d74

2) Alignment issue for the title/label of each pinned item and nav group header.
Screenshot 2024-04-29 at 12 12 50 PM

3) Pinned section should be always visible and with an empty state when no elements are pinned.

4) While reorder we should disable the initial nav item that is dragged

5) Drag & drop handler (cursor) should appear when you hover the drag icon/action and whole you drag and drop the cursor should be different.

6) When you pin an item you should have a toaster message

7) No items from the pinned section can be active similar to the rest of the items.

fabienhebert (approved):

packages/plus/src/components/Navigation/components/PinnedItems.tsx
```diff
@@ -10,20 +45,100 @@ export const PinnedItems = () => {
...
+        <div>
+          <Item
+            label={locales['navigation.pinned.item.group.label']}
+            categoryIcon="pin"
+            categoryIconVariant="neutral"
+            toggle={toggle}
+            type="pinnedGroup"
+          >
+            {pinnedItems.map((item, index) => (
+              <div style={{ position: 'relative' }} key={item}>
```
DorianMaliszewski, line 101: ? maybe use styled component

packages/plus/src/components/Navigation/components/PinnedItems.tsx
```diff
@@ -10,20 +45,100 @@ export const PinnedItems = () => {
...
+                />
+                <Item
+                  label={item}
+                  type="pinned"
+                  index={index}
+                  toggle={toggle}
+                />
+              </div>
+            ))}
+            <div style={{ position: 'relative' }}>
```
DorianMaliszewski, line 115: Same here use a styled component

packages/plus/src/components/Navigation/components/Item.tsx
```diff
@@ -46,45 +49,55 @@ const ExpandedPinnedButton = styled(Button)`
...
 `
 
-const StyledMenuItem = styled(MenuV2.Item)`
+const StyledBadge = styled(Badge)``
+
+const StyledMenuItem = styled(MenuV2.Item, {
+  shouldForwardProp: prop => !['isPinnable'].includes(prop),
+})<{
+  isPinnable?: boolean
+}>`
```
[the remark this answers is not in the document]
matthprost: Cannot be used as MenuV2.Item does not spread all it's props

PR #3722, opened 2024-04-29
fix(SelectInputV2): use string or string[] as value prop instead of w…

DorianMaliszewski:
## Type

- Enhancement

#### What is expected?

Instead of passing a whole option as value prop we can pass string or string[] and it affect the SelectInput component. It makes the component more easy to use and compliant with basic select tags.

1. Change all checking type, includes, ... by OptionType to string

2. Add some props on SelectInputFieldV2 story to test it 

3. Remove format function in SelectInputFieldV2 because we dont need it

### DIFF
--- a/.changeset/fluffy-windows-dance.md
+++ b/.changeset/fluffy-windows-dance.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/ui": minor
+---
+
+fix(SelectInputV2): pass string or string[] for values instead of an option (more compliant to native select tag). component is now controllable
--- a/.changeset/fluffy-windows-dance.md
+++ b/.changeset/fluffy-windows-dance.md
@@ -2,4 +2,4 @@
 "@ultraviolet/ui": minor
 ---
 
-fix(SelectInputV2): pass string or string[] for values instead of an option (more compliant to native select tag). component is now controllable
+Fix <SelectInputV2 /> component to pass string or string[] for values instead of an option (more compliant to native select tag). component is now controllable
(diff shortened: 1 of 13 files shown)


.changeset/fluffy-windows-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/ui": minor
+---
+
+fix(SelectInputV2): pass string or string[] for values instead of an option (more compliant to native select tag). component is now controllable
```
matthprost, line 5:
```suggestion
Fix <SelectInputV2 /> component to pass string or string[] for values instead of an option (more compliant to native select tag). component is now controllable
```

matthprost (changes requested):

packages/form/src/components/SelectInputFieldV2/__stories__/Multiple.stories.tsx
```diff
@@ -5,4 +5,14 @@ import { planets } from './resources'
...
+  <SelectInputFieldV2
+    selectAllGroup
+    selectAll={{
+      label: 'Select all',
+    }}
+    name="options"
+    multiselect
+    options={planets}
+  />
+)
```
matthprost, line 18: There is already a story for select all `packages/ui/src/components/SelectInputV2/__stories__/SelectAll.stories.tsx`

matthprost (changes requested):

packages/form/src/components/SelectInputFieldV2/__stories__/Multiple.stories.tsx
matthprost: ok my bad it's the field

lisalupi (approved):

PR #3740, opened 2024-05-02
fix(Navigation): some fixes

matthprost:
## Type

- Bug

#### What is expected?

Fix `<NavigationComponent />` to have correct empty state, pin / unpin to be div instead of button and fix on click on pin / unpin


philibea:
Missing some tests right ?

philibea (approved):

PR #3745, opened 2024-05-03
fix(EmptyState): reduce text sizes on small size

matthprost:
## Type

- Enhancement

#### What is expected?

Update `<EmptyState />` component to have smaller text on small size

Before:

Screenshot 2024-05-03 at 10 32 43

After:

Screenshot 2024-05-03 at 10 32 01


mhervouet:
In small size : 
- buttons should be medium
- space between text and button 16px (not 24px)

matthprost:
> In small size :
> 
> * buttons should be medium
> * space between text and button 16px (not 24px)

Button is set by the user directly so it will have to be correctly set on console when using empty state.
I'm fixing the space

fabienhebert (approved):

lisalupi (approved):

PR #3748, opened 2024-05-03
Feat/add fallback font theme

alexandre-combemorel:
## Summary
Add matching fallback font in case fonts are not yet loaded

## Type
- Enhancement

#### What is expected?

If fonts fail to load or are not yet loaded, then it should display the closest type of font
Based on google font fallback
Screenshot 2024-05-03 at 16 15 57

### DIFF
--- a/packages/themes/src/themes/console/dark.ts
+++ b/packages/themes/src/themes/console/dark.ts
@@ -371,7 +371,7 @@ export const darkTheme = {
   theme: 'dark',
   typography: {
     body: {
-      fontFamily: 'Inter',
+      fontFamily: 'Inter, sans-serif',
       fontSize: '16px',
       fontWeight: 'Regular',
       letterSpacing: '0',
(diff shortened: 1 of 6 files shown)


packages/themes/src/themes/console/dark.ts
```diff
@@ -371,7 +371,7 @@ export const darkTheme = {
   theme: 'dark',
   typography: {
     body: {
-      fontFamily: 'Inter',
+      fontFamily: 'Inter, sans-serif',
```
matthprost, line 374: As written on top of each files those are generated based on what design send us back from figma tokens. If you want to change this you need to change it from figma directly or see with a designer to do so.

matthprost (changes requested):

packages/themes/src/themes/console/dark.ts
alexandre-combemorel: I didn't check properly the top of the file. I will have a look with designer or eventually check the script token generation

alexandre-combemorel:
This is what it look like when its broken

Capture_d_écran_2024-05-02_à_15 51 08
Capture_d_écran_2024-05-02_à_15 51 11

matthprost:
The update has been merged here: https://github.com/scaleway/ultraviolet/pull/3802

You can rebase to check changes or close the PR 😄

PR #3773, opened 2024-05-14
fix: font size on inputs

lisalupi:
## Type

- Refactor

#### What is expected?

Font-size changes for small and medium inputs : 14px instead of 16px

1. Changed font-size front "body" to "bodySmall" for inputs & their label when size="medium" or "small"

| Page |   TextInputV2  |
| :--- | :--------: 
| Before | Capture d’écran 2024-05-14 à 12 02 08 | 
| After | Capture d’écran 2024-05-14 à 12 01 39 |

### DIFF
--- a/packages/ui/src/components/SelectInputV2/__tests__/index.test.tsx
+++ b/packages/ui/src/components/SelectInputV2/__tests__/index.test.tsx
@@ -2,7 +2,7 @@ import { beforeAll, describe, expect, jest, test } from '@jest/globals'
 import { fireEvent, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import type { ReactNode } from 'react'
-import { act } from 'react-dom/test-utils'
+import { act } from 'react'
 import { SelectInputV2 } from '..'
 import {
   renderWithTheme,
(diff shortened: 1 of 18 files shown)


packages/ui/src/components/SelectInputV2/__tests__/index.test.tsx
```diff
@@ -780,7 +780,7 @@ describe('SelectInputV2', () => {
 
     await userEvent.keyboard('[Backspace]')
     await userEvent.keyboard('[Enter]')
-  }, 6000)
+  }, 10000)
```
matthprost, line 783: I didn't see it on SelectInputV2, but why setting a timeout? There is a general timeout for tests is it too low for select input to work?

matthprost (changes requested):

packages/ui/src/components/SelectInputV2/__tests__/index.test.tsx
lisalupi: Yes, sometimes the default one is not enough and the test fails

PetmePant (approved):

matthprost (approved):

PR #3775, opened 2024-05-14
feat(Toggle): error state

lisalupi:
## Type
- Feature

#### What is expected?

`<Toggle />`: Added prop `error`


mhervouet (approved):

packages/ui/src/components/Toggle/index.tsx
```diff
@@ -216,7 +242,18 @@ export const Toggle = forwardRef(
...
+                as="p"
+                variant="bodySmall"
+                prominence="default"
+                sentiment="danger"
+                disabled={disabled}
+              >
+                {error}
+              </Text>
+            ) : null}
+            {helper && typeof error !== 'string' ? (
```
matthprost, line 256: Not sure about this condition. Here you say that if helper exists and type of error is not a string then we can show helper. So either we want to show the error, if it's not a type of string, or we want to show the helper if error just doesn't exists.

In that case I think we want the second one, showing the helper if error doesn't exists and showing the error if it exists.

matthprost (changes requested):

packages/ui/src/components/Toggle/index.tsx
lisalupi: If the error is not a string it is a boolean, so there is not "error text" to show. Do we still want to hide the helper in that case?

PetmePant (approved):

packages/ui/src/components/Toggle/index.tsx
matthprost: @lisalupi I believe we should as we are in a error state, helper is only shown in normal state.

matthprost (approved):

PR #3781, opened 2024-05-16
feat(Stepper): various enhancements and refactor

lisalupi:
## Type

- Enhancement
- Refactor

#### What is expected?

`<Stepper />` enhancement: 
- Added a new disabled state
- Added possibility to navigate throughout the steps for the user by clicking on the bullets
- Refactor code : use `<Stepper.Step />` for the steps instead of a list of children
- New style overall

1. Updated Stepper style

2. Created new subComponent to deal with steps

3. Added interactivity


packages/ui/src/components/Stepper/__stories__/index.stories.tsx
```diff
@@ -3,8 +3,13 @@ import { Stepper } from '..'
 
 export default {
   component: Stepper,
+  decorators: [StoryComponent => <StoryComponent />],
   title: 'Components/Navigation/Stepper',
```
matthprost, line 7: You can add `subcomponents` here and add `Stepper.Step` into it so we will see the prop documented on storybook.

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -95,101 +183,170 @@ const StyledContainer = styled.div<{
...
+  index: number
+  disabled?: boolean
+  title?: ReactNode
+  children?: ReactNode
 }
 
-const StepperNumbers = ({
-  temporal,
+const Step = ({
+  index,
```
matthprost, line 194: I'm thinking that maybe the index could be guessed and provided in the parent, wdyt?

matthprost (changes requested):

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -1,17 +1,85 @@
...
+  animated?: boolean
+  /**
+   * When true, the user can navigate through the steps by clicking on the bullets
+   */
+  interactive?: boolean
+  /**
+   * Number of the active step
+   */
+  selected?: number
+  children: ReactNode[]
```
matthprost, line 41:
```suggestion
  children: ReactNode
```

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -94,102 +231,139 @@ const StyledContainer = styled.div<{
...
-type StepperNumbersProps = {
-  temporal: Temporal
-  children: ReactNode
-  CurrentStep: number
-  size?: 'small' | 'medium'
-}
-
-const StepperNumbers = ({
-  temporal,
+const Step = ({
```
matthprost, line 234: Can you move step into a separate file? The current file seems a bit big now, it can be nice to split in two

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -21,27 +89,92 @@ const loadingAnimation = keyframes`
...
+  isActive: boolean
+  isDone: boolean
+  'data-disabled': boolean
+  'data-interactive': boolean
+}>`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: flex-start;
+  transition: all 300ms;
```
matthprost, line 127:
```suggestion
  transition: all 300ms;
```

This can have side effects with animations, better define exactly what you want to have an animation applied on

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -94,102 +231,139 @@ const StyledContainer = styled.div<{
...
-  size?: 'small' | 'medium'
-}
-
-const StepperNumbers = ({
-  temporal,
+const Step = ({
+  index = 0,
+  onClick,
+  disabled = false,
+  title,
```
matthprost, line 238: props misses `className` and `data-testId`

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -94,102 +231,139 @@ const StyledContainer = styled.div<{
...
-
-    <StyledText
-      as="span"
-      variant={size === 'medium' ? 'body' : 'bodySmall'}
-      prominence={temporal === 'next' ? 'weak' : 'default'}
+          onClick?.(index)
+        }
+      }}
+      data-disabled={disabled}
+      data-testid={`stepper-step-${index}`}
```
matthprost, line 268: Here you could say something like `data-testid={dataTestId ?? `stepper-step-${index}`}` after adding data test id

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -1,17 +1,85 @@
...
+  disabled?: boolean
+  /**
+   * Title of the step
+   */
+  title?: ReactNode
+  /**
+   * @deprecated
+   * Do not use with Stepper.Step: use prop "title" instead
+   */
+  children?: ReactNode
```
matthprost, line 67: I don't understand how can this be deprecated as it's brand new? 😅

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -21,27 +89,92 @@ const loadingAnimation = keyframes`
...
+const StyledStepContainer = styled.div<{
+  isActive: boolean
+  isDone: boolean
+  'data-disabled': boolean
+  'data-interactive': boolean
+}>`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: flex-start;
```
matthprost, line 126: This could be a `Stack` with all properties added into JSX

packages/ui/src/components/Stepper/index.tsx
```diff
@@ -1,17 +1,85 @@
...
+  children?: ReactNode
+}
+
+type StepperContentProps = {
+  children: ReactNode
+  animated: boolean
+  separator: boolean
+}
+
+const State = createContext<ContextType>({
```
matthprost, line 76: You can also externalize `createContext` into an individual file that you could call `StepperProvider.tsx` even if there is not much into it, it will increase readability

matthprost (changes requested):

packages/ui/src/components/Stepper/index.tsx
lisalupi: It is for backward compatibility only :/

matthprost (approved):

JulienSaguez (approved):

PetmePant (approved):

packages/ui/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,247 @@
...
+  transition: text-decoration-color 250ms ease-out;
+  text-decoration-thickness: 1px;
+  text-underline-offset: 2px;
+  text-decoration-color: transparent;
+  white-space: normal;
+`
+
+const StyledStepContainer = styled(Stack)<{
+  isActive: boolean
+  isDone: boolean
```
BABAK0T0, line 80:
```suggestion
  'data-is-active': boolean
  'data-is-done': boolean
```
data-attributes won't re-compute the style and allow us to get static file

packages/ui/src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,247 @@
...
+  'data-disabled': boolean
+  'data-interactive': boolean
+  'data-hide-separator': boolean
+  isAnimated: boolean
```
BABAK0T0, line 84: Same here

PR #3801, opened 2024-05-21
feat(UnitInputField): new component

lisalupi:
## Type
- Feature

#### What is expected?

New component `<UnitInputField />`


matthprost (approved):
Looks good to me, great job ! 🚀

packages/form/src/components/UnitInputField/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    | 'min'
+    | 'options'
+    | 'placeholder'
+    | 'placeholderUnit'
+    | 'size'
+    | 'unitValue'
+    | 'required'
+    | 'width'
+    | 'helper'
+    | 'selectInputWidth'
```
DorianMaliszewski, line 30: I dont like that may be we need another way to handle sizes inside the input ?

What do you think about using a Row and let the user pass a custom templateColumns ? Like this he can manage both side ?

packages/form/src/components/UnitInputField/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    | 'value'
+    | 'max'
+    | 'min'
+    | 'options'
+    | 'placeholder'
+    | 'placeholderUnit'
+    | 'size'
+    | 'unitValue'
+    | 'required'
+    | 'width'
```
DorianMaliszewski, line 28: I dont like this too we should drop this prop

packages/form/src/components/UnitInputField/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    | 'helper'
+    | 'selectInputWidth'
+  > & {
+    onChange?: ComponentProps<typeof UnitInput>['onChange']
+    onChangeUnitValue?: ComponentProps<typeof UnitInput>['onChangeUnitValue']
+    label: string
```
DorianMaliszewski, line 34: You need to add the validate prop as we have in other Fields (like TextInputField)

DorianMaliszewski (approved):
LGTM :rocket:

PR #3802, opened 2024-05-21
feat(tokens): new breakpoints and fallback for each font family

matthprost:
## Type

- Enhancement

#### What is expected?

New tokens `breakpoints` and added a fallback for each font family


packages/themes/src/themes/console/light.ts
```diff
@@ -4,6 +4,14 @@
  * PLEASE DO NOT EDIT HERE
  */
 export const lightTheme = {
+  breakpoints: {
```
philibea, line 7: can we sort ?

matthprost: @philibea why would you want to sort it? 😅

philibea: in size order :)

philibea:
@matthprost you should update snapshots

packages/themes/src/themes/console/light.ts
matthprost: @philibea Im not sure it's useful the tokens are use as object directly, ordering will need to increase the generating script size for not much I think

alexandre-combemorel (approved):

philibea (approved):

PR #3864, opened 2024-06-12
fix: navigation loses focus

lisalupi:
## Type

- Bug

#### What is expected?

Navigation items will not lose focus when space pressed

### DIFF
--- a/packages/plus/src/components/Navigation/components/Item.tsx
+++ b/packages/plus/src/components/Navigation/components/Item.tsx
@@ -49,7 +49,6 @@ const StyledIcon = styled(Icon, {
 
     ${({ active, theme }) => (active ? `background: ${theme.colors.primary.backgroundHover};` : null)}
     }
-  }
 `
 
 const NeutralButtonLink = css`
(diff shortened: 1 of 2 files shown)


packages/plus/src/components/Navigation/components/Item.tsx
```diff
@@ -142,29 +140,29 @@ const StyledContainer = styled(Stack)`
   ${NeutralButtonLink};
   border-radius: ${({ theme }) => theme.radii.default};
 
-  &[data-has-no-expand="false"] {
+  &[data-has-no-expand='false'] {
```
matthprost, line 143: Seems like it changed from `"` to `'`, did you install biome correctly on your IDE and removed prettier?

matthprost (changes requested):

packages/plus/src/components/Navigation/components/Item.tsx
lisalupi: I'll check

johnrazeur (approved):

philibea (approved):

alexandre-combemorel (approved):

PR #3871, opened 2024-06-13
chore: migrate to eslint9

johnrazeur:
## Type

- Feature

#### What is expected?

Migrate to eslint 9.


packages/ui/.jest/a11y.test.tsx
```diff
@@ -1,140 +0,0 @@
-/**
- * TODO/ outdated script with vitest migration
- **/
```
matthprost, line 3: Why deleting it? 😢

matthprost (changes requested):

package.json
```diff
@@ -87,7 +87,7 @@
       }
     }
   },
-  "devDependencies": {
+  "dependencies": {
```
matthprost, line 90: Not sure about this one too hahah

matthprost (changes requested):

philibea (approved):

alexandre-combemorel (approved):

packages/ui/.jest/a11y.test.tsx
philibea: it's not used any more, and we be replace by the next pr of axe

package.json
```diff
@@ -107,7 +107,9 @@
     "@emotion/jest": "11.11.0",
     "@emotion/react": "11.11.4",
     "@emotion/styled": "11.11.5",
-    "@scaleway/eslint-config-react": "3.18.4",
+    "@eslint/compat": "^1.1.0",
```
matthprost, line 110:
```suggestion
    "@eslint/compat": "1.1.0",
```

matthprost (changes requested):

package.json
johnrazeur: We should not have devDependencies in the root package.json

PR #3931, opened 2024-06-28
fix(Illustrations): add doc and update generated files

lisalupi:
## Type

- Bug
- Documentation
- Migration

1. More details on doc

2. Add a header to every generated files in illustrations


philibea (approved):

.changeset/stale-rocks-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/illustrations": patch
+---
+
+Update header generated files and add doc
```
matthprost, line 5: No need for changeset when it's non-functional feature 😄

matthprost (changes requested):

PR #3939, opened 2024-07-01
feat(Illustrations): ipam illustration

lisalupi:
## Type

- Feature

#### What is expected?

New illustration: `ipam`


.changeset/calm-wasps-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/illustrations": patch
```
matthprost, line 2: Could be a minor instead as it's an addition and not a fix 😄

matthprost (changes requested):

matthprost (approved):

vincentaudebert (approved):

PR #3968, opened 2024-07-04
fix(SearchInput): add new custom shortcut and children now optional

matthprost:
## Type

- Enhancement

#### What is expected?

Improve `<SearchInput />`:
  - prop `shortcut` now takes an array of string in addition to boolean. This way you can define multiple shortcuts for the same input: `shortcut={['/', 's']` for example.
  - prop `children` is now optional. If not provided the popup will not be displayed and the input will behave like a regular input.
  - new prop `className`

Screenshot 2024-07-04 at 17 57 48


lisalupi (approved):

philibea (approved):

packages/ui/src/components/SearchInput/Key.tsx
```diff
@@ -22,18 +23,43 @@ const Container = styled.div`
...
+  )
+
+  return (
+    <Container
+      data-disabled={disabled}
+      data-children-length={children.length > 1}
+    >
+      <Text as="span" variant="caption" sentiment="neutral" disabled={disabled}>
+        {isSpecialKey
+          ? KEYS_MATCH[children as keyof typeof KEYS_MATCH]
```
philibea, line 60: do we need to cast here ? can't we have a typeguard function ?

packages/ui/src/components/SearchInput/index.tsx
```diff
@@ -153,32 +157,80 @@ export const SearchInput = forwardRef(
       }
     }
 
-    const isMacOS = navigator.userAgent.includes('Mac')
+    useEffect(() => {
+      if (typeof window !== 'undefined') {
```
philibea, line 161: maybe create a variable isBrowser and used it on all the repository ?

packages/ui/src/components/SearchInput/index.tsx
```diff
@@ -66,13 +67,16 @@ export const SearchInput = forwardRef(
...
       disabled,
+      className,
     }: SearchInputProps,
     ref: Ref<HTMLInputElement>,
   ) => {
     const focusedLinkIndex = useRef(0)
     const popupRef = useRef<HTMLDivElement>(null)
     const [containerWidth, setContainerWidth] = useState(0)
     const [searchTerms, setSearchTerms] = useState('')
+    const [isMacOS, setIsMacOS] = useState(false)
```
philibea, line 78: Do we need a react state for this ?

packages/ui/src/components/SearchInput/Key.tsx
philibea: isSpecialKey is not well typed, that why you have to cast.

```
 const isSpecialKey = useMemo(
    () =>
      Object.keys(KEYS_MATCH).find(
        key => key === children,
      ) as (typeof KEYS_MATCH)[keyof typeof KEYS_MATCH],
    [children],
  )
```

packages/ui/src/components/SearchInput/Key.tsx
```diff
@@ -22,18 +23,43 @@ const Container = styled.div`
   }
 `
 
+export const KEYS_MATCH = {
+  Enter: '↵',
+  ' ': '␣',
```
BABAK0T0, line 28:
```suggestion
  'Space': '␣',
```

packages/ui/src/components/SearchInput/index.tsx
matthprost: @philibea yes we need as we cannot know if `window` exists or not depending on if the code is SSR or CSR

packages/ui/src/components/SearchInput/Key.tsx
matthprost: I did that too first but it's casting also isSpecialKey even though I agree it's cleaner this way

packages/ui/src/components/SearchInput/Key.tsx
matthprost: @BABAK0T0 the key event is a blank space like I added, I'm not checking the code but the key. Check on https://www.toptal.com/developers/keycode and press you space bar and the `event.key` part

Vinzius (approved):

PR #4002, opened 2024-07-11
fix(TagInputField): add regex prop

matthprost:
## Type

- Enhancement

#### What is expected?

Add support of `regex` on `TagInputField` component


radhi-nasser-scaleway (approved):

packages/form/src/components/TagInputField/index.tsx
```diff
@@ -62,7 +65,22 @@ export const TagInputField = <
     rules: {
       required,
       shouldUnregister,
-      validate,
+      validate: {
+        ...(regexes
+          ? {
+              pattern: value =>
+                regexes.every(
```
johnrazeur, line 72: use `validateRegex`

johnrazeur (approved):

packages/form/src/components/TagInputField/index.tsx
matthprost: 🫡

PR #4007, opened 2024-07-12
feat(Modal): introducing nested modals

matthprost:
## Type

- Enhancement

### Introducing: Nested modals ✨ 

Modals were already capable of being nested and without any issue. We still wanted to push it forward and add some nice animation and move back the parent modal when a children one appear.

The update is non-breaking and will apply automatically.

Screen Recording 2024-07-12 at 16 18 35


packages/ui/src/components/Modal/ModalProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const ModalContext = createContext<ModalContextValues | undefined>(
+  undefined,
+)
+
+export const useModal = () => {
+  const context = useContext(ModalContext)
+  if (!context) {
+    throw new Error('useModal must be used within a ModalProvider')
```
philibea, line 24: this one is easy to test :)

packages/ui/src/components/Modal/components/Dialog.tsx
```diff
@@ -209,38 +240,45 @@ export const Dialog = ({
...
-            ref={dialogRef}
-            tabIndex={0}
-          >
-            {open ? children : null}
-          </StyledDialog>
-        </StyledBackdrop>,
-        containerRef.current,
-      )
-    : null
+  const position = [...openedModales].reverse().indexOf(id) // reverse method mutate array so we need to create a new array
```
fabienhebert, line 243: We could use `lastIndexOf` instead of reversing no ?

Otherwhile just loop on every modals is not that costly, why going from the last one ?

fabienhebert: Oh ok you store the last modal as last item but you need to considerer it as position : 0, hmm okay why not then, you can probably ignore last feedback, may be add it could worth to add a comment to explain this part

fabienhebert (approved):

packages/ui/src/components/Modal/components/Dialog.tsx
matthprost: yes that's a bit tricky, the last modal open is first in position as it's size won't change and the first opened one is reversed. I will add a comment to explain better

packages/ui/src/components/Modal/ModalProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return context
+}
+
+type ModalProviderProps = {
+  children: ReactNode
+}
+
+export const ModalProvider = ({ children }: ModalProviderProps) => {
+  const [openedModales, setOpenedModales] = useState<string[]>([])
```
BABAK0T0, line 35:
```suggestion
  const [openedModals, setOpenedModals] = useState<string[]>([])
```

BABAK0T0 (approved):

mhervouet (approved):

PR #4019, opened 2024-07-16
refactor: improve tagslist component

radhi-nasser-scaleway:
## Type

- Enhancement

### Summarise concisely:

Allow TagList component to to not overflow it's parent by ignoring the provided threshold in this case.
When we have only one tag to show, we show and we ellipsis it with a tooltip so the TagList dosen't overflow as well.

image

|   Before   |      After |
| :--------: | ---------: |
| before | after |

### DIFF
--- a/packages/ui/src/components/TagList/index.tsx
+++ b/packages/ui/src/components/TagList/index.tsx
@@ -28,14 +28,14 @@ const TagsWrapper = styled.span`
 `
 
 const StyledTagContainer = styled.div<{
-  elementsGap: string
+  gap: string
   multiline?: boolean
   popoverTriggerWidth?: number
 }>`
   display: flex;
   align-items: center;
   color: ${({ theme }) => theme.colors.neutral.text};
-  gap: ${({ elementsGap }) => elementsGap};
+  gap: ${({ gap }) => gap};
   ${({ multiline }) => multiline && `flex-wrap: wrap;`};
 
   // to handle the case where we have one tag and we need to ellipsis it
@@ -247,7 +247,7 @@ export const TagList = ({
   return (
     <StyledContainer className={className} data-testid={dataTestId}>
       <StyledTagContainer
-        elementsGap={TAGS_GAP}
+        gap={TAGS_GAP}
         multiline={multiline}
         popoverTriggerWidth={popoverTriggerWidth}
         ref={containerRef}
@@ -270,7 +270,7 @@ export const TagList = ({
             whiteSpace: 'nowrap',
           }}
         >
-          <StyledTagContainer elementsGap={TAGS_GAP}>
+          <StyledTagContainer gap={TAGS_GAP}>
             {potentiallyVisibleTags.map((tag, index) => renderTag(tag, index))}
           </StyledTagContainer>
         </div>,
@@ -284,7 +284,7 @@ export const TagList = ({
           onClose={() => setIsPopoverVisible(false)}
           placement={popoverPlacement}
           content={
-            <StyledTagContainer multiline elementsGap={TAGS_GAP}>
+            <StyledTagContainer multiline gap={TAGS_GAP}>
               {hiddenTags.map((tag, index) => renderTag(tag, index))}
             </StyledTagContainer>
           }
--- a/packages/ui/src/components/TagList/index.tsx
+++ b/packages/ui/src/components/TagList/index.tsx
@@ -54,7 +54,7 @@ const StyledTagContainer = styled.div<{
   `};
 `
 
-type TagType =
+export type TagType =
   | string
   | { label: string; icon: NonNullable<ComponentProps<typeof Tag>['icon']> }
 
@@ -111,6 +111,12 @@ export const TagList = ({
   const measureRef = useRef<HTMLDivElement>(null)
   const popoverTriggerRef = useRef<HTMLDivElement>(null)
 
+  // a flag to keep of when we show the component as we we might update the visible tags list
+  // after the first render ( to know if we should add ellipsis to the first tag or readjust
+  // the tags when joined with the popover trigger might overflow the parent ) and this causes
+  // some flickering
+  const [isReady, setIsReady] = useState(false)
+
   const [isPopoverVisible, setIsPopoverVisible] = useState(false)
   const [popoverTriggerWidth, setPopoverTriggerWidth] = useState(0)
   const [visibleTags, setVisibleTags] = useState<TagType[]>([])
@@ -154,11 +160,12 @@ export const TagList = ({
     if (multiline) {
       setVisibleTags(potentiallyVisibleTags)
       setHiddenTags(surelyHiddenTags)
+      setIsReady(true)
 
       return
     }
 
-    const parentWidth = containerRef.current.parentElement?.clientWidth || 0
+    const parentWidth = containerRef.current.parentElement?.offsetWidth || 0
 
     const measureElements: HTMLCollection =
       measureRef.current.children[0].children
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/TagList/index.tsx
```diff
@@ -23,25 +27,46 @@ const TagsWrapper = styled.span`
   padding-right: 8px;
 `
 
-const StyledTagContainer = styled.div<{ multiline?: boolean }>`
+const StyledTagContainer = styled.div<{
+  elementsGap: string
```
BABAK0T0, line 31:
```suggestion
  gap: string
```

packages/ui/src/components/TagList/index.tsx
```diff
@@ -23,25 +27,46 @@ const TagsWrapper = styled.span`
...
+  multiline?: boolean
+  popoverTriggerWidth?: number
+}>`
   display: flex;
   align-items: center;
   color: ${({ theme }) => theme.colors.neutral.text};
-  gap: ${({ theme }) => theme.space['1']};
+  gap: ${({ elementsGap }) => elementsGap};
```
BABAK0T0, line 38:
```suggestion
  gap: ${({ gap }) => gap};
```

packages/ui/src/components/TagList/index.tsx
```diff
@@ -82,77 +107,197 @@ export const TagList = ({
...
-            copiable={copiable}
-            copyText={copyText}
-            copiedText={copiedText}
-            icon={typeof tag === 'object' ? tag.icon : undefined}
-          >
-            {getTagLabel(tag)}
-          </Tag>
-        ))}
+      <StyledTagContainer
+        elementsGap={TAGS_GAP}
```
BABAK0T0, line 250:
```suggestion
        gap={TAGS_GAP}
```

packages/ui/src/components/TagList/index.tsx
radhi-nasser-scaleway: https://github.com/scaleway/ultraviolet/pull/4019/commits/02695659507e017b35637af17440ae513d72593a

packages/ui/src/components/TagList/index.tsx
radhi-nasser-scaleway: https://github.com/scaleway/ultraviolet/pull/4019/commits/02695659507e017b35637af17440ae513d72593a

packages/ui/src/components/TagList/index.tsx
```diff
@@ -82,77 +107,197 @@ export const TagList = ({
...
+    for (let i = 0; i < measureElements.length; i += 1) {
+      accumulatedWidth +=
+        (measureElements[i] as HTMLDivElement).offsetWidth +
+        parseInt(TAGS_GAP, 10)
+      if (accumulatedWidth <= parentWidth) {
+        measuredVisibleTags.push(tags[i])
+      } else {
+        measuredHiddenTags.push(tags[i])
+      }
+    }
```
matthprost, line 176: We tend to avoid `for` / `let` and using `.push()` on arrays due to mutation. Instead you could do it with a reduce:

```tsx
const { measuredVisibleTags, measuredHiddenTags, accumulatedWidth } = measureElements.reduce((acc, element, index) => {
  const newAccumulatedWidth = acc.accumulatedWidth + (measureElements[i] as HTMLDivElement).offsetWidth +
        parseInt(TAGS_GAP, 10)
  
  return 
     ({ 
       measuredVisibleTags: [...(acc.measuredVisibleTags), accumulatedWidth <= parentWidth && tags[i]], 
       measuredHiddenTags: [...(acc.measuredHiddenTags), accumulatedWidth > parentWidth && tags[i]], 
       accumulatedWidth: newAccumulatedWidth 
      })

}, {measuredVisibleTags: [], measuredHiddenTags: [], accumulatedWidth: 0})
```

packages/ui/src/components/TagList/index.tsx
```diff
@@ -82,77 +107,197 @@ export const TagList = ({
...
+        setPopoverTriggerWidth(newPopoverTriggerWidth)
+      }
+
+      // add a first tag with ellipsis when the first tag does not fit in the parent container
+      if (visibleTags.length === 0 && hiddenTags.length > 0) {
+        const visibleTagsCopy = [...visibleTags]
+        const hiddenTagsCopy = [...hiddenTags]
+
+        const tagToMove = hiddenTagsCopy.pop()
+        visibleTagsCopy.push(tagToMove as TagType)
```
matthprost, line 200: Same here we should avoid mutation as much as possible, you will have to create a new array:

```tsx
const newArray = [...visibleTagsCopy, tagToMove]
```

packages/ui/src/components/TagList/index.tsx
```diff
@@ -82,77 +107,197 @@ export const TagList = ({
...
+      const parentWidth = tagsContainer?.parentElement?.clientWidth || 0
+
+      if (
+        hiddenTags.length > 0 &&
+        tagsContainerWidth + newPopoverTriggerWidth > parentWidth
+      ) {
+        const visibleTagsCopy = [...visibleTags]
+        const hiddenTagsCopy = [...hiddenTags]
+
+        const tagToMove = visibleTagsCopy.pop()
```
matthprost, line 220:
```suggestion
        const tagToMove = visibleTagsCopy.filter((_ index) => index === visibleTagsCopy.length - 1)
```

packages/ui/src/components/TagList/index.tsx
```diff
@@ -82,77 +107,197 @@ export const TagList = ({
...
+        hiddenTagsCopy.unshift(tagToMove as TagType)
```
matthprost, line 221:
```tsx
const newArray = [tagToMove, ...hiddenTagsCopy]
```

matthprost (changes requested):

matthprost (changes requested):

packages/ui/src/components/TagList/index.tsx
```diff
@@ -82,77 +107,253 @@ export const TagList = ({
...
+            position: 'absolute',
+            whiteSpace: 'nowrap',
+          }}
+        >
+          <StyledTagContainer gap={TAGS_GAP}>
+            {potentiallyVisibleTags.map((tag, index) => renderTag(tag, index))}
+          </StyledTagContainer>
+        </div>,
+        document.body,
+      )}
```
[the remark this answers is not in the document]
radhi-nasser-scaleway: There's no other method ( as far as I know )to compute the rendered element width, I tried doing it this was at first:
```
const tagElement = document.createElement('div');
tagElement.className = 'tag';
offscreenContainer.appendChild(tagElement);
ReactDOM.render(<Tag />, tagElement);
```
But I had issues with providers not being there, such as the `ThemeProvider` as it used in the tag to get spacing, color and stuff, so ended up doing the current solution

And nice catch for the leftover in the dom, it is being cleaned now

packages/ui/src/components/TagList/index.tsx
```diff
@@ -23,25 +27,46 @@ const TagsWrapper = styled.span`
...
+  ${({ popoverTriggerWidth }) =>
+    popoverTriggerWidth &&
+    `
+      &:has(.ellipsed) {
+        width: calc(100% - ${popoverTriggerWidth}px); // to let space for the +X button
+        max-width: fit-content;
+      }
+    
+      & * {
+        width: 100% !important;
```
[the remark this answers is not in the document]
radhi-nasser-scaleway: the `!important` was a leftover, I got rid of it

as for the `& > *`, I want to target the div below each tag and the span below that as well, I tried as you suggested but I it's requiring some props I don't have in the scope of my component
image
 so I've changed to target only the div and span I wanted in the first place `& span, div`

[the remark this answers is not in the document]
matthprost: oh okay then go with the `& span, div` it should be fine

packages/ui/src/components/TagList/index.tsx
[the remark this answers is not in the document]
matthprost: okay no problem, most important is to keep the dom clean and it's nice to see what other path you tried 🙏

matthprost (approved):

JulienSaguez (approved):

PR #4026, opened 2024-07-16
fix: readOnly case

fabienhebert:
## Type

- Bug

### Summarise concisely:

`KeyValueField` :  prop `readonly` becomes `readOnly` to fit React standard
`TextArea`: `data-readOnly` becomes `data-readonly` to fit standard

### DIFF
--- a/.changeset/young-berries-dress.md
+++ b/.changeset/young-berries-dress.md
@@ -3,4 +3,4 @@
 "@ultraviolet/ui": patch
 ---
 
-`readonly` props becomes `readOnly` and fix `data-readonly` case
+`KeyValueField` : `readonly` props becomes `readOnly`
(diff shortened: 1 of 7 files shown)


vincentaudebert (approved):

.changeset/young-berries-dress.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ultraviolet/form": patch
+"@ultraviolet/ui": patch
+---
+
+`readonly` props becomes `readOnly` and fix `data-readonly` case
```
matthprost, line 6: Maybe write on which component api changed so it's easier to migrate

matthprost (changes requested):

matthprost (approved):

PR #4032, opened 2024-07-17
refactor(SliderField): match possible values with slider value

matthprost:
## Type

- Refactor

#### What is expected?

Refactoring on `<Slider />`: `possibleValues` no longer exists and has been regroupped under `options` prop.

`<SliderField />` : The component will automatically set the value of each options in the form values so you don't have to do the matching on submit.

https://github.com/user-attachments/assets/5057ff31-87bb-4a4f-b952-3ffaeb74fc2e


DorianMaliszewski (approved):
very nice job

alexandre-combemorel (approved):

PR #4077, opened 2024-07-25
feat: add title tag to SVG for accebility purpose

radhi-nasser-scaleway:
## Type

- Enhancement

#### What is expected?

`Icon`s have now an optional title prop, than can be used to add/update a `Title` tag to be ridden by accessibility tools.

Added an optional prop `title` to `Icon`

### DIFF
--- a/packages/ui/src/components/Button/index.tsx
+++ b/packages/ui/src/components/Button/index.tsx
@@ -251,7 +251,6 @@ type CommonProps = {
   disabled?: boolean
   iconPosition?: 'left' | 'right'
   iconVariant?: ComponentProps<typeof Icon>['variant']
-  iconTitle?: string
   fullWidth?: boolean
   isLoading?: boolean
   'aria-label'?: string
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Button/index.tsx
```diff
@@ -251,6 +251,7 @@ type CommonProps = {
   disabled?: boolean
   iconPosition?: 'left' | 'right'
   iconVariant?: ComponentProps<typeof Icon>['variant']
+  iconTitle?: string
```
matthprost, line 254: I think aria-label is enough already for having a button with icon only

matthprost (changes requested):

packages/ui/src/components/Button/index.tsx
radhi-nasser-scaleway: removed until I get a reply form numa https://github.com/scaleway/ultraviolet/pull/4077/commits/d7db7f3173f342dd6bf57bff21c80c0cefba6c22

PR #4109, opened 2024-08-05
feat(Slider): component can be controlled or not

DorianMaliszewski:
## Type

- Enhancement

#### What is expected?

Slider can be controlled or not

1. Make the component working controlled or not

2.

3.


packages/ui/src/components/Slider/components/SingleSlider.tsx
```diff
@@ -124,24 +120,21 @@ export const SingleSlider = ({
...
 
-  const onChangeCallback = useCallback(
-    (newValue: number) => {
-      setValueToShow(options ? options[newValue].value : newValue)
-      setComputedValue(newValue ?? min)
-      onChange?.(newValue ?? min)
-    },
-    [min, onChange, options],
-  )
+  const internalOnChangeRef = useRef((newValue: number) => {
```
philibea, line 123: Do we need the useRef ?

packages/ui/src/components/Slider/components/SingleSlider.tsx
```diff
@@ -124,24 +120,21 @@ export const SingleSlider = ({
...
+    setSelectedIndex(newValue ?? min)
+    onChange?.(newValue ?? min)
+  })
 
   // Make sure that min <= value <= max
   useEffect(() => {
     if (value < min) {
```
philibea, line 130: can we create a function for this one ? i think it might be used somewhere else ?

packages/ui/src/components/Slider/components/SingleSlider.tsx
DorianMaliszewski: Yes because we use it in useEffect and eslint rule doesnt allow us to omit it in deps array

packages/ui/src/components/Slider/components/SingleSlider.tsx
DorianMaliszewski: We dont use this anywhere else, its just to ensure that the component is working when passing a bad value

packages/ui/src/components/Slider/__stories__/Options.stories.tsx
```diff
@@ -18,6 +18,8 @@ export const Options: StoryFn<typeof Slider> = args => {
   const [value, setValue] = useState(3)
   const [doubleValue, setDoubleValue] = useState([1, 3])
 
+  console.log(value)
```
Lawndlwd, line 21: oups

Lawndlwd (approved):

ValentinRLL (approved):

philibea (approved):

PR #4173, opened 2024-08-29
fix(EstimateCost): item to have better responsiveness

matthprost:
## Type

- Enhancement

#### What is expected?

Fix `<EstimateCost.Item />` to have better responsiveness when using `shouldBeHidden` prop


DorianMaliszewski (approved):

packages/plus/src/components/EstimateCost/__tests__/__snapshots__/Regular.test.tsx.snap
```diff
@@ -6023,7 +6023,7 @@ exports[`EstimateCost - Regular Item > render basic props with overlay 1`] = `
   border: 0;
 }
 
-@media (max-width: 1420px) {
+@media (max-width: 1600px) {
```
philibea, line 6026: should we use our breakpoint ?

radhi-nasser-scaleway (approved):

packages/plus/src/components/EstimateCost/__tests__/__snapshots__/Regular.test.tsx.snap
matthprost: Done ✅

PR #4195, opened 2024-09-04
fix: onChange type and value for some fields

johnrazeur:
## Type

- Bug

Fix some onChange type and value.


matthprost (approved):
❤️

PR #4201, opened 2024-09-05
fix(SelectInputV2): search by label

lisalupi:
## Type

- Enhancement

#### What is expected?

- prop `searchText` is now mandatory if the value of an option is not a string
- search will look for matches in the description if it is of type string

1. Search by `searchText`. Search on label only if it is a string. Search is not on value anymore.

2. When a string description is provided, search for matches in it

3. Stories fixed
4. Tests


packages/ui/src/components/SelectInputV2/SearchBarDropdown.tsx
```diff
@@ -78,20 +78,24 @@ export const SearchBarDropdown = ({
...
-          filteredOptions[group] = filteredOptions[group].filter(option =>
-            option.searchText
-              ? option.searchText.match(regex)
-              : option.value.match(regex),
+          filteredOptions[group] = filteredOptions[group].filter(
+            option =>
+              (option.searchText && !!option.searchText.match(regex)) ||
+              (typeof option.label === 'string' && option.label.match(regex)) ||
+              (typeof option.description === 'string' &&
+                option.description.match(regex)),
```
matthprost, line 86: I think it doesn't hurt to search into the value in addition of the rest 😄 

Screenshot 2024-09-06 at 10 32 13

https://github.com/user-attachments/assets/60991418-d3ab-4356-be4f-f531aab78fa7

matthprost (changes requested):

matthprost (approved):

packages/ui/src/components/SelectInputV2/SearchBarDropdown.tsx
```diff
@@ -78,20 +78,26 @@ export const SearchBarDropdown = ({
       if (!Array.isArray(options)) {
         const filteredOptions = { ...options }
         Object.keys(filteredOptions).map((group: string) => {
-          filteredOptions[group] = filteredOptions[group].filter(option =>
-            option.searchText
-              ? option.searchText.match(regex)
-              : option.value.match(regex),
+          filteredOptions[group] = filteredOptions[group].filter(
```
johnrazeur, line 81: Can we move this inside a function ?

johnrazeur (approved):

PR #4214, opened 2024-09-10
feat(selectableCard): illustrations and icons

lisalupi:
## Type

- Feature
#### What is expected?

New props `productIcon` & `illustration`: add an icon/illustration on the right of the card 

Capture d’écran 2024-09-10 à 11 20 56
Capture d’écran 2024-09-10 à 11 20 52


packages/ui/src/components/SelectableCard/index.tsx
```diff
@@ -132,6 +165,11 @@ export type SelectableCardProps = {
   tooltip?: string
   label?: ReactNode
   'data-testid'?: string
+  productIcon?: PascalToCamelCaseWithoutSuffix<
```
philibea, line 168: you can't have both props in the same time right ?

packages/ui/src/components/SelectableCard/__tests__/index.test.tsx
```diff
@@ -4,6 +4,9 @@ import { renderWithTheme, shouldMatchEmotionSnapshot } from '@utils/test'
 import { describe, test, vi } from 'vitest'
 import { SelectableCard } from '..'
 
+const illustration =
+  'https://ultraviolet.s3.fr-par.scw.cloud/illustrations/products/appleSilicon/apple-silicon-wire.svg'
+
```
matthprost, line 9: this might be risky because the link might die someday and we won't understand why the test fail. Maybe import locally a svg that you will use in your test instead

packages/ui/src/components/SelectableCard/index.tsx
lisalupi: you can, but you shouldn't

packages/ui/src/components/SelectableCard/index.tsx
```diff
@@ -205,6 +285,8 @@ export const SelectableCard = forwardRef(
           data-testid={dataTestId}
           data-type={type}
           data-has-label={!!label}
+          data-has-illustration={!!illustration}
+          data-has-icon={!!productIcon && !illustration}
```
matthprost, line 289: you could merge those in one data, like `data-image="illustration | product"`

matthprost (approved):

matthprost (changes requested):
miss approved haha

packages/ui/src/components/SelectableCard/index.tsx
lisalupi: styling is a little bit different for illustrations and icons

packages/ui/src/components/SelectableCard/index.tsx
matthprost: you can enforce the type by saying it's one or another

BABAK0T0 (approved):

packages/ui/src/components/SelectableCard/index.tsx
matthprost: yes exactly so you could do like you did but merging both, something like:

```css
  &[data-image="illustration"] {
    padding: ${({ theme }) => theme.space[0]};
  }
  &[data-image="product"] {
    padding: ${({ theme }) => theme.space[0]};
    padding-right: ${({ theme }) => theme.space['2']};
  }
```

BABAK0T0 (approved):

matthprost (approved):

PR #4226, opened 2024-09-12
feat(ContentCard): content card custom heading tag

lisalupi:
## Type

- Enhancement

### Summarise concisely:
new prop `headingTag` to change tag of header

### DIFF
--- a/packages/plus/src/components/ContentCard/index.tsx
+++ b/packages/plus/src/components/ContentCard/index.tsx
@@ -3,7 +3,7 @@ import { css } from '@emotion/react'
 import styled from '@emotion/styled'
 import { Icon } from '@ultraviolet/icons/legacy'
 import { Stack, Text } from '@ultraviolet/ui'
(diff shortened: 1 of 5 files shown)


packages/plus/src/components/ContentCard/index.tsx
```diff
@@ -111,6 +111,11 @@ type ContentCardProps = {
   icon?: ReactNode
   subtitle?: string
   title: string
+  /*
+   * Define the HTML Section Heading element level (h1 - h6) to use for the header
+   */
+  headingTag?: ElementType
```
matthprost, line 117: I think we can be more restrictive on that enforcing only heading elements (h1, h2, h3, h4, h5, h6)

matthprost (changes requested):

matthprost (approved):

radhi-nasser-scaleway (approved):

PR #4230, opened 2024-09-12
feat(themes): optimization of bundle size barrel file

matthprost:
## Type

- Enhancement

## Summary

`@ultraviolet/themes` currently export all themes in a barrel file which can cause the bundle to be oversized while not all themes are imported.

With this update we separate exports and only include the imported ones in the bundle which reduce drastically the usage for that specific package.

### Before

In a simple project started with create-react-app and having this in the `App.tsx`:

```typescript
import "./styles.css"
import { ThemeProvider } from "@emotion/react"
import { consoleLightTheme } from "@ultraviolet/themes"

export default function App() {
  return (
    <ThemeProvider theme={consoleLightTheme}>
      <div className="App">
        <h1>Hello CodeSandbox</h1>
        <h2>Start editing to see some magic happen!</h2>
      </div>
    </ThemeProvider>
  )
}
```

The bundle size was `205.68KB`:
Screenshot 2024-09-12 at 14 53 31

We can see that dark, light and darker theme are included and all deprecated in the bundle but we only use `consoleLightTheme` in the code of the app.

### After 

In the same project with the same typescript in `App.tsx`, only change is the package of @ultraviolet/themes being updated to this PR changes we obtain a bundle of `169.45KB`:

Screenshot 2024-09-12 at 14 54 20

We see that only light theme is included and not the others.

### Conclusion

This update provide change the bundle size from `205.68KB` to `169.45KB` for a simple react app. 
A reduction of **17.61%** 🚀


packages/themes/src/themes/console/dark/index.ts
```diff
@@ -0,0 +1,8 @@
+import deepmerge from 'deepmerge'
+import { deprecatedDarkTokens } from '../deprecated/dark'
+import { darkTheme } from './__generated__'
+
+export const consoleDarkTheme = deepmerge(
+  darkTheme,
+  deprecatedDarkTokens,
+) as typeof darkTheme & typeof deprecatedDarkTokens
```
alexandre-combemorel, line 8: Would it work to just type: `export const consoleDarkTheme: typeof darkTheme & typeof deprecatedDarkTokens` ?

alexandre-combemorel (approved):

alexandre-combemorel:
Just to understand, we still import this way: `import { consoleLightTheme } from "@ultraviolet/themes"` 
So we rely on tree shaking to get rid of the other export within `themes/index.ts`, why not doing 
`import { consoleLightTheme } from "@ultraviolet/themes/dark"` directly?

fabienhebert (approved):

alexandre-combemorel (approved):

packages/themes/package.json
```diff
@@ -46,6 +46,11 @@
       "types": "./dist/index.d.ts",
       "require": "./dist/index.cjs",
       "default": "./dist/index.js"
+    },
+    "./console/*": {
```
philibea, line 50: there is a mistake here, missing a wildcard also inside types,require, default

packages/themes/package.json
```diff
@@ -46,6 +46,11 @@
...
+      "types": "./dist/themes/console/index.d.ts",
+      "require": "./dist/themes/console/index.cjs",
+      "default": "./dist/themes/console/index.js"
```
philibea, line 53: "types": "./dist/themes/console/*/index.d.ts",
      "require": "./dist/themes/console/*/index.cjs",
      "default": "./dist/themes/console/*/index.js"

PR #4249, opened 2024-09-17
feat(Snippet): rows and scroll

lisalupi:
## Type

- Enhancement

### Summarise concisely:
new props `rows` (number) to define the max number of rows to show and `noExpandable` (boolean) to control show/hide option


packages/ui/src/components/Snippet/index.tsx
```diff
@@ -8,16 +8,25 @@ import { Stack } from '../Stack'
 import { Text } from '../Text'
 
 const LINES_BREAK_REGEX = /\r\n|\r|\n/
+const LINE_HEIGHT = 20
```
matthprost, line 11: might be better to get that value from the theme instead of setting a global fixed value

packages/ui/src/components/Snippet/index.tsx
```diff
@@ -212,13 +239,26 @@ export const Snippet = ({
     >
       <StyledStack>
         {hasShowMoreButton ? (
-          <Expandable minHeight={120} opened={showMore}>
-            <CodeContent prefix={prefix} multiline={multiline} lines={lines}>
+          // Height = number of rows * height of a line + padding
+          <Expandable minHeight={rows * LINE_HEIGHT + 32} opened={showMore}>
```
matthprost, line 243: Same for the 32 value, hardcoded is no good, you can get this from theme

packages/ui/src/components/Snippet/index.tsx
```diff
@@ -212,13 +239,26 @@ export const Snippet = ({
...
           </Expandable>
         ) : (
-          <CodeContent prefix={prefix} multiline={multiline} lines={lines}>
+          <CodeContent
+            prefix={prefix}
+            multiline={multiline}
+            lines={lines}
+            noExpandable={noExpandable}
+            rows={rows}
+          >
```
matthprost, line 261: There is a small issue with the copy button container, it seems that we applied a background (I don't know why) so when the scrollbar appear it goes under. Can you remove the ButtonContainer background so it doesn't do that?

https://github.com/user-attachments/assets/4ecdd660-4dd2-4a7d-9bbf-df9271c3fee9

matthprost (changes requested):

packages/ui/src/components/Snippet/index.tsx
lisalupi: The background is here to avoid overlap between the snippet text and the copy button for long single lines (overflow-x)

matthprost (approved):

vincentaudebert (approved):

mhervouet (approved):

PR #4252, opened 2024-09-17
fix(Modal): minor changes

matthprost:
## Type

- Bug

#### What is expected?

Fix `<Modal />` component to always display the one behind and fix flacky animation

Before:

https://github.com/user-attachments/assets/de371373-9a68-47d5-b6fd-46427ad9a826

After:

https://github.com/user-attachments/assets/0adbd8cc-eb3d-4627-bc2a-7fbfd433bc06

### DIFF
--- a/packages/ui/src/components/Modal/components/Dialog.tsx
+++ b/packages/ui/src/components/Modal/components/Dialog.tsx
@@ -5,7 +5,7 @@ import type {
   MouseEventHandler,
   ReactEventHandler,
 } from 'react'
-import { useCallback, useEffect, useRef } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 import { createPortal } from 'react-dom'
 import { slideFromBottom } from '../../../utils/animations'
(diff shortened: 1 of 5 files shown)


lisalupi (approved):

packages/ui/src/components/Modal/components/Dialog.tsx
```diff
@@ -111,9 +123,13 @@ export const Dialog = ({
     previsousOpenedModales,
   } = useModal()
 
+  useEffect(() => {
+    setIsVisible(true)
```
philibea, line 127: this almost like putting the initial value to isVisible to true, so why using a useEffect for this ?

matthprost: There is slight difference, this useEffect content will be run only after the first render. It is important in that case because we need to run an animation and if the first render didn't occur there will be a flickering in the animation as the position are being sets.

alexandre-combemorel (approved):

PR #4262, opened 2024-09-19
feat(Row): change padding on Row and List.Row

lisalupi:
## Type

- Feature

#### What is expected?
- `<Row />`  New prop `padding` to set a custom padding
- `<List.Row />` : New prop `paddingExpandable` to set a custom padding inside the expandable content

### DIFF
--- a/packages/ui/src/components/List/Row.tsx
+++ b/packages/ui/src/components/List/Row.tsx
@@ -1,5 +1,5 @@
 import styled from '@emotion/styled'
-import type { ForwardedRef, ReactNode } from 'react'
+import type { CSSProperties, ForwardedRef, ReactNode } from 'react'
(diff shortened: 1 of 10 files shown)


packages/ui/src/components/List/Row.tsx
```diff
@@ -106,6 +111,7 @@ export const Row = forwardRef(
       sentiment = 'neutral',
       expanded,
       className,
+      paddingExpandable,
```
matthprost, line 114:
```suggestion
      expandablePadding,
```

matthprost (changes requested):

matthprost (approved):

alexandre-combemorel (approved):

PR #4282, opened 2024-09-26
fix(SelectInputV2): auto scroll

lisalupi:
## Type

- Bug

#### What is expected?

When the dropdown is at the bottom of the screen, there is be an autoscroll to see it completely. 
Works inside a modal too.


packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -674,11 +683,39 @@ export const Dropdown = ({
     options,
     displayedOptions,
   } = useSelectInput()
+  const theme = useTheme()
   const [searchBarActive, setSearchBarActive] = useState(false)
   const [defaultSearchValue, setDefaultSearch] = useState<string | null>(null)
   const ref = useRef<HTMLDivElement>(null)
   const [search, setSearch] = useState<string>('')
   const [maxWidth, setWidth] = useState(refSelect.current?.offsetWidth)
+  const [marginBottom, setMarginBottom] = useState(false)
```
matthprost, line 692: Why don't we just always set the margin bottom? If the dropdown didn't reach the bottom of screen it will be an invisible margin no?

matthprost (changes requested):

matthprost (approved):

alexandre-combemorel (approved):

radhi-nasser-scaleway (approved):

PR #4307, opened 2024-10-03
feat(Chip): new component

lisalupi:
## Type
- Feature

#### What is expected?

New component : `chip`

1. Created a new component

2. Added stories

3. Added tests


mhervouet:
image
Spacing need to be fixed: 

Large size:
- 8px when there's a trailing icon (eg: clear)

Medium size:
- 12px spacing by default
- 8px when there's a trailing icon (eg: clear)

mhervouet:
Need to remove the "dropdown" example also it's no no longer relevant.

alexandre-combemorel (approved):

packages/ui/src/components/Chip/ChipIcon.tsx
```diff
@@ -0,0 +1,73 @@
...
+  onClick?: () => void
+  'data-testid'?: string
+}
+
+export const ChipIcon = ({
+  name,
+  onClick,
+  'data-testid': dataTestId,
+}: ChipIconType) => {
+  const { isActive, disabled } = useContext(ChipContext)
```
matthprost, line 43: Here you need to check if context doesn't exists it means the component has been used outside of a Chip component

packages/ui/src/components/Chip/ChipIcon.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  return (
+    <Container
+      onClick={event => {
+        if (!disabled) {
+          event.stopPropagation()
+          onClick?.()
+        }
+      }}
+      type="button"
```
matthprost, line 59: not sure for this, some icons will be just as illustrative and others will be some buttons. In your example you can see that if I click on filter icon (which is illustrative) it doesn't trigger the Chip while it should:

https://github.com/user-attachments/assets/1d7d009a-720f-47d7-942b-b14e0d8ef43b

packages/ui/src/components/Chip/ChipIcon.tsx
```diff
@@ -0,0 +1,73 @@
...
+import * as Icon from '@ultraviolet/icons'
+import { useContext } from 'react'
+import type { PascalToCamelCaseWithoutSuffix } from 'src/types'
+import { ChipContext } from './ChipContext'
+
+const Container = styled.button`
+background: none;
+border:none;
+cursor: pointer;
+padding:0;
```
matthprost, line 11:
```suggestion
border: none;
cursor: pointer;
padding: 0;
```

matthprost (changes requested):

packages/ui/src/components/Chip/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+  component: Chip,
+  decorators: [
+    StoryComponent => (
+      <div style={{ display: 'flex', flexDirection: 'column' }}>
+        <StoryComponent />
+      </div>
+    ),
+  ],
+  title: 'Components/Badges/Chip',
+} as Meta<typeof Chip>
```
matthprost, line 14: you miss Chip.Icon here in the parameters so we can see the prop of that component in storybook

packages/ui/src/components/Chip/ChipIcon.tsx
```diff
@@ -0,0 +1,84 @@
...
+`
+
+type IconType = PascalToCamelCaseWithoutSuffix<keyof typeof Icon, 'Icon'>
+
+type ChipIconType = {
+  /**
+   * Add an icon in the chip
+   */
+  name: IconType
+  type: 'trailing' | 'leading'
```
matthprost, line 38: Seems like you use this to specify if the icon is on the left or right of the content. I think you can remove this and manage the spacing in the Chip parent directly using stack and gap. wdyt?

packages/ui/src/components/Chip/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+    font-family: ${theme.typography.bodySmall.fontFamily};
+    font-weight: ${theme.typography.bodySmall.weight};
+    letter-spacing: ${theme.typography.bodySmall.letterSpacing};
+    line-height: ${theme.typography.bodySmall.lineHeight};
+    text-transform: ${theme.typography.bodySmall.textCase};
+    text-decoration: ${theme.typography.bodySmall.textDecoration};
+    height: ${theme.space[4]};
+    padding: ${theme.space['0.5']} ${theme.space['2']};`}
+
+  }
```
matthprost, line 77: You don't need to redo all of this, instead wrap the children into a `Text` component and set the `as` prop to `text` if `typeof children === string` and to div if not. This will apply the same properties. Don't forget to set the variant depending on the size.

packages/ui/src/components/Chip/ChipIcon.tsx
```diff
@@ -0,0 +1,84 @@
...
+import * as Icon from '@ultraviolet/icons'
+import { useContext } from 'react'
+import type { PascalToCamelCaseWithoutSuffix } from 'src/types'
+import { ChipContext } from './ChipContext'
+
+const Container = styled.button`
+  background: none;
+  border:none;
+  cursor: pointer;
+  padding:0;
```
matthprost, line 11:
```suggestion
  background: none;
  border: none;
  cursor: pointer;
  padding: 0;
```

packages/ui/src/components/Chip/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export { Playground } from './Playground.stories'
+export { Icons } from './Icons.stories'
+export { Groups } from './Groups.stories.tsx'
+export { Size } from './Size.stories'
```
matthprost, line 19: Would be nice also to have an example with disabled state. With and without the Chip.Icon having a onclick on it so we can check if it works normally on this state.

packages/ui/src/components/Chip/ChipIcon.tsx
```diff
@@ -0,0 +1,84 @@
...
+  name,
+  type,
+  onClick,
+  'data-testid': dataTestId,
+}: ChipIconType) => {
+  const { isActive, disabled, chipContext } = useContext(ChipContext)
+
+  if (!chipContext) {
+    throw new Error('Chip.Icon can only be used inside a Chip component.')
+  }
```
matthprost, line 53: I think you miss understood what I meant haha.
You don't need to create `chipContext` in a context to check if the context exists. You can simply check directly if the variable returned by `useContext()` actually exists. We have plenty of examples like here: https://github.com/scaleway/ultraviolet/blob/7a47067b94292fa23be258a85a3b48cca7c9cdba/packages/ui/src/components/Modal/ModalProvider.tsx#L27

You could first check the context variable and the destructure it if it exists.

matthprost (changes requested):

matthprost (changes requested):

matthprost (approved):

PR #4330, opened 2024-10-09
fix(SelectInputV2): group shown when empty

matthprost:
## Type

- Bug

#### What is expected?

SelectInputV2 regression: the group name is shown while its content is empty. It shouldn't.

| url  | Screenshot 2024-10-09 at 14 30 26 | Screenshot 2024-10-09 at 14 30 34 |

### DIFF
--- a/packages/ui/src/components/SelectInputV2/Dropdown.tsx
+++ b/packages/ui/src/components/SelectInputV2/Dropdown.tsx
@@ -441,8 +441,7 @@ const CreateDropdown = ({
           ) : null}
           {Object.keys(displayedOptions).map((group, index) => (
             <Stack key={group} gap={0.25}>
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -441,8 +441,7 @@ const CreateDropdown = ({
           ) : null}
           {Object.keys(displayedOptions).map((group, index) => (
             <Stack key={group} gap={0.25}>
-              {displayedOptions[group].length > 0 ||
```
philibea, line 444: can we add a test unit on this type of issue ?

lisalupi (approved):

PR #4334, opened 2024-10-09
fix: multiple input component that labels are not clickable

matthprost:
## Type

- Bug

#### What is expected?

Radio, Checkbox and Toggle are not working properly. Their labels are not clickable.

### DIFF
--- a/packages/ui/src/components/Radio/index.tsx
+++ b/packages/ui/src/components/Radio/index.tsx
@@ -137,6 +137,12 @@ const RadioContainer = styled.div`
 
 const StyledLabel = styled.label`
   flex: 1;
+  cursor: pointer;
+`
+
+const StyledTextLabel = styled(Text)`
+  flex: 1;
+  cursor: pointer;
 `
 
 const MargedText = styled(Text)`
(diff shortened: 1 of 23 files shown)


packages/ui/src/components/Radio/index.tsx
```diff
@@ -236,16 +242,23 @@ export const Radio = forwardRef(
...
+              <>
                 {typeof label === 'string' ? (
-                  <Text as="label" variant="body" prominence="default">
+                  <StyledTextLabel
+                    as="label"
+                    variant="body"
+                    prominence="default"
+                    htmlFor={`${computedName}-${value}`}
+                  >
                     {' '}
```
philibea, line 253: `&nbsp;`

packages/ui/src/components/Radio/index.tsx
```diff
@@ -236,16 +242,23 @@ export const Radio = forwardRef(
...
             {label ? (
-              <StyledLabel htmlFor={`${computedName}-${value}`}>
+              <>
                 {typeof label === 'string' ? (
-                  <Text as="label" variant="body" prominence="default">
+                  <StyledTextLabel
+                    as="label"
+                    variant="body"
+                    prominence="default"
+                    htmlFor={`${computedName}-${value}`}
```
philibea, line 251: useId ?

philibea (approved):

lisalupi (approved):

packages/ui/src/components/Radio/index.tsx
matthprost: there is a risks to break other stuff I think, computed name uses `useId()`

DorianMaliszewski (approved):

PR #4339, opened 2024-10-10
feat(SelectInputV2): various improvements

lisalupi:
## Type

- Enhancement

#### What is expected?

`<SelectInputV2 />`:
- Hide search when less than 6 items (even when `searchable = true`)
- Clickable `footer` to close the dropdown
- `clearable = false` by default now

1. Changed default value of prop clearable
2. Changed type of prop footer
3. Manage case selectInput in a modal
4. Updated stories
5. Added test for footer


packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -220,14 +220,23 @@ const handleClickOutside = (
...
+    if (isInsideModal) {
+      // Timeout when the selectInput is inside a modal to not close both the modal and the dropdown if it scrolls
+      setTimeout(() => {
+        setIsDropdownVisibile(false) // hide dropdown when clicking outside of the dropdown
+        onSearch(options) // reset displayed options to default when dropdown is hidden
+      }, 100)
+    } else {
+      setIsDropdownVisibile(false)
+      onSearch(options)
+    }
```
matthprost, line 239: haha no this we absolutely don't want. The issue with this is probably due to the modal itself more than the select input.

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -220,14 +220,23 @@ const handleClickOutside = (
   refSelect: RefObject<HTMLDivElement>,
   onSearch: Dispatch<SetStateAction<DataType>>,
   options: DataType,
+  isInsideModal: boolean,
```
matthprost, line 223: Same here this prop should never exists

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -738,16 +748,33 @@ export const Dropdown = ({
...
+          event,
+          ref,
+          setIsDropdownVisible,
+          refSelect,
+          onSearch,
+          options,
+          false,
+        ),
+      )
+    }
```
matthprost, line 777: Same here the select input should have no knowledge about the modal. This is an issue that the modal has more than select input

packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -764,16 +791,32 @@ export const Dropdown = ({
...
-        handleClickOutside(
-          event,
-          ref,
-          setIsDropdownVisible,
-          refSelect,
-          onSearch,
-          options,
-        ),
-      )
+      if (!modalElement) {
```
matthprost, line 794: Same

packages/ui/src/components/SelectInputV2/__stories__/EmptyState.stories.tsx
```diff
@@ -37,7 +41,7 @@ export const EmptyState: StoryFn<typeof SelectInputV2> = args => (
         <EmptyStateComponent
           bordered
```
matthprost, line 42: Can you remove the bordered in the example ?

packages/ui/src/components/SelectInputV2/__stories__/Searchable.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+  searchable: false,
+  disabled: false,
+  helper: 'helper',
+}
+
+Searchable.parameters = {
+  docs: {
+    description: {
+      story:
+        'Add a search bar in the dropdown to search through the different options. If there are less than 6 options, the search bar will **not** appear.',
```
matthprost, line 37:
```suggestion
        'Add a search bar in the dropdown to search through the different options. If there are less than 6 options, the search bar will **not** appear even if the prop is set to `true`.',
```

matthprost (changes requested):

matthprost (approved):
Maybe wait for me to come back from vacation to merge this one as there will be nobody to fix in case of issue 😅

PR #4357, opened 2024-10-17
fix(CopyButton): fix children type

johnrazeur:
## Type

- Bug

Fix `CopyButton` children prop type.

### DIFF
--- a/packages/ui/src/components/CopyButton/index.tsx
+++ b/packages/ui/src/components/CopyButton/index.tsx
@@ -1,3 +1,4 @@
+import type { ReactNode } from 'react'
 import useClipboard from 'react-use-clipboard'
 import { Button } from '../Button'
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/CopyButton/index.tsx
```diff
@@ -16,7 +17,7 @@ type CopyButtonProps = {
   bordered?: boolean
   className?: string
   'data-testid'?: string
-  children?: string
+  children?: ReactNode
```
matthprost, line 20: This was done on purpose to be a string, why do you need a react node ? We don't want the devs to be able to put anything else than a string

matthprost (changes requested):

packages/ui/src/components/CopyButton/index.tsx
matthprost: Is it because of translation?

PR #4377, opened 2024-10-22
fix: slider and popup to work better

matthprost:
## Type

- Bug

#### What is expected?

- Fix `<Slider />` not to have min or max width and take full width when it can
- Fix `<Popup />` to have better arrow positioning

| url  | Screenshot 2024-10-22 at 11 28 47 | Screenshot 2024-10-22 at 11 28 51 |

### DIFF
--- a/packages/ui/src/components/Slider/components/DoubleSlider.tsx
+++ b/packages/ui/src/components/Slider/components/DoubleSlider.tsx
@@ -4,7 +4,7 @@ import { useEffect, useId, useMemo, useRef, useState } from 'react'
 import { NumberInputV2 } from '../../NumberInputV2'
 import { Stack } from '../../Stack'
 import { Text } from '../../Text'
-import { SLIDER_WIDTH, THUMB_SIZE } from '../constant'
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/Slider/components/DoubleSlider.tsx
```diff
@@ -187,7 +189,7 @@ export const DoubleSlider = ({
   // Get slider size (for options)
   useEffect(() => {
     const setWidthResize = () => {
-      setWidth(refSlider.current?.offsetWidth ?? SLIDER_WIDTH.max)
+      setWidth(refSlider.current?.offsetWidth ?? 0)
```
alexandre-combemorel, line 192: I find 0 to be confusing, can we not put `Number(refSlider.current?.offsetWidth)` so we know that this value will always be taken, as long as the useEffect set a default value

alexandre-combemorel (approved):

JulienSaguez (approved):

lisalupi (approved):

PR #4459, opened 2024-11-14
fix(TextArea): row auto behavior

lisalupi:
## Type

- Bug

#### What is expected?

`<TextArea />` and `<TextAreaField />` : new prop `autoExpandMax`

We have 3 cases : 
- `rows = x` : the input will have a **fixed height**, equals to the height of x rows
- `maxRows = x`: the input size will fit the content. When there are x+1 rows, it will stop to grow (**upper bound** in number of rows)
- `maxRows = x` and `rows = y` : the input will have a min-height of y (**lower bound**) and max-height of x (**upper bound**)

### DIFF
--- a/packages/ui/src/components/TextArea/index.tsx
+++ b/packages/ui/src/components/TextArea/index.tsx
@@ -212,11 +212,11 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
         const maxHeight = maxRows * lineHeight
 
         textArea.style.height = `${textArea.scrollHeight + 2}px`
-        textArea.style.maxHeight = `calc(${maxHeight}px + ${padding} + ${padding}`
+        textArea.style.maxHeight = `calc(${maxHeight}px + 2*${padding})`
 
         if (typeof rows === 'number') {
           const minHeight = rows * lineHeight
-          textArea.style.minHeight = `calc(${minHeight}px + ${padding} + ${padding}`
+          textArea.style.minHeight = `calc(${minHeight}px + 2*${padding})`
         }
       }
     }, [value, rows, theme, maxRows])
--- a/packages/ui/src/components/TextArea/index.tsx
+++ b/packages/ui/src/components/TextArea/index.tsx
@@ -136,9 +136,13 @@ type TextAreaProps = {
    */
   helper?: ReactNode
   /**
-   * Number of rows to display. If 'auto', the textarea will grow with the content and won't be resizable.
+   * Number of rows to display. If 'auto', the textarea will grow with the content and won't be resizable **WARNING**: `auto` is deprecated - use prop `autoExpandMax` instead
    */
   rows?: number | 'auto'
+  /**
+   * Text area will grow with the content and won't be resizable with a maximum number of rows.
+   */
+  autoExpandMax?: number
   minLength?: number
   maxLength?: number
   tooltip?: string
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/TextArea/index.tsx
```diff
@@ -164,7 +168,8 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
       value,
       onChange,
       placeholder,
-      rows = 3,
+      rows = 'auto',
```
matthprost, line 171: Keep rows at 3 by default

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -136,9 +136,13 @@ type TextAreaProps = {
    */
   helper?: ReactNode
   /**
-   * Number of rows to display. If 'auto', the textarea will grow with the content and won't be resizable.
+   * Number of rows to display. If 'auto', the textarea will grow with the content and won't be resizable **WARNING**: `auto` is deprecated - use prop `autoExpandMax` instead
    */
   rows?: number | 'auto'
```
matthprost, line 141: We should keep `auto`, and make `rowsMax` in the way that if we have `row="auto" maxRows={10}` it will follow the number of lines until the 10th one

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -164,7 +168,8 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
       value,
       onChange,
       placeholder,
-      rows = 3,
+      rows = 'auto',
+      autoExpandMax,
```
matthprost, line 172: rename in `rowsMax` with a comment to explain it needs to be used in association with `rows="auto"`

packages/ui/src/components/TextArea/index.tsx
```diff
@@ -194,12 +199,27 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
...
-        textArea.style.height = `${textArea.scrollHeight === 61 ? 46 : textArea.scrollHeight + 2}px`
+        textArea.style.height = `${textArea.scrollHeight + 2}px`
+      } else if (textArea && maxRows) {
+        const lineHeight = parseFloat(getComputedStyle(textArea).lineHeight)
+
+        textArea.style.height = 'auto'
+        const maxHeight = maxRows * lineHeight
+
+        textArea.style.height = `${textArea.scrollHeight + 2}px`
+        textArea.style.maxHeight = `calc(${maxHeight}px + ${padding} + ${padding}`
```
johnrazeur, line 215:
```suggestion
        textArea.style.maxHeight = `calc(${maxHeight}px + ${2 * padding}`
```

johnrazeur:
I think padding bottom should be increase:
image

johnrazeur:
I think padding bottom should be increase:
image

johnrazeur (approved):

packages/ui/src/components/TextArea/index.tsx
lisalupi: padding is a string

matthprost:
```suggestion
        textArea.style.maxHeight = `calc(${maxHeight}px + 2* (${ padding})`
```

Should work this way then

matthprost (approved):

PR #4467, opened 2024-11-18
feat: shift click for multiselect

lisalupi:
## Type

- Enhancement

#### What is expected?

`<List />` and `<Table />` : hold shift to multiselect when they are selectable

### DIFF
--- a/packages/ui/src/components/List/ListContext.tsx
+++ b/packages/ui/src/components/List/ListContext.tsx
@@ -1,10 +1,17 @@
-import type { ComponentProps, Dispatch, ReactNode, SetStateAction } from 'react'
+import type {
+  ComponentProps,
+  Dispatch,
+  ReactNode,
+  RefObject,
+  SetStateAction,
+} from 'react'
 import {
   createContext,
   useCallback,
   useContext,
   useEffect,
   useMemo,
+  useRef,
   useState,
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/List/ListContext.tsx
```diff
@@ -209,6 +210,68 @@ export const ListProvider = ({
     ],
   )
 
+  const [lastCheckedIndex, setLastCheckedIndex] = useState<null | number>(null)
+
+  useEffect(() => {
+    const checkboxes = document.querySelectorAll<HTMLInputElement>(
+      '[name="list-select-checkbox"]',
+    )
```
matthprost, line 218: This won't work if you have multiple list / table in the same page. You need to find a more specific way to target it. Maybe the ref is the solution


https://github.com/user-attachments/assets/345d781c-6b91-4548-913f-07f4314ce37e

matthprost (changes requested):

DorianMaliszewski:
Is it possible to add some unitary tests on it ?

DorianMaliszewski (approved):

matthprost (approved):

DorianMaliszewski (approved):

matthprost (approved):

PR #4473, opened 2024-11-19
feat: switchButtonField

lisalupi:
## Type

- Feature

#### What is expected?

New component `<SwitchButtonField />`


packages/form/src/components/SwitchButtonField/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  TFieldValues extends FieldValues,
+  TFieldName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
+>({
+  name,
+  leftButton,
+  rightButton,
+  onBlur,
+  onFocus,
+  tooltip,
+}: SwitchButtonFieldProps<TFieldValues, TFieldName>) => {
```
matthprost, line 28: We are trying to switch a bit the strategy for from elements, now we want to pick all the props and exclude the one we don't want (here it will be `name`). Then in the TSX we wan't to spread the props. This way if a new prop is added on the component it will be automatically propagated on froms 😄

matthprost (changes requested):

packages/form/src/components/SwitchButtonField/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { useController } from 'react-hook-form'
+import type { BaseFieldProps } from '../../types'
+
+type SwitchButtonFieldProps<
+  TFieldValues extends FieldValues,
+  TFieldName extends FieldPath<TFieldValues>,
+> = BaseFieldProps<TFieldValues, TFieldName> &
+  Omit<ComponentProps<typeof SwitchButton>, 'value' | 'onChange' | 'name'> & {
+    name: FieldPath<TFieldValues>
+  }
```
DorianMaliszewski, line 13: can We add the original onChange like for onBlur ? if we want to upfate other field based on this onChange ?

DorianMaliszewski (approved):

.changeset/spotty-cups-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/form": patch
```
matthprost, line 2:
```suggestion
"@ultraviolet/form": minor
```

matthprost (approved):

PR #4525, opened 2024-11-29
fix(fonts): package json exports

matthprost:
## Type

- Bug

#### What is expected?

Fix package.json export


packages/fonts/package.json
```diff
@@ -34,6 +34,7 @@
   ],
   "type": "module",
   "exports": {
+    ".": { "style": "./dist/fonts.css" },
     "./fonts.css": "./dist/fonts.css"
```
alexandre-combemorel, line 38: Have you tested on shire directly? I could only make it work with 
```
"exports": {
    ".": {
      "require": "./dist/fonts.css",
      "import": "./dist/fonts.css",
      "default": "./dist/fonts.css"
    },
    "./fonts.css": {
      "require": "./dist/fonts.css",
      "import": "./dist/fonts.css",
      "default": "./dist/fonts.css"
    }
  },
```

matthprost: yes you are right, I will push a new fix soon 😄

alexandre-combemorel (approved):

PR #4535, opened 2024-12-03
refactor: list and table to be an actual table in dom

matthprost:
# Refactoring of `<List />` and `<Table />` components 📦

## List

List used to be a bunch of `div` with different `role` to work like a table but it suffered from columns issues if not correctly set.

It now uses a real `table` element and `thead` and `tbody` to correctly handle columns and rows.
Each elements uses `display: table-*` to correctly handle columns and rows.
The expandable is tricked, by adding a new `tr` on click and a `transform` to move it up under the clicked row.

## Table

Table was correctly using `table` element but the columns were not correctly handled.

It now uses `display: table-*` for each elements of the table.

So now if someone sets a total width that is not fitting the screen it will automatically relocate the width space as seen here:

Screenshot 2024-12-03 at 11 14 42

The expandable is an added `tr` under the clicked row.

| Table  | Screenshot 2024-12-03 at 11 12 21 | Screenshot 2024-12-03 at 11 15 04 |
| List  | Screenshot 2024-12-03 at 17 19 28 | Screenshot 2024-12-03 at 17 19 47 |

### Bonus: responsiveness improval ✨ 

List will now shrink with available space until no more space is available and it will trigger horizontal scroll.

Before:

https://github.com/user-attachments/assets/21ece723-f49c-4e93-954b-00fad00a6bb3

After:

https://github.com/user-attachments/assets/ed5ace69-00fd-460f-87e1-c8d8ef4c699c


DorianMaliszewski (approved):

packages/ui/src/components/Table/Row.tsx
```diff
@@ -201,9 +206,9 @@ export const Row = ({
               : undefined
           }
         >
-          {expandable}
+          <Cell colSpan={6}>{expandable}</Cell>
```
fabienhebert, line 209: Could you explain this `6` ?

fabienhebert (approved):

PR #4537, opened 2024-12-04
refactor: pagination

lisalupi:
## Type

- Refactor

#### What is expected?

Refactor and enhancement of `<Pagination />` : 
- Number of results on the bottom left 
- Number of items to display
- Styling

| url  | Capture d’écran 2024-12-04 à 16 00 36|  Capture d’écran 2024-12-04 à 16 00 02 |
| url  | Capture d’écran 2024-12-04 à 16 02 06 |  Capture d’écran 2024-12-04 à 16 01 17 |

PerPage : 
Capture d’écran 2024-12-04 à 16 02 54

### DIFF
--- a/packages/ui/src/components/Pagination/PaginationButtons.tsx
+++ b/packages/ui/src/components/Pagination/PaginationButtons.tsx
@@ -1,32 +1,27 @@
 import styled from '@emotion/styled'
 import { useCallback, useMemo } from 'react'
 import { Button } from '../Button'
+import { Stack } from '../Stack'
 import { Text } from '../Text'
 import { getPageNumbers } from './getPageNumbers'
 
-const PageNumbersContainer = styled.div`
-  display: flex;
-  gap: ${({ theme }) => theme.space['1']};
+const PageNumbersContainer = styled(Stack)`
   margin: 0 ${({ theme }) => theme.space['1']};
 `
 
-const PageSwitchContainer = styled.div`
-  display: flex;
-  gap: ${({ theme }) => theme.space['1']};
+const PageButton = styled(Button, {
+  shouldForwardProp: prop => !['width'].includes(prop),
+})<{ width: string }>`
+  width: ${({ theme, width }) => (width === 'small' ? theme.sizing[400] : theme.sizing[500])};
 `
 
-const StyledContainer = styled.div`
-  display: flex;
-`
-const PageButton = styled(Button)`
-  width: ${({ theme }) => theme.sizing[600]};
-`
-
-const Ellipsis = styled(Text)`
+const Ellipsis = styled(Text, {
+  shouldForwardProp: prop => !['width'].includes(prop),
+})<{ size: string }>`
   align-content: center;
   padding: ${({ theme }) => theme.space[1]};
-  height: ${({ theme }) => theme.sizing[600]};
-  width: ${({ theme }) => theme.sizing[600]};
+  height: ${({ theme, size }) => (size === 'small' ? theme.sizing[400] : theme.sizing[500])};
+  width: ${({ theme, size }) => (size === 'small' ? theme.sizing[400] : theme.sizing[500])};
 `
 
 type PaginationButtonsProps = {
@@ -37,6 +32,7 @@ type PaginationButtonsProps = {
   pageTabCount?: number
   className?: string
   'data-testid'?: string
+  perPage: boolean
 }
 
 type MakeButtonProps = {
@@ -45,6 +41,7 @@ type MakeButtonProps = {
   disabled?: boolean
   page: number
   handlePageClick: (pageNumber: number) => () => void
+  perPage: boolean
 }
 const MakeButton = ({
   hasEllipsisBefore,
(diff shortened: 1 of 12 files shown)


matthprost (changes requested):
You have some issues with your coverage, it seems that you are calling `onChange` in the test without using `vi.fn()`. I think this is what makes the codecov broken

packages/ui/src/components/Pagination/PaginationButtons.tsx
```diff
@@ -0,0 +1,152 @@
...
+`
+
+const StyledContainer = styled.div`
+  display: flex;
+`
+const PageButton = styled(Button)`
+  width: ${({ theme }) => theme.space[6]};
+`
+
+const Ellipsis = styled(PageButton)`
```
matthprost, line 24: Why making it a button?

packages/ui/src/components/Pagination/PaginationButtons.tsx
```diff
@@ -0,0 +1,152 @@
...
+          return (
+            <React.Fragment key={pageNumber}>
+              <Ellipsis
+                aria-label="ellipsis"
+                key={`ellipsis-page-${pageNumber}`}
+                disabled={disabled}
+                variant="ghost"
+                sentiment="neutral"
+                type="button"
+                tabIndex={-1}
```
matthprost, line 120: This won't be necessary if we don't make this as a button

packages/ui/src/components/Pagination/__stories__/PerPage.stories.tsx
```diff
@@ -0,0 +1,70 @@
+import styled from '@emotion/styled'
+import type { StoryFn } from '@storybook/react'
+import { useState } from 'react'
+import { Pagination } from '..'
+import { Badge } from '../../Badge'
+import { Stack } from '../../Stack'
+
+const NUMBER_OF_ITEMS = 134
+const StyledList = styled.ul`
+  height: 210px;
```
matthprost, line 10: no px please, rem only

packages/ui/src/components/Pagination/PaginationButtons.tsx
```diff
@@ -0,0 +1,152 @@
...
+                onClick={handlePageClick(pageNumber)}
+                type="button"
+              >
+                {pageNumber}
+              </PageButton>
+            )
+          }
+
+          return (
+            <React.Fragment key={pageNumber}>
```
matthprost, line 112: This is no good, we have an empty div here that shouldn't be there

Screenshot 2024-12-05 at 15 09 37

packages/ui/src/components/Pagination/PaginationButtons.tsx
```diff
@@ -0,0 +1,152 @@
...
+                disabled={disabled}
+                variant="ghost"
+                sentiment="neutral"
+                type="button"
+                tabIndex={-1}
+              >
+                ...
+              </Ellipsis>
+              <PageButton
+                aria-label={`Page ${pageNumber}`}
```
matthprost, line 125: not needed I think, the children of the button should be explicit enough

packages/ui/src/components/Pagination/PerPage.tsx
```diff
@@ -0,0 +1,73 @@
...
+  },
+  {
+    value: '50',
+    label: '50',
+  },
+  {
+    value: '100',
+    label: '100',
+  },
+]
```
matthprost, line 24:
```suggestion
const optionsItemsPerPage = [
  {
    value: '10',
    label: '10',
  },
  {
    value: '25',
    label: '25',
  },
  {
    value: '50',
    label: '50',
  },
  {
    value: '100',
    label: '100',
  },
] as const
```

matthprost (changes requested):

mhervouet:
For pagination with item number per page, the size should be small for `selectInput` and `Buttons`

DorianMaliszewski (approved):

packages/ui/src/components/Pagination/getPageNumbers.ts
```diff
@@ -1,30 +1,38 @@
...
-  return Array.from({ length: end - start + 1 }, (_, index) => start + index)
+  const pagesToShowBeforeTruncation = truncationBefore
+    ? [1]
+    : Array.from({ length: range }, (_, index) => index + 1)
+  const pagesToShowAfterTruncation = truncationAfter
+    ? [pageCount]
+    : Array.from({ length: range }, (_, index) => index + pageCount - range + 1)
+  const pagesToShowBetweenTruncation =
+    truncationBefore && truncationAfter
+      ? Array.from({ length: 3 }, (_, index) => currentPage + index - 1)
```
johnrazeur, line 30: The `3` should be in a const.

johnrazeur (approved):

packages/ui/src/components/Pagination/PaginationButtons.tsx
```diff
@@ -0,0 +1,153 @@
+import styled from '@emotion/styled'
+import { useCallback, useMemo } from 'react'
+import { Button } from '../Button'
+import { Text } from '../Text'
+import { getPageNumbers } from './getPageNumbers'
+
+const PageNumbersContainer = styled.div`
```
vincentaudebert, line 7: Could we not use Row and Stack around these? Maybe styled but at least it would reduce CSS

PR #4572, opened 2024-12-11
feat(Text): white space

lisalupi:
## Type

- Feature
#### What is expected?

`<Text />`: new prop `whiteSpace`

1. Added a new prop 

2. Used the new prop in various Ultraviolet components (Snippet, Navigation...)


packages/ui/src/components/Text/__tests__/index.test.tsx
```diff
@@ -92,4 +92,20 @@ describe('Text', () => {
         Lorem Ipsum
       </Text>,
     ))
+
+  test(`with monochrome `, () =>
```
radhi-nasser, line 96:
```suggestion
  test(`with monochrome`, () =>
```

radhi-nasser (approved):

packages/plus/src/components/Navigation/components/Item.tsx
```diff
@@ -590,6 +588,7 @@ export const Item = memo(
                   }
                   animation={animation}
                   disabled={disabled}
+                  whiteSpace="pre-wrap"
```
matthprost, line 591: Here it's no going to work as before: `white-space: ${({ animation }) => (animation ? 'nowrap' : 'normal')};` its conditioned by animation

packages/plus/src/components/Navigation/components/Item.tsx
```diff
@@ -602,6 +601,7 @@ export const Item = memo(
                     animation={animation}
                     disabled={disabled}
                     subLabel
+                    whiteSpace="pre-wrap"
```
matthprost, line 604: same here

packages/plus/src/components/Navigation/components/Item.tsx
```diff
@@ -805,7 +805,7 @@ export const Item = memo(
             flex={1}
             width="100%"
           >
-            <WrapText as="span" variant="bodySmall">
+            <WrapText as="span" variant="bodySmall" whiteSpace="pre-wrap">
```
matthprost, line 808: Here it's also incorrect, the condition is `white-space: ${({ animation }) => (animation ? 'nowrap' : 'normal')};` so if animation doesn't exists, which is the case here, it should be normal not pre-wrap

packages/ui/src/components/Text/index.tsx
```diff
@@ -19,6 +19,7 @@ const PROMINENCES = {
 
 type ProminenceProps = keyof typeof PROMINENCES
 type PlacementProps = React.CSSProperties['textAlign']
+type WhiteSpaceProps = React.CSSProperties['whiteSpace']
```
matthprost, line 22: I know it was there before but we avoid `React.` in our code base, instead you can import CSSProperties from react above

matthprost (changes requested):

packages/plus/src/components/Navigation/components/Item.tsx
lisalupi: Previously there was `white-space: pre-wrap` which took over `white-space: ${({ animation }) => (animation ? 'nowrap' : 'normal')};` (see l131 & l136) , that's why I put pre-wrap here

matthprost (approved):

PR #4587, opened 2024-12-16
fix: table and list to hide scroll bar is no overflow

matthprost:
## Type

- Bug

#### What is expected?

Hide scroll bar on overflow x if there is no overflow.

| url  | Screenshot 2024-12-16 at 08 55 32 | Screenshot 2024-12-16 at 08 55 39 |
| url | Screenshot 2024-12-16 at 11 19 17 | Screenshot 2024-12-16 at 11 19 25 |

### DIFF
--- a/packages/ui/src/components/List/HeaderRow.tsx
+++ b/packages/ui/src/components/List/HeaderRow.tsx
@@ -1,10 +1,9 @@
-import { useTheme } from '@emotion/react'
 import styled from '@emotion/styled'
 import type { ReactNode } from 'react'
 import { Checkbox } from '../Checkbox'
 import { HeaderCell } from './HeaderCell'
 import { useListContext } from './ListContext'
-import { EXPANDABLE_COLUMN_SIZE, SELECTABLE_CHECKBOX_SIZE } from './constants'
+import { SELECTABLE_CHECKBOX_SIZE } from './constants'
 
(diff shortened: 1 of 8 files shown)


radhi-nasser-scaleway (approved):

philibea (approved):

packages/ui/src/components/List/HeaderRow.tsx
```diff
@@ -61,7 +58,9 @@ export const HeaderRow = ({ children, hasSelectAllColumn }: RowProps) => {
     <thead>
       <StyledHeaderRow>
         {hasSelectAllColumn ? (
-          <NoPaddingHeaderCell width={theme.sizing[SELECTABLE_CHECKBOX_SIZE]}>
+          <NoPaddingHeaderCell
+            maxWidth={theme.sizing[SELECTABLE_CHECKBOX_SIZE]}
```
philibea, line 62: maxWidth c'est du natif ? il faudrait le filter sur le styled no ?

philibea:
```typescript
{
  shouldForwardProp: prop => !['maxWidth'].includes(prop),
})<{
  maxWidth: string
}>`
```

matthprost: true I don't why I made my life so compicated

PR #4604, opened 2024-12-19
feat(MenuV2): add search feature in menu

matthprost:
## Type

- Enhancement

#### What is expected?

- New prop `searchable`, when set to `true` a search bar will appear on the MenuV2. This can be useful for searching for a specific items quick. The searchbar is focused automatically when menu appear.
- New prop `hideOnClickItem`: this is to replace function as children being used to close the menu when an item is clicked.
- Global refactoring of the folder structure and addition of a provider needed for state management

Menu with prop `searchable` and `hideOnClickItem` set to `true`:

https://github.com/user-attachments/assets/c1c31a95-36be-42ea-8e0e-57c2b761f8a9


mhervouet (approved):

lisalupi (approved):

radhi-nasser-scaleway (approved):

BABAK0T0 (approved):

packages/ui/src/components/MenuV2/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const SIZES = {
+  small: '180px',
+  medium: '280px',
+  large: '380px',
```
fabienhebert, line 4: just a question, not an ask for changes:  no REM sizes here ? I saw a min-width of 20rem

packages/ui/src/components/MenuV2/MenuContent.tsx
```diff
@@ -0,0 +1,198 @@
...
+    })
+
+    const onSearch = useCallback(
+      (value: string) => {
+        if (typeof children === 'object') {
+          setLocalChild(searchChildren(children, value))
+        }
+      },
+      [children],
+    )
```
fabienhebert, line 131: we are sure we wan't to give responsabiliy to Menu to search instead of just giving responsabilities to component caller

But i'm fine also with this solution which is automatic

```
<Menu
     onSearch={setSearchedValue}
     currentSearchValue={searchedValue}
 >
     {projects.filter(project)
         => project.name.contains(searchValue) || ...even condition on invisible properties
```

fabienhebert (approved):

Lawndlwd (approved):

packages/ui/src/components/MenuV2/index.tsx
```diff
@@ -1,209 +1,22 @@
...
-const FwdMenu = forwardRef(
+import type { Ref } from 'react'
+import { forwardRef } from 'react'
+import { Menu } from './MenuContent'
+import { MenuProvider } from './MenuProvider'
+import { Group } from './components/Group'
+import Item from './components/Item'
+import type { MenuProps } from './types'
+
+const Component = forwardRef(
```
alexandre-combemorel, line 9: plus besoin avec React 19 non ?

packages/ui/src/components/MenuV2/MenuContent.tsx
alexandre-combemorel: I think Menu item should be more or less the same accros implementation, nothing like crazy DOM structure.
The helper function seems pretty good for nested check. So I'm keen to leave the responsibility to the DS.
We have the same for SelectInput, make sense for Menu as well

alexandre-combemorel (approved):

packages/ui/src/components/MenuV2/constants.ts
matthprost: yes you are correct

packages/ui/src/components/MenuV2/MenuContent.tsx
matthprost: Yes it's better to leave this responsibility to DS it will be easier to maintain I think.

packages/ui/src/components/MenuV2/index.tsx
matthprost: we still support react 18 for now so I guess we need to keep it

Issue #4618, opened 2024-12-24
[Bug]: List checkbox doesn't update state each click


DorianMaliszewski:
### Is it Front-End or Design related only?

Front-End

### Library

UI

### Current Behavior

Actually, when clicking on a checkbox in the List (since React 19) there are sometimes the checkbox doesn't update its state. On my video I put a plugin on mouse click (Red reverb), as you can see there are some clicks that don't trigger the update (Notable example between 0:18 and 0:24)

https://github.com/user-attachments/assets/023c256a-42ff-4c96-a63c-bf523c236e06

### Expected Behavior

Each time a click happen it should update the state of the checkbox as it was with React 18

### Context

Was diving into a problem when updating on internal console to React 19. You can see the bug when ordering a Domain by playing with checkboxes on the first step of the form

philibea:
> ### Is it Front-End or Design related only?
> Front-End
> 
> ### Library
> UI
> 
> ### Current Behavior
> Actually, when clicking on a checkbox in the List (since React 19) there are sometimes the checkbox doesn't update its state. On my video I put a plugin on mouse click (Red reverb), as you can see there are some clicks that don't trigger the update (Notable example between 0:18 and 0:24)
> 
>  20241224_110603.mp4 
> ### Expected Behavior
> Each time a click happen it should update the state of the checkbox as it was with React 18
> 
> ### Context
> Was diving into a problem when updating on internal console to React 19. You can see the bug when ordering a Domain by playing with checkboxes on the first step of the form



I made a quick with playwright to see this issue https://github.com/scaleway/ultraviolet/pull/4619

PR #4619, opened 2024-12-24
feat(checkbox-list): shift event checkbox was broken

philibea:
## Type

- Enhancement

#### What is expected?

Add the possibility to test with playwright

Show that the checkbox event is broken on the list 
Close https://github.com/scaleway/ultraviolet/issues/4618


packages/ui/src/components/Checkbox/index.tsx
```diff
@@ -374,11 +356,11 @@ export const Checkbox = forwardRef(
...
             aria-invalid={!!error}
             aria-describedby={error ? `${localId}-hint` : undefined}
-            aria-checked={state === 'indeterminate' ? 'mixed' : state}
+            aria-checked={checked === 'indeterminate' ? 'mixed' : checked}
             aria-label={ariaLabel}
-            checked={state === 'indeterminate' ? false : state}
+            checked={checked === true ? checked : undefined}
             inputSize={size ?? theme.sizing['300']}
-            onChange={onLocalChange}
+            onChange={onChange}
```
fabienhebert, line 363: should we keep logic `onChange={progress ? onChange : undefined}` ? I guess not but a reminder just in case, may be we could disabled checkbox instead when `inProgress`

fabienhebert (approved):

packages/ui/src/components/Checkbox/index.tsx
philibea: done with disabled={disabled || progress}

packages/ui/src/components/Table/HeaderRow.tsx
```diff
@@ -13,8 +13,7 @@ type HeaderRowProps = {
 export const HeaderRow = ({ children, hasSelectAllColumn }: HeaderRowProps) => {
   const {
     allRowSelectValue,
-    selectAll,
-    unselectAll,
+    checkboxAllHandler,
```
vincentaudebert, line 16: Why not `selectAllHandler` ?

packages/ui/src/components/Radio/__stories__/Template.stories.tsx
```diff
@@ -4,7 +4,7 @@ import { Radio } from '..'
 export const Template: StoryFn<typeof Radio> = args => <Radio {...args} />
 
 Template.args = {
-  checked: false,
+  // checked: false,
```
vincentaudebert, line 7: delete?

packages/ui/src/components/Checkbox/index.tsx
```diff
@@ -332,35 +328,30 @@ export const Checkbox = forwardRef(
...
-        setState(current =>
-          current === 'indeterminate' ? false : event.target.checked,
-        )
-      },
-      [onChange, progress, setState],
-    )
+    const isDisabled = disabled || progress
+    /**
+     * note: checked should be true | undefined as it's seems to break e2e playwright test actually
+     * as the checked attribue is added to html, maybe related to playwright toCheck code || emotion should not add this attributes when it's false or react 19 related ?
```
vincentaudebert, line 341:
```suggestion
     * as the checked attribute is added to html, may be related to playwright toCheck code || emotion should not add this attribute when it's false or react 19 related ?
```

vincentaudebert:
IMHO on shift + click

Should mass check if one of the boxes is uncheck
Should mass uncheck if all the boxes are checked.

vincentaudebert (approved):

PR #4635, opened 2025-01-02
feat: timeInputV2

lisalupi:
## Type

- Feature

#### What is expected?

New component `<TimeInputV2 />`


packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+                    :
+                  </CustomText>
+                )}
+              </Stack>
+            )
+          })}
+          {timeFormat === 12 ? (
+            <Input
+              value={time.period?.toUpperCase()}
+              placeholder={placeholder.period ?? '-M'}
```
matthprost, line 402: I'd say we should set AM instead of -M

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+            }
+            const fullName = () => {
+              if (type === 'h') return 'hours'
+              if (type === 'm') return 'minutes'
+
+              return 'seconds'
+            }
+
+            return (
+              <Stack key={type} gap={0} direction="row">
```
matthprost, line 355:
```suggestion
              <Stack key={type} direction="row">
```

I don't think its needed

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+        </Text>
+        {required ? <AsteriskIcon size={8} sentiment="danger" /> : null}
+        {labelDescription ? (
+          <Text as="label" variant="bodySmall">
+            {labelDescription}
+          </Text>
+        ) : null}
+        {labelDescription && typeof labelDescription !== 'string'
+          ? labelDescription
+          : null}
```
matthprost, line 325: this is a bug, we will display labelDescription twice

```suggestion
        {labelDescription ? (
            {labelDescription}
        ) : null}
```

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+  onChange?: (time: Time, timePeriod?: string) => void
+  onBlur?: (event: FocusEvent<HTMLInputElement>) => void
+  onFocus?: (event: FocusEvent<HTMLInputElement>) => void
+  className?: string
+  id?: string
+  size?: 'small' | 'medium' | 'large'
+  timeFormat?: 12 | 24
+}
+
+const DEFAULT_PLACEHOLDER = { h: '00', m: '00', s: '00' } as const
```
matthprost, line 169:
```suggestion
const DEFAULT_PLACEHOLDER = { h: '00', m: '00', s: '00' } as const

type TimeInputProps = {
  label?: ReactNode
  placeholder?: Time
  value?: Time
  clearable?: boolean
  required?: boolean
  labelDescription?: ReactNode
  helper?: string
  disabled?: boolean
  readOnly?: boolean
  error?: boolean | string
  'data-testid'?: string
  onChange?: (time: Time, timePeriod?: string) => void
  onBlur?: (event: FocusEvent<HTMLInputElement>) => void
  onFocus?: (event: FocusEvent<HTMLInputElement>) => void
  className?: string
  id?: string
  size?: 'small' | 'medium' | 'large'
  timeFormat?: 12 | 24
}

```

Better for readability to have the type of the function right above the function itself

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+              return 'seconds'
+            }
+
+            return (
+              <Stack key={type} gap={0} direction="row">
+                <Input
+                  value={time[type]}
+                  placeholder={placeholder[type]}
+                  data-size={size}
+                  data-disabled={disabled}
```
matthprost, line 360: prop disabled already exists, what is the purpose of the data-disabled here ?

in css you can directly access to disabled by doing `&:not(disabled)`

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+  value?: Time
+  clearable?: boolean
+  required?: boolean
+  labelDescription?: ReactNode
+  helper?: string
+  disabled?: boolean
+  readOnly?: boolean
+  error?: boolean | string
+  'data-testid'?: string
+  onChange?: (time: Time, timePeriod?: string) => void
```
matthprost, line 160:
```suggestion
  onChange?: (value: Time, valuePeriod?: string) => void
```

To be consistent between our inputs

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+              <Stack key={type} gap={0} direction="row">
+                <Input
+                  value={time[type]}
+                  placeholder={placeholder[type]}
+                  data-size={size}
+                  data-disabled={disabled}
+                  data-readonly={readOnly}
+                  readOnly
+                  disabled={disabled}
+                  aria-label={`${fullName()}-input`}
```
matthprost, line 364: Aria label should be set via a prop instead. It should be set when no label is set to give indication to screen reader.

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+  error?: boolean | string
+  'data-testid'?: string
+  onChange?: (time: Time, timePeriod?: string) => void
+  onBlur?: (event: FocusEvent<HTMLInputElement>) => void
+  onFocus?: (event: FocusEvent<HTMLInputElement>) => void
+  className?: string
+  id?: string
+  size?: 'small' | 'medium' | 'large'
+  timeFormat?: 12 | 24
+}
```
matthprost, line 167: here you could add `aria-label` and conditionally  required either `label` or `aria-label`

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+`
+
+type TimeInputProps = {
+  label?: ReactNode
+  placeholder?: Time
+  value?: Time
+  clearable?: boolean
+  required?: boolean
+  labelDescription?: ReactNode
+  helper?: string
```
matthprost, line 155: Should be a react node, sometimes we need to add a link in the helper

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,469 @@
...
+  }
+
+  // Go to previous input
+  const handlePrevious = (type: 'h' | 'm' | 's') => {
+    if (type === 'm') refHours.current?.focus()
+    if (type === 's') refMinutes.current?.focus()
+  }
+
+  return (
+    <Stack gap={0.5} className={className} id={id} data-testid={dataTestId}>
```
matthprost, line 312: The id and data-testid should be on the `TimeInputWrapper` instead

matthprost (changes requested):

BABAK0T0 (approved):

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,511 @@
...
+  className,
+  id,
+  autoFocus,
+  'data-testid': dataTestId,
+  placeholder = DEFAULT_PLACEHOLDER,
+}: TimeInputProps) => {
+  const correctEmpty = timeFormat === 24 ? EMPTY_TIME_24 : EMPTY_TIME_12
+  const [time, setTime] = useState<Time>(
+    value ? formatValue(value, timeFormat) : correctEmpty,
+  )
```
matthprost, line 209: There is an issue for in the value. If I look at the old time input we use to have a date that includes the time into it. This is the right way to do it as we want to use the standard date type for it. Here we use a custom thing and it differ with our old usage and does not take in account all the specificity of a time (time zone, etc)

Screenshot 2025-01-09 at 11 47 36
Screenshot 2025-01-09 at 11 47 47

matthprost: I guess each time we change the time you could create a `new Date()` with the current date and time set on the input

matthprost (changes requested):

packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -0,0 +1,511 @@
...
+                }
+              }}
+              onKeyDown={event => {
+                if (!readOnly && !disabled) {
+                  if (event.key === 'ArrowUp' || event.key === 'ArrowDown') {
+                    event.preventDefault()
+                    const newTime = {
+                      ...time,
+                      period:
+                        time.period === 'am' ? 'pm' : ('am' as 'am' | 'pm'),
```
johnrazeur, line 456:
```suggestion
                        time.period === 'am' ? 'pm' : 'am',
```
You shouldn't need that.

packages/ui/src/components/TimeInputV2/index.tsx
johnrazeur: I reopen this comment.

johnrazeur (approved):

matthprost (approved):

mhervouet (approved):

Issue #4625, opened 2024-12-30
[Bug]: Shift + click on listing


vincentaudebert:
### Is it Front-End or Design related only?

Front-End

### Library

UI

### Current Behavior

1) If checkbox is "shift detected", there is one click latence before being able to click on another checkbox without shift.

2) If I keep shift key down and I click from A to B only "shift marked" checkboxes are selected/unselected

### Expected Behavior

1) It should click on the other checkbox directly.

2) All checkboxes from A to B should be selected/unselected (cf gmail for instance)

### Context

1) https://github.com/user-attachments/assets/9df8b2ab-8a94-4cff-b0d4-80f5eacce30f

2) https://github.com/user-attachments/assets/46e33fd7-de12-4af8-869e-6e0da8ecc68c

PR #4636, opened 2025-01-03
fix: shift select

lisalupi:
## Type

- Enhancement

#### What is expected?

Have a more intuitive behavior when shift+click on a selectable list/table (see 2. #4625) 

1. Removed hover behavior to replace with indexes

Before: 

https://github.com/user-attachments/assets/47b96313-03f4-44d7-bf40-53895e08d137

After : 

https://github.com/user-attachments/assets/7ce7d241-69b5-449a-b329-dbcb74e407e3

### DIFF
--- a/packages/ui/src/components/List/ListContext.tsx
+++ b/packages/ui/src/components/List/ListContext.tsx
@@ -4,7 +4,6 @@ import {
   useContext,
   useEffect,
   useMemo,
-  useRef,
   useState,
 } from 'react'
 import type {
@@ -46,7 +45,8 @@ export type ListContextValue = {
   selectRow: (rowId: string) => void
   unselectAll: () => void
   unselectRow: (rowId: string) => void
-  refList: RefObject<HTMLInputElement[]>
+  refList: RefObject<HTMLInputElement>[]
+  setRefList: Dispatch<SetStateAction<RefObject<HTMLInputElement>[]>>
   handleOnChange: (value: string, checked: boolean) => void
 }
 
@@ -89,7 +89,7 @@ export const ListProvider = ({
   const [selectedRowIds, setSelectedRowIds] = useState<RowState>({})
   const [lastCheckedCheckbox, setLastCheckedCheckbox] = useState<string>()
   const [inRange, setInRange] = useState<Set<number | string>>(new Set([]))
-  const refList = useRef<HTMLInputElement[]>([])
+  const [refList, setRefList] = useState<RefObject<HTMLInputElement>[]>([])
 
   const registerExpandableRow = useCallback(
     (rowId: string, expanded = false) => {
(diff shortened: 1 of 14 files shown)


packages/ui/src/components/List/Row.tsx
```diff
@@ -277,16 +277,13 @@ export const Row = forwardRef<HTMLTableRowElement, RowProps>(
...
       const { current } = checkboxRef
 
-      if (current) {
-        mapCheckbox.set(id, current)
+      if (refAtEffectStart && current && !refAtEffectStart.includes(current)) {
+        refList.current.push(current)
       }
-
-      return () => {
-        mapCheckbox.delete(id)
```
philibea, line 287: we must remove the checkbox of the refList.current when the row is delete.

packages/ui/src/components/List/Row.tsx
```diff
@@ -277,16 +277,13 @@ export const Row = forwardRef<HTMLTableRowElement, RowProps>(
     const canClickRowToExpand = !disabled && !!expandable && !expandButton
 
     useEffect(() => {
+      const refAtEffectStart = refList.current
       const { current } = checkboxRef
 
-      if (current) {
-        mapCheckbox.set(id, current)
+      if (refAtEffectStart && current && !refAtEffectStart.includes(current)) {
```
philibea, line 283: !refAtEffectStart.includes(current)

it's work ?

packages/ui/src/components/List/ListContext.tsx
```diff
@@ -196,55 +194,54 @@ export const ListProvider = ({
...
-
-        if (isHoverActive) {
-          setInRange(prev => new Set([...prev, checkbox.value]))
-        }
-
-        if (!lastCheckedIndex && !checkbox.disabled) {
-          setLastCheckedIndex(checkbox.value)
-        }
+      // Ensure that only existing checkboxes are in refList
+      if (refList.current) {
```
philibea, line 198: this was done previously with this 
```
useEffect(() => {
    const { current } = checkboxRowRef

    if (current) {
      mapCheckbox.set(id, current)
    }

    return () => {
      mapCheckbox.delete(id)
    }
  }, [mapCheckbox, id])
```

packages/ui/src/components/List/ListContext.tsx
```diff
@@ -196,55 +194,54 @@ export const ListProvider = ({
...
+          )
+          const lastCheckedIndex = targetCheckbox
+            ? refList.current.indexOf(targetCheckbox)
+            : undefined
+
+          if (lastCheckedIndex !== undefined) {
+            const start = Math.min(lastCheckedIndex, index)
+            const end =
+              Math.max(lastCheckedIndex, index) +
+              (Math.max(lastCheckedIndex, index) === index ? 0 : 1)
```
fabienhebert, line 223: I don't know if it's more readable, this comment is purely optionnal

```
let end = Math.max(lastCheckedIndex, index)
end +=  end === index ? 0 : 1
```

fabienhebert (approved):
it should be ok for me after considering @philibea feedbacks

lisalupi:
> Why removing hover effect?

It was not intuitive & it is overridden by the new version of shift-select (similar to how it was before)

packages/ui/src/components/List/Row.tsx
lisalupi: it is done somewhere else (l.199 of `ListContext.tsx`)

packages/ui/src/components/List/ListContext.tsx
```diff
@@ -196,99 +195,91 @@ export const ListProvider = ({
...
+    [selectRows],
+  )
+
+  useEffect(() => {
+    // Re-arrange refList everytime it is modified in order to keep the write order of the elements in it.
+    const allCheckboxes = [
+      ...document.querySelectorAll('input[type="checkbox"]:not([value="all"])'),
+    ].map(element => element as HTMLInputElement)
+    refList.current = allCheckboxes.map(checkbox => checkbox)
+  }, [refList.current.length])
```
matthprost, line 281: By doing this you take the risk to select every checkbox in the page. Let's say you have a 2 list in the same page you will select checkboxes for both list while only one is being edited.

We should avoid using a querySelector for interacting with the list / table

matthprost (changes requested):

packages/ui/src/components/Table/Row.tsx
```diff
@@ -170,23 +171,20 @@ export const Row = ({
...
-    const { current } = checkboxRowRef
-
-    if (current) {
-      mapCheckbox.set(id, current)
-    }
-
-    return () => {
-      mapCheckbox.delete(id)
-    }
-  }, [mapCheckbox, id])
```
matthprost, line 183: Why changing this? Here you automatically register and unregister the checkbox ref into the list provider and update the list of it depending on rows. This is useful especially if you have dynamic rows which happens quite often (deleting an element, etc)

matthprost (changes requested):

DorianMaliszewski (approved):

matthprost (approved):

PR #4652, opened 2025-01-09
fix: onchange and border

lisalupi:
## Type

- Bug
- Enhancement

#### What is expected?
`<SelectInputV2 />`: 
- onChange handle on checkboxes fix
- visual indication of focused item for keyboard users (if the element was already selected, it was not possible to see that the element had focus)
- no automatic scroll when using "enter" to select an item
Before ("Venus" has focus) : 
Capture d’écran 2025-01-09 à 15 41 29

After (when navigating with keyboard) : 
Capture d’écran 2025-01-09 à 15 40 30

### DIFF
--- a/packages/ui/src/components/SelectInputV2/Dropdown.tsx
+++ b/packages/ui/src/components/SelectInputV2/Dropdown.tsx
@@ -134,13 +134,7 @@ const DropdownItem = styled.div<{
     background-color: ${({ theme }) => theme.colors.primary.background};
     color: ${({ theme }) => theme.colors.primary.text};
     cursor: pointer;
-    outline: none;
   }
-
-
-  &:focus-visible {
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/SelectInputV2/Dropdown.tsx
```diff
@@ -136,6 +137,11 @@ const DropdownItem = styled.div<{
     outline: none;
   }
 
+
+  &:focus-visible {
+      border: ${({ theme }) => theme.colors.primary.borderWeakDisabled} 1px solid;
+    }
```
matthprost, line 143: I think the best and easiest for focus is to use the native focus of the browser instead of applying a custom one

matthprost (changes requested):

alexandre-combemorel (approved):

matthprost (approved):

PR #4737, opened 2025-02-05
feat(DateInput): new prop input

lisalupi:
## Type

- Feature

#### What is expected?
Added a new prop `input` to show only the calendar par of `DateInput`
Also fixed a problem in storybook for `DateField` (value was not updating for non-range values)


.changeset/cold-bananas-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/ui": patch
+---
+
+`<DateInput />`: new prop `input` to show calendar as a card, with no input
```
matthprost, line 5: You also need to patch DateField from forms. currently the prop `input` is not spread so we cannot use this new feature in forms.

packages/ui/src/components/DateInput/components/CalendarContent.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { Daily } from './CalendarDaily'
+import { Monthly } from './CalendarMonthly'
+
+const CapitalizedText = styled(Text)`
+  display: inline-block;
+  text-transform: lowercase;
+
+  &::first-letter {
+    text-transform: uppercase;
+  }
```
matthprost, line 17: Why not using `text-transform: capitalize;` ?

packages/ui/src/components/DateInput/components/CalendarContent.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { Button } from '../../Button'
+import { Stack } from '../../Stack'
+import { Text } from '../../Text'
+import { DateInputContext } from '../Context'
+import { getNextMonth, getPreviousMonth } from '../helpers'
+import { Daily } from './CalendarDaily'
+import { Monthly } from './CalendarMonthly'
+
+const CapitalizedText = styled(Text)`
+  display: inline-block;
```
matthprost, line 12: not sure this is needed as you're using a span it should already be inline. But I'm not 100% sure, you need to check haha

matthprost (changes requested):

JulienSaguez (approved):

PR #4738, opened 2025-02-05
feat: new component label

lisalupi:
## Type

- Feature

### Summarise concisely:
New component `Label` that can be used to replace the prop `label` inside input components (DateInput, TextInputV2...). 
Also, those components (the non-deprecated ones) use `Label` now!


packages/ui/src/components/Label/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+ * Label is used inside all of our input components, but it can be used outside for design purposes
+ */
+export const Label = ({
+  children,
+  labelDescription,
+  required,
+  size = 'medium',
+  htmlFor,
+  id,
+  'aria-labelledby': ariaLabelledBy,
```
matthprost, line 25: Not sure why is this here? This is supposed to be for input that need to be labeled using the id of the label

packages/ui/src/components/Label/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+          id={id}
+          variant={size === 'large' ? 'bodyStrong' : 'bodySmallStrong'}
+          sentiment="neutral"
+          htmlFor={htmlFor}
+          prominence="strong"
+          aria-labelledBy={ariaLabelledBy}
+        >
+          {children}
+        </Text>
+        <AsteriskIcon sentiment="danger" size={8} />
```
matthprost, line 41: Can you use a sizing on the icon instead of a value ? Like xsmall or small ?
Sizing with a number is a deprecated behavior

packages/ui/src/components/Label/__tests__/index.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </div>,
+    )
+
+    const input = screen.getByTestId('test-input')
+    const label = screen.getByText('Label')
+
+    await userEvent.click(label)
+    expect(input).toHaveFocus()
+    expect(asFragment()).toMatchSnapshot()
+  })
```
matthprost, line 62: This is great nice job 😄

packages/ui/src/components/Label/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Label is used inside all of our input components, but it can be used outside for design purposes
+ */
+export const Label = ({
+  children,
+  labelDescription,
+  required,
+  size = 'medium',
+  htmlFor,
```
matthprost, line 23:
```suggestion
  size = 'large',
  htmlFor,
```

Better follow the same default value of other inputs, which is `large`

matthprost (changes requested):

packages/ui/src/components/Label/index.tsx
lisalupi: xsmall is too big (12px), we use 8px for every asterisk

matthprost: This needs to be reported to designers then, and maybe we will need to implement a xxsmall size 😄

lisalupi: they don't use icons, only * with text. I can switch to text instead of icon, to match them

alexandre-combemorel (approved):

johnrazeur (approved):

matthprost (approved):

PR #4741, opened 2025-02-06
fix(DateInput): range fixes

lisalupi:
## Type

- Bug

#### What is expected?
`<DateInput />`: 
- rework range logic: it is not possible to choose the end date before the start date, and once both have been selected, clicking on a new date will always reset the range (instead of changing the end value on some conditions)
-  fix styling issue on hover for days in range (and refactor of the code for those elements)


packages/ui/src/components/DateInput/Context.tsx
```diff
@@ -40,6 +40,8 @@ export type ContextProps = {
   startDate?: Date | null
   endDate?: Date | null
   format?: (value?: Date) => string | undefined
+  hoveredDate?: Date | null
+  setHoveredDate: Dispatch<SetStateAction<Date | null>>
```
matthprost, line 44: hmm I think this can be managed only with CSS don't you think?

matthprost (changes requested):

PR #4742, opened 2025-02-07
refactor(icons): remove some legacy icons usages in the lib and improve button

matthprost:
## Type

- Refactor

#### What is expected?

- System Icons from `@ultraviolet/icons` have a change in their sizing. `large` became `medium` and a new `large` sizing has been created (around 24px)
- `<Button />` the prop `icon`, `iconPosition` and `iconSentiment` has been marked as deprecated and will be removed in future major release. You should use the icon component directly in the children of the button:
```tsx
// Before
import { Button } from '@ultraviolet/ui'

<Button icon="pencil" iconPosition="right" iconVariant="outline">
  Edit
</Button>
```

```tsx
// After
import { Button } from '@ultraviolet/ui'
import { PencilOutlineIcon } from '@ultraviolet/icons'

<Button>
  Edit <PencilOutlineIcon />
</Button>
```
- Fix of other legacy usages of icons into the library


alexandre-combemorel (approved):

lisalupi (approved):
Shouldn't we switch the size used in some components (such as DateInput or SubmitErrorAlert) from "large" to "medium" to keep the icon the same size as before ?

matthprost:
> Shouldn't we switch the size used in some components (such as DateInput or SubmitErrorAlert) from "large" to "medium" to keep the icon the same size as before ?

You are totally right I will fix it

PR #4756, opened 2025-02-11
refactor: scripts

lisalupi:
## Type

- Refactor

#### What is expected?

1. Move all scripts to utils/scripts

2. .(m)js => .tsx


utils/scripts/generate-icons-file.tsx
matthprost, line 1: You need to change `COMMENT_HEADER` to add the new path just like you did for other scripts 😄 
And you will need to run the following command "pnpm icons:update"

matthprost (changes requested):

alexandre-combemorel (approved):

matthprost (approved):

PR #4765, opened 2025-02-13
fix(Popup): better compute placement

matthprost:
## Type

- Bug

#### What is expected?

We use to compute popup that has a parent other than `document.body` for a better positioning but there is actually no reasons to do it for `document.body` too.

| url  | Screenshot 2025-02-13 at 17 55 39 | Screenshot 2025-02-13 at 17 55 44 |


packages/ui/src/components/Slider/__stories__/Options.stories.tsx
lisalupi, line 1: isn't it the same as [another PR](https://github.com/scaleway/ultraviolet/pull/4764/files) ?

lisalupi (approved):

johnrazeur (approved):

packages/ui/src/components/Slider/__stories__/Options.stories.tsx
matthprost: I rebased the PR it should be fine now

PR #4780, opened 2025-02-18
fix(Labe): required

lisalupi:
## Type

- Bug

#### What is expected?
"\*" was part of the label and should not

### DIFF
--- a/packages/ui/src/components/Label/index.tsx
+++ b/packages/ui/src/components/Label/index.tsx
@@ -39,8 +39,7 @@ export const Label = ({
           as="span"
           variant={size === 'large' ? 'bodyStrong' : 'bodySmallStrong'}
           sentiment="danger"
-          htmlFor={htmlFor}
-          aria-hidden="true"
+          aria-label="required"
         >
           *
         </Text>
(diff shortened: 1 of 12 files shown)


radhi-nasser-scaleway (approved):

matthprost (changes requested):
label description is also a label

packages/ui/src/components/Label/index.tsx
```diff
@@ -36,10 +36,11 @@ export const Label = ({
           {children}
         </Text>
         <Text
-          as="label"
+          as="span"
           variant={size === 'large' ? 'bodyStrong' : 'bodySmallStrong'}
           sentiment="danger"
           htmlFor={htmlFor}
+          aria-hidden="true"
```
matthprost, line 43:
```suggestion
          variant={size === 'large' ? 'bodyStrong' : 'bodySmallStrong'}
          sentiment="danger"
          aria-label="required"
```

matthprost (changes requested):

fabienhebert (approved):

matthprost (approved):

PR #4847, opened 2025-03-06
fix(SteppedListCard): rename component and add prop completed

lisalupi:
## Type

- Bug

#### What is expected?
- **Breaking change**: `<SteppedListContainer />` is now `<SteppedListCard />`
- Propagate prop `completed` to `SteppedListCard.Step` to choose pre-completed steps


philibea (approved):

packages/plus/src/components/SteppedListCard/__tests__/index.test.tsx
```diff
@@ -3,109 +3,109 @@ import { userEvent } from '@testing-library/user-event'
...
-        <SteppedListContainer.Step
+        <SteppedListCard.Step
           stepNumber={1}
           subHeader="step1title"
           image={<img src={blockStorageWire} width={200} alt="blockStorage" />}
           data-testid="step1id"
         >
           {nextStep => (
             <>
-              step1desc{' '}
```
matthprost, line 101: why removing the space ? Isn't gonna change the layout ?

packages/plus/src/components/SteppedListCard/SteppedListContent.tsx
```diff
@@ -22,15 +22,27 @@ type SteppedListContentProps = {
...
   const containerData = useContext(Data)
 
+  useEffect(() => {
+    containerData.setDone(prevDone => {
+      const updatedDone = [...prevDone]
+      updatedDone[stepNumber - 1] = completed
+
+      return updatedDone
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
matthprost, line 43: using this flag is source of side effects, it doesn't work if you put all the dependencies?

matthprost (changes requested):

matthprost (approved):

PR #4886, opened 2025-03-12
fix(SelectInputV2): change dropdown scroll to auto

matthprost:
## Type

- Bug

#### What is expected?

Fix `SelectInputV2` not to force scroll in the dropdown

| url  | Screenshot 2025-03-12 at 18 21 11 | Screenshot 2025-03-12 at 18 21 16 |

### DIFF
--- a/packages/ui/src/components/SelectInputV2/__stories__/resources.tsx
+++ b/packages/ui/src/components/SelectInputV2/__stories__/resources.tsx
@@ -38,6 +38,37 @@ export const dataUnGrouped = [
     value: 'venus',
     label: 'Venus',
   },
+  {
+    value: 'earth',
+    label: 'Earth',
+    description: 'Our home planet',
+    searchText: 'earth',
+  },
+  {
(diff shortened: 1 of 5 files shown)


fabienhebert (approved):

philibea (approved):

packages/ui/src/components/SelectInputV2/__stories__/resources.tsx
```diff
@@ -38,37 +38,6 @@ export const dataUnGrouped = [
     value: 'venus',
     label: 'Venus',
   },
-  {
```
lisalupi, line 41: why remove all that?

lisalupi (approved):

packages/ui/src/components/SelectInputV2/__stories__/resources.tsx
matthprost: right, it was for testing the scroll I forgot to revert it back to normal

PR #4919, opened 2025-03-20
feat(CodeEditor): various improvements

lisalupi:
## Type
- Enhancement
- Documentation
- Migration
- Refactor

### Summarise concisely:
`<CodeEditor />`: 
- **BREAKING** Rename prop `title` to `label` and change style
- New prop `helper`
- New prop `disabled`
- New prop `copyButton`
- Update component style to match ultraviolet theme (the component always uses Dark Theme colors)

Before : 
Capture d’écran 2025-03-20 à 16 37 55

After :
Capture d’écran 2025-03-20 à 16 38 07

### DIFF
--- a/.changeset/proud-things-live.md
+++ b/.changeset/proud-things-live.md
@@ -1,0 +1,10 @@
+---
+"@ultraviolet/plus": patch
+---
+
+`<CodeEditor />`: 
+- **BREAKING** Rename prop `title` to `label` and change style
+- New prop `helper`
+- New prop `disabled`
+- New prop `copyButton`
+- Update component style to match ultraviolet theme (the component always uses Dark Theme colors)
--- a/.changeset/proud-things-live.md
+++ b/.changeset/proud-things-live.md
@@ -3,7 +3,7 @@
 ---
 
 `<CodeEditor />`: 
-- **BREAKING** Rename prop `title` to `label` and change style
+- Rename prop `title` to `label` and change style
 - New prop `helper`
 - New prop `disabled`
 - New prop `copyButton`
--- a/packages/plus/src/components/CodeEditor/CodeEditor.tsx
+++ b/packages/plus/src/components/CodeEditor/CodeEditor.tsx
@@ -42,6 +42,7 @@ const EditorContainer = styled.div`
 
   &[data-disabled="true"] {
     pointer-events: none;
+    user-select: none;
 
     .cm-editor {
       background-color: ${consoleDarkTheme.colors.neutral.backgroundWeakDisabled};
@@ -61,6 +62,13 @@ const EditorContainer = styled.div`
       display: none;
     }
 
+    .cm-selectionMatch {
+      background-color: transparent;
+    }
+
+    .cm-selectionLayer {
+      display: none;
+    }
   }
 `
 const StyledStack = styled(Stack)`
@@ -131,9 +139,7 @@ export const CodeEditor = ({
 }: CodeEditorProps) => (
   <StyledStack gap={0.5} data-disabled={disabled}>
     {label || title ? (
-      <Label labelDescription={labelDescription} disabled={disabled}>
-        {label ?? title}
-      </Label>
+      <Label labelDescription={labelDescription}>{label ?? title}</Label>
     ) : null}
     <EditorContainer data-disabled={disabled}>
       <CodeMirror
@@ -154,6 +160,11 @@ export const CodeEditor = ({
         aria-label={ariaLabel}
         data-testid={dataTestId}
         className={className}
+        editable={!disabled || !readOnly}
+        aria-disabled={disabled}
+        onUpdate={() => {
+          if (disabled) document.getSelection()?.empty()
+        }}
       />
       {copyButton && !disabled ? (
         <StyledCopyButton
--- a/packages/plus/src/components/CodeEditor/CodeEditor.tsx
+++ b/packages/plus/src/components/CodeEditor/CodeEditor.tsx
@@ -8,16 +8,65 @@ import { consoleDarkTheme } from '@ultraviolet/themes'
 import { CopyButton, Label, Stack, Text } from '@ultraviolet/ui'
 import type { ComponentProps, ReactNode } from 'react'
 
-const StyledText = styled(Text)`
-  background-color: ${({ theme }) => theme.colors.neutral.backgroundStrong};
-  padding: ${({ theme }) => `${theme.space['1']} ${theme.space['2']}`};
-  border-radius: ${({ theme }) => `${theme.radii.default}`};
-  width: 100%;
-`
-
 const EditorContainer = styled.div`
   position: relative;
   width: 100%;
+
+  .cm-editor {
+    font-family: ${({ theme }) => theme.typography.code.fontFamily};
+    font-size:  ${({ theme }) => theme.typography.code.fontSize};
+    background-color: ${consoleDarkTheme.colors.neutral.backgroundWeak};
+    border-radius: ${({ theme }) => theme.space['0.5']};
+  }
+
+  .cm-content {
+    padding-block: ${({ theme }) => theme.space['2']};
+  };
+
+  .cm-gutters {
+    background-color: ${consoleDarkTheme.colors.neutral.backgroundHover};
+  }
+
+  .cm-lineNumbers {
+    padding-left: ${({ theme }) => theme.space['1']};
+  }
+
+  .cm-gutterElement {
+    color: #545454;
+  }
+
+  .cm-scroller {
+    border-radius: ${({ theme }) => theme.space['0.5']};
+  }
+
+
+  &[data-disabled="true"] {
+    pointer-events: none;
+
+    .cm-editor {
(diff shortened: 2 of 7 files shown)


philibea (approved):

mhervouet (approved):

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -2,54 +2,142 @@ import styled from '@emotion/styled'
...
+const EditorContainer = styled.div`
+  position: relative;
   width: 100%;
 `
 
+const StyledCopyButton = styled(CopyButton)`
+  position: absolute;
+  top: ${({ theme }) => theme.space['1']};
+  right: ${({ theme }) => theme.space['1']};
+  z-index: 1;
```
matthprost, line 18: There is a risk while doing it that, if a modal appears on top of it the copy button will be above it. There should be a a better way to do it while not using z-index. For instance, if in the dom you added this element after the one you're trying to show it above, it should work

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -2,54 +2,142 @@ import styled from '@emotion/styled'
...
+`
+
+const StyledStack = styled(Stack, {
+  shouldForwardProp: prop => !['disabled'].includes(prop),
+})<{ disabled: boolean }>`
+  cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'auto')};
+
+  .cm-editor {
+    font-family: 'JetBrains Mono', monospace;
+    font-size: 14px;
```
matthprost, line 36: this could come for the theme

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -2,54 +2,142 @@ import styled from '@emotion/styled'
...
+    background-color: ${({ disabled }) => (disabled ? consoleDarkTheme.colors.neutral.backgroundWeakDisabled : consoleDarkTheme.colors.neutral.backgroundWeak)};
+    border-radius: ${({ theme }) => theme.space['0.5']};
+
+    ${({ disabled }) =>
+      disabled
+        ? `
+      pointer-events: none;
+      color: ${consoleDarkTheme.colors.neutral.textDisabled}`
+        : ''}
```
matthprost, line 45: Maybe we should also add a `cursor: not-allowed;`

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -2,54 +2,142 @@ import styled from '@emotion/styled'
...
+  .cm-gutterElement {
+    color: ${({ disabled }) => (disabled ? consoleDarkTheme.colors.neutral.textWeakDisabled : '#545454')};
+  }
+
+  .cm-scroller {
+    border-radius: ${({ theme }) => theme.space['0.5']};
+  }
+`
 type CodeEditorProps = {
-  title?: string
```
matthprost, line 16: keep it but mark it as deprecated instead. We don't want to have a breaking change (yet)

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -2,54 +2,142 @@ import styled from '@emotion/styled'
...
   height?: string
   readOnly?: boolean
+  disabled?: boolean
+  helper?: ReactNode
+  /**
+   * When set to true, a copy button is displayed in the top right corner of the editor.
+   * If a string is provided, it is used as the button's label. Otherwise, no label is displayed.
+   */
+  copyButton?: boolean | string
+  label?: string
```
matthprost, line 90: maybe also add labelDescription ? So we are same with other inputs

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -2,54 +2,142 @@ import styled from '@emotion/styled'
...
+  svg > path {
+    fill: ${consoleDarkTheme.colors.other.icon.product.original.fill};
+  }
+
+  &:hover {
+    background-color: ${consoleDarkTheme.colors.neutral.backgroundHover};
+  }
+`
+
+const StyledStack = styled(Stack, {
```
matthprost, line 29: Why applying this style on `Stack` an not on `CodeMirror` ?

matthprost (changes requested):

packages/plus/src/components/CodeEditor/CodeEditor.tsx
lisalupi: we have it at line 32

.changeset/proud-things-live.md
```diff
@@ -0,0 +1,10 @@
+---
+"@ultraviolet/plus": patch
+---
+
+`<CodeEditor />`: 
+- **BREAKING** Rename prop `title` to `label` and change style
```
matthprost, line 6: You need to change the changelog according to review feedback

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -30,26 +110,51 @@ export const CodeEditor = ({
...
 }: CodeEditorProps) => (
-  <div>
-    {title ? (
+  <StyledStack gap={0.5} disabled={disabled}>
+    {label ? <Label labelDescription={labelDescription}>{label}</Label> : null}
+    {title && !label ? (
       <StyledText as="h3" variant="headingSmall">
         {title}
       </StyledText>
     ) : null}
```
matthprost, line 126:
```suggestion
    {label || title ? <Label labelDescription={labelDescription}>{label ?? title}</Label> : null}
```
So this way we have the same style whatever prop is set

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -12,14 +13,93 @@ const StyledText = styled(Text)`
...
+  disabled?: boolean
+  helper?: ReactNode
+  /**
+   * When set to true, a copy button is displayed in the top right corner of the editor.
+   * If a string is provided, it is used as the button's label. Otherwise, no label is displayed.
+   */
+  copyButton?: boolean | string
+  label?: string
+  id?: string
+  labelDescription?: ReactNode
```
matthprost, line 102: I think aria-label, data-testid and className are also missing in the props

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -30,26 +110,51 @@ export const CodeEditor = ({
...
+        onChange={onChange}
+        value={value}
+        extensions={[langs[extensions]()]}
+        onBlur={onBlur}
+        basicSetup={{
+          highlightActiveLine: false,
+          highlightActiveLineGutter: false,
+        }}
+        id={id}
+        theme={material}
```
matthprost, line 141: is it normal that I can still edit the code editor in disabled mode ? Also I can click and the cursor is blinking

Screenshot 2025-04-02 at 17 21 06

matthprost (changes requested):

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -2,24 +2,109 @@ import styled from '@emotion/styled'
...
+      border-left-color: transparent;
+      display: none;
+    }
+
+  }
+`
+const StyledStack = styled(Stack)`
+  &[data-disabled] {
+    cursor: not-allowed;
+  }
```
matthprost, line 67: Seems like this doesn't work I have cursor not allowed even when the component is not disabled

matthprost (changes requested):

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -34,27 +119,57 @@ export const CodeEditor = ({
...
-    {title ? (
-      <StyledText as="h3" variant="headingSmall">
-        {title}
-      </StyledText>
+  <StyledStack gap={0.5} data-disabled={disabled}>
+    {label || title ? (
+      <Label labelDescription={labelDescription} disabled={disabled}>
+        {label ?? title}
+      </Label>
+    ) : null}
```
matthprost, line 137: Why is the label light gray in disabled mode? doesn't seem we have that anywhere else
Screenshot 2025-05-05 at 09 18 00

packages/plus/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -34,27 +119,57 @@ export const CodeEditor = ({
   height,
   readOnly,
   autoCompletion,
+  disabled = false,
```
matthprost, line 122: There is still an issue with disabled mode I can select the content, this should be in readonly instead

Screenshot 2025-05-05 at 09 17 44

matthprost (changes requested):

matthprost (approved):

matthprost (commented):
Screenshot 2025-05-05 at 15 11 59

There is a weird outline around the input, if we could use the native or our own one it would be better

PR #4946, opened 2025-03-27
fix: selectableCard height

lisalupi:
## Type

- Bug

#### What is expected?
There was an issue with the height of a SelectableCard with a tooltip : 
- Add prop `containerFullHeight` to `Popup` and `Tooltip`,
- Use it to fix height issue for `SelectableCard`

### DIFF
--- a/packages/ui/src/components/SelectableCard/index.tsx
+++ b/packages/ui/src/components/SelectableCard/index.tsx
@@ -362,7 +362,9 @@ export const SelectableCard = forwardRef(
         if (tooltip) {
           return (
             <Stack flex={1}>
-              <Tooltip text={tooltip}>{subChildren}</Tooltip>
+              <Tooltip text={tooltip} containerFullHeight>
+                {subChildren}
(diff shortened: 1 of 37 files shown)


packages/ui/src/components/SelectableCard/index.tsx
```diff
@@ -352,7 +352,9 @@ export const SelectableCard = forwardRef(
         if (tooltip) {
           return (
             <Stack flex={1}>
-              <Tooltip text={tooltip}>{subChildren}</Tooltip>
+              <Tooltip text={tooltip} containerFullHeight>
+                {subChildren}
+              </Tooltip>
             </Stack>
```
matthprost, line 358: Doesn't it work if instead you do that to the parent of the tooltip? Maybe using `flex="1 1 auto"` might do something.

matthprost (changes requested):

matthprost (approved):

johnrazeur (approved):

PR #4968, opened 2025-04-02
feat(Modal): add image prop

lisalupi:
## Type
- Feature

#### What is expected?

Add prop `image` to modal

### DIFF
--- a/packages/ui/src/components/Modal/__stories__/Image.stories.tsx
+++ b/packages/ui/src/components/Modal/__stories__/Image.stories.tsx
@@ -11,7 +11,6 @@ export const Image: StoryFn = props => (
   <Stack gap={2} direction="row">
     <Modal
       image={image}
-      imageAlt="illustration"
       size="medium"
       disclosure={<Button>Open Modal with image (incorrect ratio)</Button>}
       {...props}
@@ -32,7 +31,6 @@ export const Image: StoryFn = props => (
     </Modal>
     <Modal
       image={imageCorrectFormat}
-      imageAlt="illustration"
       size="xsmall"
       disclosure={<Button>Open Modal with image (correct ratio)</Button>}
       {...props}
@@ -57,8 +55,7 @@ export const Image: StoryFn = props => (
 Image.parameters = {
   docs: {
     description: {
-      story:
-        'Add an image at the top of the modal. Do not forget to add an alternative text to the image using prop `imageAlt` for accessibility.',
+      story: 'Add an image at the top of the modal.',
     },
   },
 }
--- a/packages/ui/src/components/Modal/components/Dialog.tsx
+++ b/packages/ui/src/components/Modal/components/Dialog.tsx
@@ -141,7 +141,6 @@ export const Dialog = ({
   dialogCss,
   backdropCss,
   image,
-  imageAlt,
 }: DialogProps) => {
   const [isVisible, setIsVisible] = useState(false)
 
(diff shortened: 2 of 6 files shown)


packages/ui/src/components/Modal/__stories__/Image.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Modal } from '..'
+import { Badge } from '../../Badge'
+import { Button } from '../../Button'
+import { Stack } from '../../Stack'
+import { Text } from '../../Text'
+
+export const Image: StoryFn = props => (
+  <Modal
+    image={accountExperienceContent as string}
+    disclosure={<Button>Open Modal with lot of content</Button>}
```
matthprost, line 13:
```suggestion
    disclosure={<Button>Open Modal with image</Button>}
```

packages/ui/src/components/Modal/__stories__/Image.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import type { StoryFn } from '@storybook/react'
+// @ts-expect-error can't import ultraviolet/illustration in ui (cyclic dependencies)
+import { accountExperienceContent } from '@ultraviolet/illustrations/products/accountExperience'
```
matthprost, line 3: Better avoid cyclic dependencies. You need to add a folder `assets` at same place as story and add manually an image there. You could also use the same image for unit testing. Check on AvatarV2 how it has been done 😄

packages/ui/src/components/Modal/components/Dialog.tsx
```diff
@@ -8,10 +8,27 @@ import type {
...
 import { MODAL_PLACEMENT, MODAL_WIDTH } from '../constants'
 import type { DialogProps, ModalPlacement, ModalSize } from '../types'
 
+const StyledImage = styled.img`
+  max-height: 13rem;
+ `
+
+const StyledDiv = styled.div`
+  width: 100%;
+  height: 240px;
```
matthprost, line 22: rem ?

packages/ui/src/components/Modal/components/Dialog.tsx
```diff
@@ -8,10 +8,27 @@ import type {
...
+import { Stack } from '../../Stack'
 import { useModal } from '../ModalProvider'
 import { MODAL_PLACEMENT, MODAL_WIDTH } from '../constants'
 import type { DialogProps, ModalPlacement, ModalSize } from '../types'
 
+const StyledImage = styled.img`
+  max-height: 13rem;
+ `
+
+const StyledDiv = styled.div`
```
matthprost, line 20: why not using `Stack` it seems you define the same as a `Stack` does

packages/ui/src/components/Modal/components/Dialog.tsx
```diff
@@ -312,8 +335,18 @@ export const Dialog = ({
...
+        data-has-image={!!image}
       >
-        {children}
+        {image ? (
+          <>
+            <StyledDiv>
+              <StyledImage src={image} alt="promotional image" />
+            </StyledDiv>
+            <StyledStack gap={5}>{children}</StyledStack>
+          </>
```
matthprost, line 346: I know you got the code from the existing one maybe it needs reflexion onto it. I think it might be easier to set the image in CSS with background. 

Do we want to have the image to take the full space or to be centered like you did ?
Might be worth questioning the design about this and adapt the code depending on that. 
Screenshot 2025-04-03 at 10 21 09

The original component was done for a specific use case, here we will try to make it more generic.

matthprost (changes requested):

matthprost (changes requested):

matthprost (approved):

JulienSaguez (approved):

PR #4977, opened 2025-04-04
fix(TimeInput): support for string value

alexandre-combemorel:
## Summary

Timeinput did not support date string as input (such as DateInput).
This MR fix it

## Type

- Bug

#### What is expected?

A user can provide a json date string as value for the input

### DIFF
--- a/packages/ui/src/components/TimeInputV2/index.tsx
+++ b/packages/ui/src/components/TimeInputV2/index.tsx
@@ -157,7 +157,7 @@ padding-inline: ${({ theme }) => theme.space['0.25']};
 
 type TimeInputProps = {
   placeholder?: Time
-  value?: Date
+  value?: Date | string | null
   clearable?: boolean
   required?: boolean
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/TimeInputV2/index.tsx
```diff
@@ -157,7 +157,7 @@ padding-inline: ${({ theme }) => theme.space['0.25']};
 
 type TimeInputProps = {
   placeholder?: Time
-  value?: Date
+  value?: Date | string | null
```
matthprost, line 160: I'm confused why we need to allow string while you could provide `value={new Date(value)}` while calling the component? I prefer if the types are strict and the manipulation is happening outside of the component as a string could be anything including a real date.

matthprost (changes requested):

alexandre-combemorel:
Harmonizing DateInput instead

PR #4988, opened 2025-04-09
fix: navigation and expandable better performances

matthprost:
## Type

- Enhancement

#### What is expected?

After a quick analysis I optimised Navigation and Expandable components.

- Expandable: I separated the logic between animated and not animated component. This way when the component is not animated we don't load all the unnecessary code for animation. This should speed up the render time when having a lot of expandable on a page without animation

- Navigation: Removed expandable animation to optimise the render and memoized component that needed to be. I also fixed WrapText to use only data attribute avoid class recompute on each changes.


.storybook/preview.tsx
```diff
@@ -12,6 +12,13 @@ import Page from './components/Page'
 import { globalStyles } from './components/globalStyle'
 import { dark, light } from './storybookThemes'
 import '@ultraviolet/fonts/fonts.css'
+import { useEffect } from 'react'
+import { scan } from "react-scan"
+
+scan({
```
vincentaudebert, line 18: normal les traces de react scan dans cette MR ?

vincentaudebert (approved):

.storybook/preview.tsx
matthprost: it needs a small rebase let me check

PR #5000, opened 2025-04-12
fix(date): datefield accept null and get clearable option

alexandre-combemorel:
## Type

- Enhancement
-
### Summarise concisely:

Add clearable option to DateInputField and DateInput.
And forward tooltip from DateInput to DateIputField which was missing

#### What is expected?

Being able to clear a DateInput and accepting null or undefined value

| url  | screenshot | Screenshot 2025-04-12 at 18 32 43 |

### DIFF
--- a/packages/form/src/components/DateField/index.tsx
+++ b/packages/form/src/components/DateField/index.tsx
@@ -45,11 +45,7 @@ const parseDate = (value: string | Date): Date =>
   typeof value === 'string' ? new Date(value) : value
 
 const isEmpty = (value?: Date | string | null): boolean => {
-  if (typeof value === 'string') {
-    return value === ''
-  }
-
-  return value === undefined || value === null
+  return !value
 }
 
 export const DateField = <
--- a/packages/form/src/components/DateField/index.tsx
+++ b/packages/form/src/components/DateField/index.tsx
@@ -44,8 +44,13 @@ type DateFieldProps<
 const parseDate = (value: string | Date): Date =>
   typeof value === 'string' ? new Date(value) : value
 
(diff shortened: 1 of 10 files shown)


packages/form/src/components/DateField/index.tsx
```diff
@@ -67,6 +72,8 @@ export const DateField = <
   selectsRange,
   size,
   placeholder,
+  tooltip,
+  clearable,
```
matthprost, line 76: clearable is not part of the specifications where do this come from?

matthprost (changes requested):

packages/form/src/components/DateField/index.tsx
alexandre-combemorel: no where, it is a open source contribution.
And on top of this the field currently can be "cleared" manually but the form still conserve the value, which looks like unexpected behaviour. The question is more, why should this field not be able to be cleared, while TimeInputField can ?

matthprost (approved):

packages/form/src/components/DateField/index.tsx
```diff
@@ -42,8 +42,13 @@ type DateFieldProps<
...
   typeof value === 'string' ? new Date(value) : value
 
-const isEmpty = (value?: Date | string): boolean =>
-  typeof value === 'string' ? value === '' : value === undefined
+const isEmpty = (value?: Date | string | null): boolean => {
+  if (typeof value === 'string') {
+    return value === ''
+  }
+
+  return value === undefined || value === null
```
johnrazeur, line 50:
```suggestion
  return !value
```

johnrazeur (approved):

PR #5053, opened 2025-04-22
fix(VerificationCode): disable auto complete on inputs

matthprost:
## Type

- Bug

#### What is expected?

Remove auto complete on VerificationCode component.


philibea (approved):

BABAK0T0 (approved):

matthprost:
> I was wondering if we should also do it for DateInput, as the autoComplete is overlapping the calendar behind and making it painful to select a date

DateInput already include this https://github.com/scaleway/ultraviolet/blob/f30c1a20a6b6755144b7639ce0b3d0123c15815a/packages/ui/src/components/DateInput/index.tsx#L323

PR #5091, opened 2025-05-02
fix(Modal): nested modals scroll

lisalupi:
## Type

- Bug

### Summarise concisely:
Fix scroll in nested modals


matthprost (approved):
Noice

radhi-nasser-scaleway (approved):

PR #5105, opened 2025-05-06
feat(Icons): new logos

lisalupi:
## Type
- Enhancement

#### What is expected?

New logos: 
- deepseek
- metabase
- tableau
- qwen
- mistral
- power-bi
 
And adapted script to work with the new logos


utils/scripts/generate-icons-file.tsx
```diff
@@ -127,9 +127,10 @@ const readSvg = async (filePath: string, suffix: string) => {
     .replace(/clip-rule=/g, 'clipRule=')
     .replace(/clip-path=/g, 'clipPath=')
     .replace(/stop-color=/g, 'stopColor=')
+    .replace(/flood-opacity=/g, 'floodOpacity=')
+    .replace(/color-interpolation-filters=/g, 'colorInterpolationFilters=')
+    .replace(/xlink:href=/g, 'xlinkHref=')
     .replace(/`/g, '\\`')
-    .replace(/height="[^"]*"/g, '')
-    .replace(/width="[^"]*"/g, '')
```
matthprost, line 132: Why removing it? Some logo doesn't work without it ?

matthprost (changes requested):

utils/scripts/generate-icons-file.tsx
lisalupi: One of them did not work without and it did not change anything for the other ones. But I think it was for the logo we removed so I can put it back (I have to check).

philibea (approved):

PR #5189, opened 2025-05-26
fix(SelectInputV2): overflow

lisalupi:
## Type

- Bug

#### What is expected?
Fix overflow with multiselect.
Now it:
- takes the correct size of the tag (instead of computing an approx. one)
-  adds ellipses instead of hiding tags when possible (ellipsis for the last element when it can be large enough and not overflow) 
-  never adds  a "+1" tag when only one element is selected (will add ellipsis if the element can not be shown completely)

### DIFF
--- a/.changeset/tidy-comics-talk.md
+++ b/.changeset/tidy-comics-talk.md
@@ -1,0 +1,11 @@
+---
+"@ultraviolet/fonts": patch
+"@ultraviolet/form": patch
+"@ultraviolet/icons": patch
+"@ultraviolet/illustrations": patch
+"@ultraviolet/plus": patch
+"@ultraviolet/themes": patch
+"@ultraviolet/ui": patch
+---
+
+`<SelectInputV2 />`: improve overflow
--- a/.changeset/tidy-comics-talk.md
+++ b/.changeset/tidy-comics-talk.md
@@ -1,10 +1,4 @@
 ---
-"@ultraviolet/fonts": patch
-"@ultraviolet/form": patch
-"@ultraviolet/icons": patch
-"@ultraviolet/illustrations": patch
-"@ultraviolet/plus": patch
-"@ultraviolet/themes": patch
 "@ultraviolet/ui": patch
 ---
 
--- a/packages/ui/src/components/SelectInputV2/SelectBar.tsx
+++ b/packages/ui/src/components/SelectInputV2/SelectBar.tsx
@@ -8,7 +8,7 @@ import {
   CloseIcon,
   PlusIcon,
 } from '@ultraviolet/icons'
-import type { RefObject } from 'react'
+import type { ReactNode, RefObject } from 'react'
 import { useEffect, useMemo, useRef, useState } from 'react'
 import { Button } from '../Button'
 import { StyledChildrenContainer } from '../Popup'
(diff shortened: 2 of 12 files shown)


.changeset/tidy-comics-talk.md
```diff
@@ -0,0 +1,11 @@
+---
+"@ultraviolet/fonts": patch
+"@ultraviolet/form": patch
+"@ultraviolet/icons": patch
+"@ultraviolet/illustrations": patch
+"@ultraviolet/plus": patch
+"@ultraviolet/themes": patch
+"@ultraviolet/ui": patch
```
matthprost, line 8: This is a mistake only UI is impacted

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -187,12 +196,17 @@ const DisplayValues = ({
       ref={refTag}
       alignItems="center"
     >
-      {nonOverflowedValues.map(option => (
+      {nonOverflowedValues.map((option, index) => (
```
matthprost, line 199: You need to add a unit test that check that behavior to ensure we are consistent over time

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -54,7 +58,7 @@ const Placeholder = styled(Text)`
 user-select: none;
 `
 
-export const StyledInputWrapper = styled(Stack)<{
+const StyledInputWrapper = styled(Stack)<{
```
matthprost, line 61: There is an overflowing issue, before:
Screenshot 2025-05-27 at 10 27 36

Now:
Screenshot 2025-05-27 at 10 27 50

It's when I resize the screen maybe you need to add an event listener for that. However it shouldn't overflow the select in any cases.

matthprost, line 61: There are some visual glitches if I play with the multi select:

https://github.com/user-attachments/assets/d05a4b18-4088-44e5-9ff1-7cc87ee8a144

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -451,3 +565,5 @@ export const SelectBar = ({
     </Tooltip>
   )
 }
+
+export { SelectBar, StyledInputWrapper }
```
matthprost, line 569: Another issue here also, there is space but somehow it only shows 2 tags fully
Screenshot 2025-05-27 at 10 33 02

matthprost: Screenshot 2025-05-27 at 10 33 51

sometimes it also show them big somehow

packages/ui/src/components/SelectInputV2/SelectBar.tsx
```diff
@@ -295,37 +337,105 @@ export const SelectBar = ({
...
+      else {
+        setOverflow(false)
+        setOverflowAmount(measuredHiddenTags)
+        setNonOverFlowedValues(measuredVisibleTags)
       }
     }
-    setNonOverFlowedValues(computedNonOverflowedValues)
-    setOverflowAmount(computedOverflowAmount)
-  }, [options, selectedData.selectedValues, width])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
matthprost, line 437: Adding this comment means you are probable doing a useEffect on something you shouldn't which can create side effects.

packages/ui/src/components/SelectInputV2/SelectBar.tsx
lisalupi: it is already tested

matthprost (changes requested):
One of the tag is primary I don't know why

Screenshot 2025-06-09 at 10 39 24

matthprost (changes requested):
There is this weird behavior that when there is a label the space is managed in a wrong way. Maybe it is because it's a react node and it's hard to cut properly. Maybe we should say that if the tag needs to be shorten then we remove description / optional info. Wdyt?

https://github.com/user-attachments/assets/08d4ec17-6af3-411d-b7a3-76d4bb118796

matthprost (approved):

PR #5208, opened 2025-06-04
fix(Loader): update component

lisalupi:
## Type

- Enhancement

### Summarise concisely:
`<Loader />`: **breaking changes**
- align size on system icons : now size can only be "xsmall" (0.75rem/12px), "small" (1rem/16px), "medium" (1.25rem/20px), "large" (1.5rem/24px), "xlarge" (2rem/32px), or "*xxlarge*" (3.5rem/56px). By default, `size = "xlarge"` to match the old size.
- remove prop `color`, use `sentiment` instead. By default, `sentiment = primary`.
- remove props `trailColor`, `text`and `strokeWidth`.


packages/ui/src/components/Button/index.tsx
```diff
@@ -306,9 +317,12 @@ export const Button = forwardRef<Element, FinalProps>(
         {isLoading ? (
           <Loader
             active
-            trailColor="transparent"
-            size="1em"
-            color="currentColor"
+            size="small"
+            sentiment={
+              variant === 'filled' && !['black', 'white'].includes(sentiment)
+                ? 'neutral'
```
matthprost, line 323: wouldn't it work better with a sentiment white instead? I feel like the loading is not much visible

matthprost:
to validate with @mhervouet especially the sentiments part 🙏

matthprost (approved):

mhervouet (approved):

PR #5229, opened 2025-06-11
feat: nested menu

lisalupi:
## Type

- Enhancement

### Summarise concisely:
Nested menus : example + add search


radhi-nasser-scaleway (approved):

packages/ui/src/components/MenuV2/__stories__/Nested.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <MenuV2.Item borderless>Power on</MenuV2.Item>
+    <MenuV2
+      disclosure={<MenuV2.Item>SubMenu click</MenuV2.Item>}
+      placement="right"
+      triggerMethod="click"
+    >
+      <MenuV2.Item>hi!</MenuV2.Item>
+    </MenuV2>
+    <MenuV2
+      disclosure={<MenuV2.Item>SubMenu hover</MenuV2.Item>}
```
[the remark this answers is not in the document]
lisalupi: I added a prop `arrowRight` so it can be added correctly

packages/ui/src/components/MenuV2/types.ts
```diff
@@ -63,6 +63,7 @@ export type MenuProps = {
    */
   hideOnClickItem?: boolean
   footer?: ReactNode
+  nested?: boolean
```
matthprost, line 66: Seen with you in pm, we need to avoid this prop and try to make it automatic in popup or it will cause issues when the menu is at the edge of the right of the screen

matthprost (changes requested):

matthprost (changes requested):

packages/ui/src/components/MenuV2/MenuContent.tsx
```diff
@@ -211,7 +216,9 @@ export const Menu = forwardRef(
         dynamicDomRendering={dynamicDomRendering}
         align={align}
       >
-        {finalDisclosure}
+        <DisclosureContext.Provider value>
```
philibea, line 219: Why do you need this context as it's always true ?

packages/ui/src/components/MenuV2/MenuProvider.tsx
```diff
@@ -33,16 +30,20 @@ export const MenuProvider = ({
   children,
   visible = false,
 }: MenuProviderProps) => {
+  const isNested = !!useContext(MenuContext) // If there is no parent Menu, then parentMenu is undefined (we do not use useMenu which will return an error)
```
philibea, line 33: You always have Menu as first children no ?  i don't understand when it's possibly happen ?

lisalupi: When a menu is not inside a menu, there is no provider around it so nested = false
When a menu is nested instide another menu, then there is a provider around the nested menu, so nested = true

PR #5255, opened 2025-06-13
fix(Menu): accessibility

lisalupi:
## Type

- Bug

#### What is expected? 
`MenuV2`: 
- focus on disclosure when closing the menu
- Better navigation with arrow keys


packages/ui/src/components/MenuV2/MenuContent.tsx
```diff
@@ -146,12 +150,13 @@ export const Menu = forwardRef(
     )
 
     useEffect(() => {
-      if (isVisible && searchable) {
+      if (isVisible) {
```
matthprost, line 153: There is a weird behavior, If I click on the menu then I press Tab, only the second item is focused ? Maybe it is because the first one is implicitly focused


https://github.com/user-attachments/assets/793f62ad-58b4-464b-b920-fcb6130f527e

matthprost (changes requested):

packages/ui/src/components/MenuV2/components/Item.tsx
```diff
@@ -167,7 +170,7 @@ const Item = forwardRef<HTMLElement, ItemProps>(
               href={href}
               target={target}
               rel={rel}
-              ref={ref as Ref<HTMLAnchorElement>}
+              ref={computedRef as Ref<HTMLAnchorElement>}
```
matthprost, line 173: On the last deployment, focus doesn't go inside the menu


https://github.com/user-attachments/assets/e60a2a4f-1912-4989-99c9-ebb728b9f3f5

matthprost (changes requested):

packages/ui/src/components/MenuV2/components/Item.tsx
```diff
@@ -142,7 +148,10 @@ const Item = forwardRef<HTMLElement, ItemProps>(
     },
     ref,
   ) => {
-    const { hideOnClickItem, setIsVisible } = useMenu()
+    const { hideOnClickItem, setIsVisible, itemsList } = useMenu()
+
+    const innerRef = useRef<HTMLButtonElement>(null)
+    const computedRef = ref ?? innerRef
```
philibea, line 154: ref is not an HTMLButtonElement that why you need to cast here ?

https://github.com/scaleway/ultraviolet/pull/5255/files#diff-d4c7de53cc545c6b7fe43a827e9d37e900fb41229803b99988b0c67bc8e40dc6R175

matthprost (changes requested):
https://github.com/user-attachments/assets/e3c737bf-3239-4cdb-b48c-d7053910d759

We are loosing the menu when focus is going into the nested one

DorianMaliszewski:
Still an issue that a sent you in MP :/

DorianMaliszewski (approved):

matthprost (approved):

PR #5267, opened 2025-06-17
fix(illustration): wire illustration wrong spread of props

matthprost:
## Type

- Bug

| url  | Screenshot 2025-06-17 at 17 10 42 | Screenshot 2025-06-17 at 17 10 55 |


lisalupi (approved):
I think you could completely delete the prop, it is not needed as far as I know

philibea (approved):

PR #5308, opened 2025-06-30
fix: remove deprecated props Menu, List and Link

lisalupi:
## Type
- Refactor

#### What is expected?
**BREAKING CHANGES**

Deprecated props removed:
- `Link`: prop "sentiment" can only by "primary" or "info" (default value) now
- `List.Cell`: prop "preventClick" removed (it is now default behavior)
- `Menu`: prop "size" and "maxWidth" removed


.changeset/easy-planes-occur.md
```diff
@@ -0,0 +1,10 @@
+---
+"@ultraviolet/ui": major
+---
+
+**BREAKING CHANGES**
+
+Deprecated props removed:
+- `Link`: prop "sentiment" can only by "primary" or "info" (default value) now
+- `List.Cell`: prop "preventClick" removed (it is now default behavior)
+- `Menu`: prop "size" and "maxWidth" removed
```
matthprost, line 10: Maybe some more details
```suggestion
- `Menu`: prop "size" and "maxWidth" removed the size is now automatic
```

matthprost (approved):

PR #5330, opened 2025-07-08
feat: offerList

lisalupi:
## Type

- Feature

#### What is expected?
New component offerList


packages/plus/src/components/OfferList/OfferList.tsx
```diff
@@ -0,0 +1,81 @@
...
+          expandable={expandable && type !== 'radio'}
+          columns={computedColumns}
+          autoCollapse={autoCollapse}
+          onSelectedChange={setSelectedRows}
+          selectable={type === 'checkbox'}
+        >
+          {children}
+        </List>
+      </OfferListProvider>
+    </StyledTable>
```
matthprost, line 76: why doing StyledTable and not on List directly?

matthprost: is the className correctly set on those components?

packages/plus/src/components/OfferList/components/Row.tsx
```diff
@@ -0,0 +1,344 @@
...
+`
+
+type RowProps = ComponentProps<typeof List.Row> & {
+  banner?: {
+    text: ReactNode
+    sentiment?: 'neutral' | 'primary' | 'warning' | 'danger'
+  }
+  offerName: string
+  badge?: {
+    text: string
```
matthprost, line 90: I have a feeling that we will have icons and more complex stuff in those badges 😅

matthprost (changes requested):

.changeset/whole-rules-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/plus": patch
```
matthprost, line 2:
```suggestion
"@ultraviolet/plus": minor
```

We release a new component, patch doesn't fit I think

matthprost (approved):

MathildeJolly (approved):

mhervouet (approved):

PR #5335, opened 2025-07-08
fix(SwitchButton): size when children changes

lisalupi:
## Type

- Bug

### Summarise concisely:
`SwitchButton`: size should adapt when the children changes


packages/ui/src/components/SwitchButton/index.tsx
```diff
@@ -98,7 +98,7 @@ export const SwitchButton = ({
     setLocalValue(value)
     setWidth(getElement(value)?.offsetWidth ?? 0)
     setPosition(getPosition(getElement(value)))
-  }, [getElement, refOptions, value])
+  }, [getElement, refOptions, value, children])
```
fabienhebert, line 101: I tried this solution, since children are react node useEffect will be triggered each render, because children comparaison will make old/current props always different, making useEffect/memoization pointless

matthprost (approved):

PR #5385, opened 2025-07-25
fix(Menu): overflow

lisalupi:
## Type

- Bug

#### What is expected?
`Menu`: the popup should shrink when there is not enough size to avoid creating more scroll


packages/ui/src/components/Menu/types.ts
```diff
@@ -52,4 +52,8 @@ export type MenuProps = {
   hideOnClickItem?: boolean
   footer?: ReactNode
   placement?: Exclude<ComponentProps<typeof Popup>['placement'], 'nested-menu'>
+  /**
+   * When set to true, the menu does not shrink (height) to avoid overflow on the page
+   */
+  noShrink?: boolean
```
matthprost, line 58: To check with @mhervouet if this is the behavior we want. It might not be obvious for the user that the menu is scrollable.

matthprost (changes requested):

matthprost (approved):

PR #5391, opened 2025-07-28
fix(Modal): remove id on modal and select input scrolll

matthprost:
## Type

- Bug

#### What is expected?

Removing the id on modal backdrop and use the context with ref instead.

### DIFF
--- a/packages/ui/src/components/SelectInput/Dropdown.tsx
+++ b/packages/ui/src/components/SelectInput/Dropdown.tsx
@@ -10,8 +10,16 @@ import type {
   RefObject,
   SetStateAction,
 } from 'react'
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/SelectInput/Dropdown.tsx
```diff
@@ -12,6 +19,7 @@ import type {
   SetStateAction,
 } from 'react'
 import { Checkbox } from '../Checkbox'
+import { ModalContext } from '../Modal/ModalProvider'
```
philibea, line 22: you can import useModal from the provider ?

lisalupi (approved):

PR #5491, opened 2025-09-02
fix(SelectInput): click oustide should work and size fix

lisalupi:
## Type

- Bug

#### What is expected?
`SelectInput`: 
- click outside should work in every context 
- fix placeholder alignment when small

Before: 
Capture d’écran 2025-09-02 à 11 13 03

After: 
Capture d’écran 2025-09-02 à 11 12 49

### DIFF
--- a/.changeset/sharp-steaks-nail.md
+++ b/.changeset/sharp-steaks-nail.md
@@ -1,0 +1,7 @@
+---
+"@ultraviolet/ui": patch
+---
+
+`SelectInput`: 
+- click outside should work in every context 
+- fix placeholder alignment when small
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/SelectInput/Dropdown.tsx
```diff
@@ -723,6 +723,29 @@ export const Dropdown = ({
...
+  )
+
+  useEffect(() => {
+    document.addEventListener('mouseup', handleClickOutside)
+
+    return () => {
+      document.removeEventListener('mouseup', handleClickOutside)
+    }
+  }, [handleClickOutside])
+
```
matthprost, line 748: This is already handled into the popup component it makes a duplicate. Maybe the issue is related to popup?

matthprost (changes requested):

vincentaudebert (approved):

PR #5492, opened 2025-09-02
refactor(Stack): vanilla extract

matthprost:
## Type

- Refactor

#### What is expected?

Refactor stack using vanilla extract


lisalupi (approved):

packages/ui/src/components/Stack/__stories__/Template.stories.tsx
```diff
@@ -3,7 +3,7 @@ import { Stack } from '..'
 import { DivWithBackground } from './DivWithBackground'
```
lisalupi, line 3: `DivWithBackground` is a styled div. Maybe we should re-do it with vanilla extract?

PR #5503, opened 2025-09-04
refactor(row): vanilla extract

matthprost:
## Type

- Refactor

#### What is expected?

Migrated Row using vanilla extract.

Note: there cannot be a dynamic object with vanilla extract, so I had to flatten the variable for responsive. This is why there is `templateColumnsLarge`, `templateColumnsMedium`, ect.

### DIFF
--- a/packages/ui/src/components/Row/styles.css.ts
+++ b/packages/ui/src/components/Row/styles.css.ts
@@ -1,7 +1,8 @@
 import { createSprinkles, defineProperties } from '@vanilla-extract/sprinkles'
 import { consoleLightTheme } from '@ultraviolet/themes'
 import { style } from '@vanilla-extract/css'
-import { templateColumn, paddings } from './variables.css'
+import { paddings, templateColumn } from './variables.css'
+import type { CSSProperties } from 'react'
 
 export const row = style({
   display: 'grid',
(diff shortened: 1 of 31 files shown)


packages/ui/src/components/Row/styles.css.ts
```diff
@@ -0,0 +1,117 @@
...
+    [key]: {
+      '@media': `screen and (min-width: ${consoleLightTheme.breakpoints[key as keyof typeof consoleLightTheme.breakpoints]})`,
+    },
+  }),
+  {},
+) as Record<keyof typeof consoleLightTheme.breakpoints, { '@media': string }>
+
+const themeSpace = Object.values(consoleLightTheme.space)
+
+const alignItemsValues = [
```
philibea, line 70: should be based on CCSProperties ?

packages/ui/src/components/Row/styles.css.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export const responsiveProperties = defineProperties({
+  conditions: themeBreakpoints,
+  defaultCondition: 'xxsmall',
+  properties: {
+    gap: themeSpace,
+    alignItems: alignItemsValues,
+    justifyContent: justifyContentValues,
+  },
+  responsiveArray: ['xxsmall', 'xsmall', 'small', 'medium', 'large'],
```
philibea, line 114: should be based on the theme ?

PR #5518, opened 2025-09-09
refactor: progress bar to vanilla and fix badge

matthprost:
## Type

- Refactor

#### What is expected?

Refactor progress bar to use vanilla extract and fix badge useless variable within css.


packages/ui/src/components/ProgressBar/variables.css.ts
```diff
@@ -0,0 +1,4 @@
+import { createVar } from '@vanilla-extract/css'
+
+export const sentiment = createVar()
```
lisalupi, line 3: I don't think you set its value anywhere?

PR #5586, opened 2025-09-24
fix(SelectInput): html tag options

lisalupi:
## Type

- Bug

#### What is expected?
`SelectInput`: placeholder should be a `span` instead of a `p`


matthprost (approved):
We need the validation from @iManu for this one 😄

radhi-nasser-scaleway (approved):

iManu (approved):
Merci !

PR #5634, opened 2025-10-03
refactor(SwitchButton): vanilla extract

lisalupi:
## Type

- Refactor

#### What is expected?
Refactor component `SwitchButton` to use vanilla extract instead of Emotion

### DIFF
--- a/packages/ui/src/components/SwitchButton/styles.css.ts
+++ b/packages/ui/src/components/SwitchButton/styles.css.ts
@@ -1,0 +1,106 @@
+import { theme } from '@ultraviolet/themes'
+import { globalStyle, style, styleVariants } from '@vanilla-extract/css'
+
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/SwitchButton/styles.css.ts
```diff
@@ -0,0 +1,106 @@
+import { theme } from '@ultraviolet/themes'
+import { globalStyle, style, styleVariants } from '@vanilla-extract/css'
+
+const SIZES = {
+  medium: '600',
+  small: '500', // sizing token from theme
+} as const
```
matthprost, line 7: not sure if this is actually needed as we use it only in one place

matthprost (approved):

fabienhebert (approved):

PR #6276, opened 2026-03-26
feat(component): add phone input

philibea:
## Type
- Feature

#### What is expected?

Adds a new PhoneInput component across the monorepo packages:

-  packages/ui: New PhoneInput component with stories, tests, and styling
-  packages/form: New PhoneInputField component (form integration) with stories and tests  
-  packages/utils: Phone number parsing utilities

|   Before   |      After |
| :--------: | ---------: |
|  Capture d’écran 2026-05-21 à 12 02 53 | Capture d’écran 2026-05-21 à 11 58 04 |
| | Capture d’écran 2026-05-21 à 11 58 15 |
| | Capture d’écran 2026-05-21 à 11 58 10 |


packages/utils/src/phone.ts
```diff
@@ -0,0 +1,111 @@
...
+const validatePhoneNumber = (phoneNumber: string, options?: { regionCode?: string }): boolean => {
+  try {
+    const parsed = parsePhoneNumber(phoneNumber, options)
+    return parsed.valid || false
+  } catch {
+    return false
+  }
+}
+
+export const phoneUtils = {
```
chambo-e: Why in a const ? 

This means it won't be able to be treeshaked no ?

packages/ui/src/components/PhoneInput/__stories__/UserControl.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const [metadata, setMetadata] = useState<string>('')
+
+  return (
+    <div style={{ display: 'flex', flexDirection: 'column', gap: '16px' }}>
+      <PhoneInput
+        defaultCountry="FR"
+        label="Phone Number (with metadata)"
+        name="phone-metadata"
+        onValueChange={data => {
+          setMetadata(JSON.stringify(data, null, 2))
```
jsulpis, line 55: the value is formatted only when the number is entirely typed, is that normal ?

Image

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+  })
+
+  it('should control the value', () => {
+    const onChange = vi.fn()
+
+    const { asFragment } = renderWithTheme(
+      <PhoneInput label="Phone" name="phone" onChange={onChange} value="+33612345678" />,
+    )
+
+    const input = screen.getByRole('textbox') as HTMLInputElement
```
jsulpis: to make sure the label is correctly attached to the input

```suggestion
    const input = screen.getByRole('textbox', { name: "Phone" }) as HTMLInputElement
```

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+    expect(input.value).toBe('+33612345678')
+    fireEvent.change(input, { target: { value: '+33678901234' } })
```
jsulpis: I think it would be better to type into the field using `userEvent.type()` instead of firing the event artificially

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+    expect(onChange).toHaveBeenCalled()
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  it('should render correctly when input is disabled', () => {
+    const { asFragment } = renderWithTheme(
+      <PhoneInput disabled label="Phone" name="phone" onChange={() => {}} value="+33612345678" />,
+    )
+    expect(asFragment()).toMatchSnapshot()
```
jsulpis, line 41:
```suggestion
    expect(asFragment()).toMatchSnapshot()
    expect(screen.getByRole('textbox')).toBeDisabled()
```

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+  it('should display error message', () => {
+    const onChange = vi.fn()
+    const errorMessage = 'Invalid phone number format'
+
+    renderWithTheme(
+      <PhoneInput error={errorMessage} label="Phone" name="phone" onChange={onChange} value="+33612345678" />,
+    )
+
+    expect(screen.getByText(errorMessage)).toBeDefined()
```
jsulpis: also, the previous test could be merged into this one ?

```suggestion
    expect(screen.getByRole("textbox")).toBeInvalid()
    expect(screen.getByRole("textbox")).toHaveAccessibleDescription(errorMessage)
```

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+  })
+
+  it('should render correctly with required field', () => {
+    const { asFragment } = renderWithTheme(
+      <PhoneInput label="Phone" name="phone" onChange={() => {}} required value="+33612345678" />,
+    )
+
+    expect(asFragment()).toMatchSnapshot()
```
jsulpis:
```suggestion
    expect(asFragment()).toMatchSnapshot()
    expect(screen.getByRole("textbox")).toBeRequired()
```

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+  it('should display asterisk for required field', () => {
+    const onChange = vi.fn()
+
+    renderWithTheme(<PhoneInput label="Phone" name="phone" onChange={onChange} required value="+33612345678" />)
+
+    expect(screen.getByText('*')).toBeDefined()
+  })
+
+  it('should use default country', () => {
```
jsulpis: (this is what the test checks)

```suggestion
  it('should have a placeholder', () => {
```

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+        label="Phone"
+        name="phone"
+        onChange={onChange}
+        placeholder="Enter phone number"
+      />,
+    )
+
+    const input = screen.getByPlaceholderText('Enter phone number')
+    await userEvent.type(input, '0612345678')
+    expect(onChange).toHaveBeenCalled()
```
jsulpis: I guess ?

```suggestion
    expect(onChange).toHaveBeenCalledWith("06 12 34 56 78")
```

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+    const callArg = onValueChange.mock.calls[0]?.[0]
+    expect(callArg).toHaveProperty('inputValue')
+    expect(callArg).toHaveProperty('formatted')
+    expect(callArg).toHaveProperty('country')
+    expect(callArg).toHaveProperty('valid')
+    expect(callArg).toHaveProperty('e164')
+    expect(callArg).toHaveProperty('international')
+  })
+
+  it('should call onValueChange with metadata', async () => {
```
jsulpis: this test looks like it's identical to the previous one ?

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+    await userEvent.type(input, '0612345678')
+
+    expect(onValueChange).toHaveBeenCalled()
+    const callArg = onValueChange.mock.calls[0]?.[0]
+    expect(callArg).toHaveProperty('inputValue')
+    expect(callArg).toHaveProperty('formatted')
+    expect(callArg).toHaveProperty('country')
+    expect(callArg).toHaveProperty('valid')
+    expect(callArg).toHaveProperty('e164')
+    expect(callArg).toHaveProperty('international')
```
jsulpis: could we check the whole object with `.toEqual()` ?

packages/ui/src/components/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+  'aria-label': ariaLabel,
+  'aria-live': ariaLive,
+  'aria-atomic': ariaAtomic,
+  className,
+  defaultCountry = 'FR',
+  disableAutoFormat = false,
+  disabled = false,
+  error: customError,
+  id,
+  label = 'Phone',
```
jsulpis: do we really want a default label ?

packages/ui/src/components/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+  tooltip,
+  name,
+  size = 'large',
+  onBlur,
+  onChange,
+  onFocus,
+  onParsingError,
+  onValueChange,
+  placeholder,
+  ref,
```
jsulpis: `ref` as a prop is a React 19 feature, we should use `forwardRef` if we support React 18.

Note: maybe we should also switch to React 18 in the repo in order to make sure we don't use React 19 features

packages/ui/src/components/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+        setCountryFlag(country)
+      } else {
+        setCountryFlag(defaultCountry)
+        if (phoneNumber.length === 10) {
+          phoneNumber = `+33${phoneNumber}`
+        }
+      }
+
+      const isValid =
+        phoneNumber.length > 4 && phoneUtils.parsePhoneNumber(phoneNumber, { regionCode: countryFlag }).valid
```
jsulpis: a variable could be used to avoid calling `phoneUtils.parsePhoneNumber()` 3 times

packages/ui/src/components/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+        valid: isValid,
+        e164: e164 ?? null,
+        international: formattedNumber ?? null,
+      }
+
+      onValueChange?.(result)
+
+      if (disableAutoFormat) {
+        return inputValue
+      }
```
jsulpis: could be moved at the top of the function to avoid all the parsing if formatting is disabled

packages/ui/src/components/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+      return inputValue
+    }
+  }
+
+  const localId = id || `phone-${name}`
+  const helperId = useId()
+
+  return (
+    <Stack aria-atomic={ariaAtomic} aria-live={ariaLive} className={className} gap={0.5} role={role}>
```
jsulpis: Why would we want to override the role of this element ? I'm also not sure about aria-atomic and aria-live attributes

packages/ui/src/components/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+      {label ? (
+        <Label htmlFor={id ?? localId} id={ariaLabelledBy} required={required} size={size}>
```
jsulpis: I guess the purpose of the `ariaLabelledBy` prop is to link the input to a label external to the component, in which case the id should not be used on this internal Label component (which is already linked to the input with the `for` attribute)

```suggestion
        <Label htmlFor={id ?? localId} required={required} size={size}>
```

packages/ui/src/components/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+          </Stack>
+          <input
+            aria-invalid={!!error}
+            aria-label={ariaLabel}
+            aria-describedby={ariaDescribedBy || (hasHelperText(helper, error, success) ? helperId : undefined)}
+            aria-labelledby={ariaLabelledBy}
+            className={phoneInputStyle.input}
+            data-testid={dataTestId}
+            data-size={size}
+            disabled={disabled}
```
jsulpis: missing attribute 

```suggestion
            disabled={disabled}
            required={required}
```

packages/ui/src/components/PhoneInput/styles.css.ts
```diff
@@ -0,0 +1,110 @@
...
+    },
+    '&[data-error="true"]': {
+      borderColor: theme.colors.danger.border,
+      color: theme.colors.danger.text,
+    },
+    '&:focus-within': {
+      border: `1px solid ${theme.colors.primary.border}`,
+      boxShadow: theme.shadows.focusPrimary,
+    },
+    "&[data-readonly='true']": {
```
jsulpis, line 56: there is no `readonly` prop nor `data-readonly` attribute

packages/ui/src/components/PhoneInput/styles.css.ts
```diff
@@ -0,0 +1,110 @@
...
+      background: theme.colors.neutral.backgroundWeak,
+      borderColor: theme.colors.neutral.border,
+    },
+    "&[data-success='true']": {
```
jsulpis: there is a `success` prop but no `data-success` attribute

packages/ui/src/components/PhoneInput/styles.css.ts
```diff
@@ -0,0 +1,110 @@
...
+    },
+    '&:focus': {
+      outline: 'none',
+    },
+    [`${inputWrapper} > &`]: {
+      color: theme.colors.neutral.text,
+    },
+    [`${inputWrapper} > &::placeholder`]: {
+      color: theme.colors.neutral.textWeak,
+    },
```
jsulpis: seems redundant with other styles applied above

```suggestion
```

packages/ui/src/components/TextInput/styles.css.ts
```diff
@@ -97,7 +97,7 @@ const input = style({
   paddingLeft: theme.space['2'],
   background: 'transparent',
   fontSize: theme.typography.bodySmall.fontSize,
-
+  color: theme.colors.neutral.text,
```
jsulpis, line 100: redundant with the color l.119 (maybe delete the l.119)

Image

packages/ui/src/components/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,213 @@
...
+            id={localId}
+            maxLength={20}
+            name={name}
+            onBlur={onBlur}
+            onChange={event => {
+              if (inputRef.current) {
+                inputRef.current.value =
+                  formatNumber({
```
jsulpis: I think the initial value should also be formatted in addition to onChange

Image

packages/form/src/components/PhoneInputField/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+        onChange?.(event)
+      }}
+      onFocus={event => {
+        onFocus?.(event)
+      }}
+      onParsingError={onParsingError}
+      placeholder={placeholder}
+      ref={field.ref}
+      required={required}
+      value={field.value}
```
jsulpis, line 106: it's missing quite a lot of props of the PhoneInput component: aria attributes, tooltip, helper, disableAutoFormat... missing `{...props}` forwarding ?

packages/form/src/components/PhoneInputField/__tests__/index.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+      },
+      {
+        errors: mockFormErrors,
+        onSubmit: value => {
+          onSubmit(value)
+        },
+      },
+    )
+
+    await userEvent.click(screen.getByText('Submit'))
```
jsulpis: just to encourage good habits 😇 (idem in the rest of the file)

```suggestion
    await userEvent.click(screen.getByRole('button', { name: 'Submit' }))
```

packages/utils/src/phone.ts
chambo-e: Also this should be in scw-lib no ? We never use directly @ultraviolet/utils in our projects

packages/form/src/components/PhoneInputField/__tests__/index.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+      },
+    )
+
+    const phoneInput = screen.getByRole('textbox')
+    await userEvent.type(phoneInput, 'invalid')
+    await userEvent.click(screen.getByText('Submit'))
+    await waitFor(() => {
+      expect(onSubmit).toHaveBeenCalledTimes(0)
+    })
+    expect(screen.getByText("This doesn't appear to be a valid phone number.")).toBeDefined()
```
jsulpis:
```suggestion
    expect(screen.getByRole('textbox')).toHaveAccessibleDescription("This doesn't appear to be a valid phone number.")
```

packages/form/src/components/PhoneInputField/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+                  1,
+                )}
+              </Snippet>
+            </Stack>
+          </Stack>
+        </Form>
+      )
+    },
+  ],
+  title: 'Form/Components/Fields/PhoneField',
```
jsulpis:
```suggestion
  title: 'Form/Components/Fields/PhoneInputField',
```

packages/ui/src/components/PhoneInput/__stories__/UserControl.stories.tsx
philibea: The shortest valid international phone numbers (with country code) are usually about 7 digits long. 

The longest allowed number, according to the standard, is 15 digits, that includes the country code, but not the + sign.

So until you have type 7number you can't guess the country and apply the right format

packages/utils/src/phone.ts
philibea: > Why in a const ?
> 
> This means it won't be able to be treeshaked no ?

true, but it's a smill library so it's was made on purpose to expose under the same object for utils.


> Also this should be in scw-lib no ? We never use directly @ultraviolet/utils in our projects

Yes we can make a lib on scw-lib, it's 95% use only by the PhoneInput, the only thing we use on the console is `parsePhoneNumber`.  Maybe we can expose this utils with another pattern as a hook inside ultraviolet/ui if you prefer ?

packages/form/package.json
```diff
@@ -51,14 +51,16 @@
     "watch:build": "vite build  --watch"
   },
   "dependencies": {
+    "@scaleway/phonenumber": "1.0.0",
     "@ultraviolet/icons": "workspace:*",
     "@ultraviolet/themes": "workspace:*",
     "@ultraviolet/ui": "workspace:*",
+    "@ultraviolet/utils": "workspace:*",
+    "awesome-phonenumber": "7.8.0",
```
jsulpis: included in `@scaleway/phonenumber` 

```suggestion
```

packages/form/package.json
```diff
@@ -51,14 +51,16 @@
     "watch:build": "vite build  --watch"
   },
   "dependencies": {
+    "@scaleway/phonenumber": "1.0.0",
     "@ultraviolet/icons": "workspace:*",
     "@ultraviolet/themes": "workspace:*",
     "@ultraviolet/ui": "workspace:*",
+    "@ultraviolet/utils": "workspace:*",
```
jsulpis: I don't see this package imported, is it used ?

packages/ui/src/components/PhoneInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+    const callArg = onValueChange.mock.calls[0]?.[0]
+    expect(callArg).toStrictEqual({
+      country: 'FR',
+      e164: null,
+      formatted: '0',
+      inputValue: '0',
+      international: '0',
+      valid: false,
+    })
+  })
```
jsulpis, line 197: the value looks wrong, maybe what you wanted is something like that:

```suggestion
    expect(onValueChange).toHaveBeenLastCalledWith(expect.objectContaining({
      country: 'FR',
      e164: null,
      formatted: '0612345678',
      inputValue: '0612345678',
      international: '0612345678',
      valid: false,
    })
    )
```

packages/ui/src/components/PhoneInput/PhoneInput.tsx
```diff
@@ -0,0 +1,222 @@
...
+          localRef.current.value =
+            formatPhoneNumber({
+              inputValue,
+            }) ?? ''
+        }
+      },
+      [formatPhoneNumber],
+    )
+
+    useEffect(() => {
```
jsulpis: THIS would be a valid use case for a `useLayoutEffect`, to avoid painting the unformatted value 😊

edit: even better if we can format the value on the first render, if it's not too complicated

jsulpis (approved):

chambo-e (approved):

PR #6278, opened 2026-03-27
feat(OrderSummary): compact version

lisalupi:
## Type

- Enhancement

#### What is expected?

`OrderSummary`: 
- new prop `compact` for a very compact view of the component, similar to `EstimateCost`,
- new prop `backgroundProminence` to be able to change the component background prominence 
- new prop `calculartorIcon` to add next to the price total

`EstimateCost`: new prop `backgroundProminence`, similar to `OrderSummary`


vincentaudebert (dismissed):

packages/ui/src/compositions/OrderSummary/styles.css.ts
```diff
@@ -106,6 +115,11 @@ const totalPrice = styleVariants({
   default: {},
 })
 
+const calculatorIcon = style({
+  marginRight: theme.space[1],
+  marginBottom: theme.space['0.5'],
```
jsulpis: the icon is not properly aligned with the text, like on the EstimateCost component that uses a flex wrapper. Also in EstimateCost there is no ":" after the label, could it be removed ? 

And do we have a Figma design for this variant ?

Image

lisalupi: i'll fix it

jsulpis (dismissed):

packages/ui/src/compositions/OrderSummary/types.ts
```diff
@@ -23,6 +23,23 @@ type PeriodProps =
       unitUnitInput?: TimeUnit
       onChangeUnitInput?: (unit: string, amount: number) => void
     }
+
+type VariantProps =
+  | {
+      variant: 'compact'
```
jsulpis: in EstimateCost there is a boolean prop "compact", could we do the same here ?

packages/ui/src/compositions/OrderSummary/styles.css.ts
```diff
@@ -2,15 +2,29 @@ import { theme } from '@ultraviolet/themes'
...
+    height: '100%',
+    minWidth: '20rem',
+  },
+  variants: {
+    backgroundProminence: {
+      default: {
+        backgroundColor: theme.colors.neutral.backgroundWeak,
+      },
+      strong: {
+        backgroundColor: theme.colors.neutral.background,
```
jsulpis, line 21: wouldn't it be more consistent that the backgroundProminence values be weak/default instead of default/strong ? or maybe weak/normal to avoid a confusion with "default" not being the default value...

lisalupi: I did that to avoid having "weak" as the default value instead of "default", but I can change to weak/normal if you prefer 👍🏼

jsulpis: ok that makes sense too, you can keep it like that

packages/ui/src/compositions/OrderSummary/types.ts
```diff
@@ -23,6 +23,23 @@ type PeriodProps =
       unitUnitInput?: TimeUnit
       onChangeUnitInput?: (unit: string, amount: number) => void
     }
+
+type CompactProps =
+  | {
+      compact: true
+      /**
+       * This prop is only usable when `variant = "compact"`
```
jsulpis:
```suggestion
       * This prop is only usable when `compact = true`
```

packages/ui/src/compositions/OrderSummary/types.ts
```diff
@@ -23,6 +23,23 @@ type PeriodProps =
...
+       */
+      backgroundProminence?: 'strong' | 'default'
+    }
+  | {
+      compact?: false
+      /**
+       * This prop is only usable when `variant = "compact"`
```
jsulpis:
```suggestion
       * This prop is only usable when `compact = true`
```

packages/ui/src/compositions/OrderSummary/__stories__/Compact.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  calculatorIcon: true,
+  compact: true,
+  backgroundProminence: 'strong',
+}
+
+Compact.parameters = {
+  docs: {
+    description: {
+      story:
+        'With `compact={true}`, you can display a small version of the component, showing only the total price. With that variant, it is possible to change the background prominence using prop `backgroundProminence` ("weak" by default).',
```
jsulpis:
```suggestion
        'With `compact={true}`, you can display a small version of the component, showing only the total price. With that variant, it is possible to change the background prominence using prop `backgroundProminence`.',
```

jsulpis (approved):

MathildeJolly (approved):

PR #6340, opened 2026-04-15
fix(Alert): padding and icon

lisalupi:
## Type

- Bug

#### What is expected?
`Alert`: fix padding and icons

### DIFF
--- a/.changeset/floppy-carpets-talk.md
+++ b/.changeset/floppy-carpets-talk.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/ui": patch
+---
+
+`Alert`: fix padding and icons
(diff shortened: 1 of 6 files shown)


Echo-IV (approved):

mhervouet (approved):

jsulpis (commented):
- in medium size the icon size should be 20px, not 24px

Image

- the close button is always neutral, it should have the sentiment of the Alert
- the close button is bigger than on Figma but on Figma it's not a button so I'm not sure how to implement that 

Image

jsulpis (changes requested):

jsulpis (changes requested):
the icon is too high, you can add a margin-top of 2px like on the small size 

Image

jsulpis (approved):

PR #6360, opened 2026-04-21
ci: update steps definitions and add pnpm cache

jsulpis:
## Type

- Refactor

### Summarize concisely:

Proposal to improve the readability of the CI.

#### What is expected?

By separating the pnpm commands in different steps, we can more easily :
- see the different steps in the UI
- see their individual duration
- browse the logs of a specific script (i.e. see the logs of the build without going through the logs of the install command)

- separate pnpm commands in different steps
- remove node-version matrix because we only use one version
- add pnpm cache where it was not specified
- edit the conditions to trigger the CI to avoid trigger when updating the PR title or description

| GitHub UI  | image | image |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
   build:
     strategy:
       matrix:
-        node: ['24']
+        node: ['24', '25']
     runs-on: ubuntu-24.04
     env:
       TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -3,17 +3,13 @@ name: CI
 on:
   merge_group:
   pull_request:
-    types: ['opened', 'edited', 'reopened', 'synchronize']
```
jsulpis, line 6: `['opened', 'reopened', 'synchronize']` is the default value

.github/workflows/ci.yml
```diff
@@ -3,17 +3,13 @@ name: CI
...
   push:
     branches:
       - main
       - beta
 
 jobs:
   build:
-    strategy:
```
chambo-e, line 14: We had the policy of supporting the current active LTS + the current node release

Don't know if that changed @philibea ?

jsulpis: the version 22 was removed 4 months ago in a [dependency update](https://github.com/scaleway/ultraviolet/pull/5922), I don't know why. I can add the current version (25) back in the matrix

philibea: We can still support a matrice yes.

philibea (approved):

lisalupi (approved):

PR #6365, opened 2026-04-22
ci: optimize the docker build to have a smaller image

jsulpis:
## Type

- Enhancement

### Summarize concisely:

Use a multi-stage build for the docker image in order to avoid sending all the source code and node_modules in the image.

#### What is expected?

- smaller docker image
- faster deploy

Use multi-stage build

| Metric |   Before   |      After |
| image size  | 3,43 GB | 253 MB |
| deploy duration  | 3-7min | 2min43 |

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:24.15.0-alpine
+FROM node:24.15.0-alpine AS builder
 WORKDIR /build
 
 ARG TURBO_TOKEN=token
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -12,6 +12,11 @@ RUN corepack enable
...
 
+FROM node:24.15.0-alpine AS runner
+WORKDIR /app
+
+COPY --from=builder /build/storybook-static ./storybook-static
+
 EXPOSE 80/tcp
 
-CMD pnpm http-server ./storybook-static -p 80
+CMD npx http-server@14.1.1 ./storybook-static -p 80
```
chambo-e, line 22: The only downside here is it's going to be a pain in the ass to manage this package version 😬

chambo-e (approved):
I have no idea why we didn't do this from the start, thanks 🫶

Dockerfile
jsulpis: yes we will try to think about updating it every once in a while 😅

lisalupi (approved):

PR #6367, opened 2026-04-23
feat(Menu.Item): add hideOnClickItem props

PierreBertinet:
## Type

- Feature

### Summarize concisely:

Propal : Add hideOnClickItem on Menu.Item props in addition to Menu props to be able to specify which item close the menu and which does not

#### What is expected?

hideOnClickItem on Menu.Item close the menu on click on the specific item even if the Menu does not have hideOnClickItem set to true

1. Added hideOnClickItem on Menu.Item

#### Why is that needed?

I need this feature to be able to close the menu when a particular element is clicked but not when other are clicked. For exemple:  I have a modal setup with a disclosure on my Menu, I do not want to hide the menu on click for that particular element because that would close the modal also. But another button on that menu open a drawer with no disclosure and that push body, so i need to close the menu on that particular element click.

### DIFF
--- a/packages/ui/src/components/Menu/__tests__/index.test.tsx
+++ b/packages/ui/src/components/Menu/__tests__/index.test.tsx
@@ -559,5 +559,38 @@ describe('menu', () => {
           <Menu.Item active>Active Props</Menu.Item>
         </Menu>,
       ))
+
+    test('should hideOnClick for specific item', async () => {
+      renderWithTheme(
+        <Menu disclosure={() => disclosure} id="menu">
+          <Menu.Item hideOnClick data-testid="hide-item">
+            Should Hide
+          </Menu.Item>
+          <Menu.Item data-testid="dont-hide-item">Should not Hide</Menu.Item>
+        </Menu>,
+      )
+      const menuButton = screen.getByRole<HTMLButtonElement>('button')
+      // Open Menu
+      await userEvent.click(menuButton)
+      const dialog = screen.getByRole('dialog')
+
+      await waitFor(() => {
+        expect(dialog).toBeVisible()
+      })
+
+      // Click item that should NOT close the menu
+      const itemNoHide = screen.getByTestId('dont-hide-item')
+      await userEvent.click(itemNoHide)
+      await waitFor(() => {
+        expect(dialog).toBeVisible()
+      })
+
+      // Click item that SHOULD close the menu
+      const itemHide = screen.getByTestId('hide-item')
+      await userEvent.click(itemHide)
+      await waitFor(() => {
+        expect(dialog).not.toBeVisible()
+      })
+    })
   })
 })
--- a/packages/ui/src/components/Menu/__tests__/index.test.tsx
+++ b/packages/ui/src/components/Menu/__tests__/index.test.tsx
@@ -563,33 +563,35 @@ describe('menu', () => {
     test('should hideOnClick for specific item', async () => {
       renderWithTheme(
         <Menu disclosure={() => disclosure} id="menu">
-          <Menu.Item hideOnClick data-testid="hide-item">
-            Should Hide
-          </Menu.Item>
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Menu/components/Item.tsx
```diff
@@ -71,11 +72,12 @@ export const Item = forwardRef<HTMLElement, ItemProps>(
       'data-testid': dataTestId,
       style,
       rightComponent,
+      hideOnClickItem,
     },
     ref,
   ) => {
     const {
-      hideOnClickItem,
+      hideOnClickItem: MenuHideOnClickItem,
```
lisalupi:
```suggestion
      hideOnClickItem: menuHideOnClickItem,
```

lisalupi (changes requested):

packages/ui/src/components/Menu/components/Item.tsx
PierreBertinet: done

lisalupi (approved):

jsulpis:
hi @PierreBertinet, could you explain why this feature is needed ?

packages/ui/src/components/Menu/components/Item.tsx
```diff
@@ -40,6 +40,7 @@ type ItemProps = {
   searchText?: string
   style?: CSSProperties
   rightComponent?: ReactNode
+  hideOnClickItem?: boolean | undefined
```
jsulpis: it's a prop on the Item component so I think `hideOnClick` would be enough, what do you think ?

jsulpis (commented):
could you add a quick test to check that with a Menu that has two items, one of them with the `hideOnClick` prop, the menu don't close when clicking on the first one and closes when clicking on the second one

packages/ui/src/components/Menu/components/Item.tsx
PierreBertinet: Done

packages/ui/src/components/Menu/__tests__/index.test.tsx
```diff
@@ -559,5 +559,38 @@ describe('menu', () => {
...
+      // Open Menu
+      await userEvent.click(menuButton)
+      const dialog = screen.getByRole('dialog')
+
+      await waitFor(() => {
+        expect(dialog).toBeVisible()
+      })
+
+      // Click item that should NOT close the menu
+      const itemNoHide = screen.getByTestId('dont-hide-item')
```
jsulpis: always use getByRole when it's possible, instead of testid.
See https://testing-library.com/docs/queries/about/#priority

idem for the other item

packages/ui/src/components/Menu/__tests__/index.test.tsx
```diff
@@ -559,5 +559,38 @@ describe('menu', () => {
...
+          <Menu.Item hideOnClick data-testid="hide-item">
+            Should Hide
+          </Menu.Item>
+          <Menu.Item data-testid="dont-hide-item">Should not Hide</Menu.Item>
+        </Menu>,
+      )
+      const menuButton = screen.getByRole<HTMLButtonElement>('button')
+      // Open Menu
+      await userEvent.click(menuButton)
+      const dialog = screen.getByRole('dialog')
```
jsulpis: the panel should not have a role dialog, that's a mistake on our side. But I see that there is a role="menu" inside that you can use 

```suggestion
      const dialog = screen.getByRole('menu')
```

jsulpis (changes requested):

packages/ui/src/components/Menu/__tests__/index.test.tsx
PierreBertinet: done

packages/ui/src/components/Menu/__tests__/index.test.tsx
PierreBertinet: done

jsulpis (approved):

PR #6369, opened 2026-04-23
fix(FileInput): validator return type and onDrop file name

lisalupi:
## Type

- Bug

### Summarize concisely:
`FileInput`: update validator type so that it can return `null` or `undefined` and fix `fileRejection` (in `onChange`) to return correct file names

### DIFF
--- a/packages/ui/src/components/FileInput/index.tsx
+++ b/packages/ui/src/components/FileInput/index.tsx
@@ -123,7 +123,7 @@ const FileInputBase = ({
           : 'Error'
 
         errorFiles.push({
-          fileName: name,
+          fileName: file.name,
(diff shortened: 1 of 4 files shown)


jsulpis (approved):

packages/ui/src/components/FileInput/index.tsx
```diff
@@ -123,7 +123,7 @@ const FileInputBase = ({
           : 'Error'
 
         errorFiles.push({
-          fileName: name,
+          fileName: file.name,
```
jsulpis, line 126: there is no test to update or create for this file.name and the return of the validator function ?

radhi-nasser-scaleway (approved):

PR #6370, opened 2026-04-23
fix(OptionSelector): remove unwanted padding and margin

lisalupi:
## Type

- Bug

### Summarize concisely:

`OptionSelector`: remove unwanted browser default padding and margin from fieldset


philibea (approved):

jsulpis (approved):
this will break the alignment where the component is used, we will have to be careful when migrating

PR #6382, opened 2026-04-27
fix(Menu): remove stop propagation on item

lisalupi:
## Type

- Bug

### Summarize concisely:
`Menu`: remove event.stopPropagation on nested menu items

### DIFF
--- a/.changeset/tiny-sites-bathe.md
+++ b/.changeset/tiny-sites-bathe.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/ui": patch
+---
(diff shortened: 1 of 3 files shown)


etienne-scaleway (approved):

Echo-IV (approved):

jsulpis (commented):
[this PR](https://github.com/scaleway/ultraviolet/pull/6015) introduced the event.stopPropagation() on both Item and MenuContent components. Should this fix be applied to the MenuContent too ?

PR #6383, opened 2026-04-28
feat(FileInput): read folders

lisalupi:
## Type

- Enhancement

#### What is expected?
`FileInput`:
- can read files inside drag and dropped directories 
- improve "accept" function on drag and drop
- "see more" should display the number of hidden files, not the total number of files


lisalupi:
might need to add more tests

philibea (approved):

jsulpis (approved):

PR #6390, opened 2026-04-30
feat(vitest): add axe setup

philibea:
## Type

-  Updated renderWithTheme: Changed from sync to async function that now:
   - Renders components as before
   - Automatically runs axe accessibility checks on the rendered container
   - Returns the render result plus axeResult for accessibility analysis

- Updated test setup:
   - Added axe matchers to vitest expect in setup.ts
   - Created type definitions in vitest-axe.d.ts

- Add a .a11y.test.tsx on Button component to test multiple theme contrast

fix:
- some test use fragment instead of fragment() which make a "Function" snapshots instead of dom.

 Feat:

- you can now use the direct result of axe inside renderTheme.
- you can use a function `expectNoViolations` at the end of the test.

   
The change affects all snapshot tests to include automatic accessibility violation checking alongside existing snapshot assertions.

We can use https://vitest.dev/api/test.html#todo to have a checklist of test already configure and to activate.


packages/illustrations/vitest.setup.ts
```diff
@@ -1,3 +1,4 @@
 import { setup } from '@utils/test'
 
+process.env.TZ = 'UTC'
```
chambo-e: You can pass this as env var in the package.json script instead no ?

chambo-e (approved):

packages/illustrations/vitest.config.ts
```diff
@@ -1,21 +1,10 @@
 import { createVitestConfig } from '@utils/test/config'
-import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin'
 
-export const vitestConfig = {
-  plugins: [
-    vanillaExtractPlugin({
-      identifiers: ({ hash }) => `uv_${hash}`,
-      unstable_mode: 'transform',
-    }),
```
lisalupi, line 9: why did you remove vanillaExtractPlugin?

packages/ui/src/components/Button/__tests__/a11y.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Button } from '..'
+
+describe('button - A11y', { tags: ['a11y'] }, () => {
+  it.each(themes)(
+    'should not have violations with (theme: %s)',
+    async (_, currentTheme) => {
+      const { container } = renderWithTheme(
+        <Button disabled onClick={() => {}}>
+          <PencilIcon />
+          Hellom
```
lisalupi:
```suggestion
          Hello
```

utils/test/src/themes.ts
```diff
@@ -0,0 +1,11 @@
...
+  consoleLightTheme,
+  consoleDarkTheme,
+  consoleDarkerTheme,
+} from '@ultraviolet/themes'
+
+export const themes = [
+  ['consoleLight', consoleLightTheme],
+  ['consoleDark', consoleDarkTheme],
+  ['consoleDarker', consoleDarkerTheme],
+] as const
```
lisalupi: I'm not sure about that. It is basically a re-export of themes from uv/themes. Why not do it in the themes package? Or maybe add a function that does the iteration on themes directly. I feel like it is strange to have the possibility to import theme from utils, and it might be confusing

packages/themes/src/themes/index.ts
```diff
@@ -1,3 +1,19 @@
-export { darkTheme as consoleDarkTheme } from './console/dark/__generated__'
-export { darkerTheme as consoleDarkerTheme } from './console/darker/__generated__'
-export { lightTheme as consoleLightTheme } from './console/light/__generated__'
+import { darkTheme as consoleDarkTheme } from './console/dark/__generated__'
+import { darkerTheme as consoleDarkerTheme } from './console/darker/__generated__'
+import { lightTheme as consoleLightTheme } from './console/light/__generated__'
+
+type ThemeName = 'consoleLightTheme' | 'consoleDarkTheme' | 'consoleDarkerTheme'
+
+export const consoleMapThemes = new Map<ThemeName, typeof consoleLightTheme>([
```
jsulpis: I don't see any usage of this, could it be removed ?

packages/themes/src/themes/index.ts
```diff
@@ -1,3 +1,19 @@
...
+  ['consoleLightTheme', consoleLightTheme],
+  ['consoleDarkTheme', consoleDarkTheme],
+  ['consoleDarkerTheme', consoleDarkerTheme],
+] as const)
+
+export const consoleThemes: (typeof consoleLightTheme)[] = [
+  consoleLightTheme,
+  consoleDarkTheme,
+  consoleDarkerTheme,
+] as const
```
jsulpis: `as const` seems redundant if there is a type l.13

packages/ui/src/components/CopyButton/__tests__/a11y.test.tsx
```diff
@@ -0,0 +1,12 @@
+import { consoleThemes } from '@ultraviolet/themes'
+import { renderWithTheme, expectNoViolations } from '@utils/test'
+import { describe, it } from 'vitest'
+import { CopyButton } from '..'
+
+describe('copy button - A11y', { tags: ['a11y'] }, () => {
```
jsulpis, line 6: do we really need this tag or is there a way to filter the tests based on the file name (`a11y.test.tsx`) ?

expect.d.ts
```diff
@@ -1,7 +1,8 @@
-import 'expect'
+import 'vitest-axe/extend-expect'
```
jsulpis, line 1: I don't understand what this file does, the [doc](https://github.com/chaance/vitest-axe#setup) says that `'vitest-axe/extend-expect'` should be imported in `vitest-setup.ts`, in which case there is no need for a declaration file. 

Also the `AxeCore` type below seems not used ?

utils/test/src/axe/expectNoViolations.ts
```diff
@@ -0,0 +1,49 @@
+import { expect } from 'vitest'
+import { axe } from 'vitest-axe'
+import type { AxeCore } from 'vitest-axe'
+
+const defaultResult: AxeCore.AxeResults = {
```
jsulpis: what is it used for and why is it exported ?

utils/test/src/axe/expectNoViolations.ts
```diff
@@ -0,0 +1,49 @@
...
+  reporter: 'v2',
+  rules,
+} as const
+
+const expectNoViolations = async (container: string | Element, options = defaultAxeOptions) => {
+  const axeResult = await axe(container, options)
+  expect(axeResult).toHaveNoViolations()
+}
+
+export { defaultResult, defaultAxeOptions, expectNoViolations }
```
jsulpis: why is `defaultAxeOptions` exported ?

packages/ui/src/components/Stack/__tests__/a11y.test.tsx
```diff
@@ -0,0 +1,17 @@
+import { renderWithTheme, expectNoViolations } from '@utils/test'
+import { describe, it } from 'vitest'
+import { Stack } from '..'
+
+describe('stack - A11Y', () => {
```
jsulpis: cf a previous comment, the a11y tag has been omitted here. We should enforce a behavior, either with the test file name or the tag, but make sure we do the same thing everywhere

packages/themes/src/themes/index.ts
philibea: you right i forgot to change it. Now it's use inside test to identifies correctly the name of the theme.

packages/ui/src/components/CopyButton/__tests__/a11y.test.tsx
philibea: can be usefull if you want to filter  pnpm test:unit --tagsFilter a11y.
Completly optional actually and juste use to isolate if necessary.

expect.d.ts
philibea: i didn't success to make it work properly on my side. This should work easily on a simple project but maybe not on monorepo as you have mutiple setup.

philibea: i remove AxeCore

utils/test/src/axe/expectNoViolations.ts
philibea: i remove this and merge the opts directly

jsulpis: still there ? Can it be _not_ exported if not necessary ? (clarify that it's a local variable)

packages/themes/src/themes/index.ts
```diff
@@ -1,3 +1,13 @@
-export { darkTheme as consoleDarkTheme } from './console/dark/__generated__'
-export { darkerTheme as consoleDarkerTheme } from './console/darker/__generated__'
-export { lightTheme as consoleLightTheme } from './console/light/__generated__'
+import { darkTheme as consoleDarkTheme } from './console/dark/__generated__'
+import { darkerTheme as consoleDarkerTheme } from './console/darker/__generated__'
+import { lightTheme as consoleLightTheme } from './console/light/__generated__'
+
+type ThemeName = 'consoleLightTheme' | 'consoleDarkTheme' | 'consoleDarkerTheme'
+
+export const consoleMapThemes = new Map<ThemeName, typeof consoleLightTheme>([
```
jsulpis: #detail but maybe `consoleThemesMap` instead ?

jsulpis (approved):

PR #6398, opened 2026-05-04
feat: update Navigation

jsulpis:
## Type

- Enhancement

### Summarize concisely:

Fix the animation and design of the Navigation component.

#### What is expected?

Smooth animation and consistent design

- remove the previous animations of the Navigation component
- replace them with CSS transitions and animations when expanding or collapsing the Navigation
- use a new hook useFlip to animate  the items vertically
- fix the design to avoid jumps in animations
- allow to toggle the navigation immediately without animation, for when it's toggled programmatically in JavaScript

**Design changes :**
- resize the navigation items in collapsed state so that they are square like on the design
- update the width of the Navigation in collapsed state (67px -> 64px) to be consistent with the design and the new size of the nav items
- resize the Scaleway logo which was too big for the collapsed state (and different from the design)
- change the padding of the navigation items in expanded state so that they are the same as the collapsed state, in order to avoid a jump in the animation (**to validate with the design**)
- fix the spacing between navigation items (there was a margin-top in addition to a gap on the container)
- remove an extra padding-top on navigation groups which is not in the design
- reserve space for the scrollbar and update its design in order to make it smaller and more subtle

| expanded  | image | image |
| collapsed  | image | image |


mhervouet (approved):

philibea (approved):

packages/ui/src/components/Separator/index.tsx
```diff
@@ -12,6 +12,10 @@ type SeparatorProps = {
   thickness?: number
   className?: string
   'data-testid'?: string
+  /**
+   * @private
+   */
+  'data-flip-id'?: string
```
lisalupi, line 18: why private for the separator but not the button?

lisalupi (approved):

packages/ui/src/components/Separator/index.tsx
jsulpis: oops good catch, I can keep it private everywhere because for now it's not intended to be used outside of the navigation. But I'm not sure this annotation has any effect anyway 🤷

PR #6402, opened 2026-05-05
feat(Helper): improve helper

lisalupi:
## Type

- Enhancement

#### What is expected?
Helper improvement:
- Standardize `helper` prop type as `ReactNode` in all components
- New component `Description` component to centralize helper logic and improve accessibility
- Implement `aria-describedby` support: when using `<Description />` instead of the internal `helper` prop, pass the Description's `id` to the component's `aria-describedby` prop to correctly link the content
- ⚠️ Some components (such as `Checkbox`) previously displayed both an error message and the helper text when both prop were defined. This is **not** the case anymore: an `error` message now always **replaces** the helper content. ⚠️ Priority: `error` > `success` > `helper`.


packages/ui/src/helpers/hasHelperText.ts
```diff
@@ -0,0 +1,10 @@
+import type { ReactNode } from 'react'
+
+export const canDisplay = (element: ReactNode) =>
+  element && typeof element !== 'boolean'
+
+export const hasHelperText = (
+  helper?: ReactNode,
+  error?: ReactNode,
+  success?: string | boolean,
```
jsulpis: why isn't the success a ReactNode like the helper and error ?

packages/ui/src/helpers/hasHelperText.ts
```diff
@@ -0,0 +1,10 @@
...
+) => helper || canDisplay(error) || typeof success === 'string'
```
jsulpis: why is there 3 different conditions for the 3 kinds of messages ?

packages/ui/src/components/Helper/__tests__/index.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <Helper id="id-helper" size="small" helper="helper" />
+        <input data-testid="test-input" aria-describedby="id-helper" />
+      </div>,
+    )
+
+    const input = screen.getByTestId('test-input')
+    const helper = screen.getByText('helper')
+
+    expect(input).toHaveAttribute('aria-describedby', 'id-helper')
+    expect(helper).toHaveAttribute('id', 'id-helper')
```
jsulpis: important:
- always use `getByRole` when it's possible, following the [testing library best practices](https://testing-library.com/docs/queries/about/#priority). `getByTestId` is the last one on the list
- prefer high-level matchers like [toHaveAccessibleDescription](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohaveaccessibledescription) that test the _behavior_ rather than the _implementation_ (there are multiple ways to set a description, or a label of an element for instance)


```suggestion
    const input = screen.getByRole('input')
    expect(input).toHaveAccessibleDescription('helper')
```

packages/ui/src/components/Slider/index.tsx
```diff
@@ -79,6 +82,11 @@ export const Slider = ({
     >
       {double ? (
         <DoubleSlider
+          aria-describedby={
+            !ariaDescribedBy && hasHelperText(helper, error)
+              ? helperId
+              : ariaDescribedBy
+          }
```
jsulpis: what do you think of this syntax instead ? To apply everywhere if it's ok for you

```suggestion
          aria-describedby={hasHelperText(helper, error) ? ariaDescribedBy || helperId : null }
```

packages/ui/src/components/TagInput/index.tsx
```diff
@@ -304,17 +301,14 @@ export const TagInput = ({
...
-        >
-          {error || success || helper}
-        </Text>
-      ) : null}
+      <Helper
+        helper={helper}
+        error={error}
+        success={success}
+        id={ariaDescribedBy ?? helperId}
+        disabled={disabled}
```
jsulpis: I see here that the condition for disabled was `disabled || readOnly`. I didn't check everywhere but maybe this condition should be applied in other components

```suggestion
        disabled={disabled || readOnly}
```

packages/ui/src/components/Helper/__stories__/Usage.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+      </Row>
+    </Stack>
+  </Stack>
+)
+
+Usage.parameters = {
+  docs: {
+    description: {
+      story:
+        'Example of a usage of this component. Here, to center the checkbox with the input and not with the input and the helper, use `Label` and `Row`. Do not forget to use prop `aria-describedby` on the input it to link the helper.',
```
jsulpis:
```suggestion
        'Example of a usage of this component. Here, to center the checkbox with the input and not with the input and the helper, use `Label` and `Row`. Do not forget to use the prop `aria-describedby` on the input to link it with the helper.',
```

packages/ui/src/components/Description/__stories__/Usage.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Row gap={2} templateColumns="1fr 1fr 1fr" alignItems="end">
+        <SelectInput
+          name="input21"
+          options={values}
+          size="medium"
+          label="Input1"
+        />
+        <SelectInput
+          name="input22"
+          options={values}
```
jsulpis:
```suggestion
        <SelectInput
          aria-describedby="id1"
          name="input21"
          options={values}
          size="medium"
          label="Input1"
        />
        <SelectInput
          aria-describedby="id2"
          name="input22"
          options={values}
```

packages/ui/src/compositions/OptionSelector/components/Selector.tsx
```diff
@@ -32,8 +33,10 @@ export const Selector = ({
   onChange: (val: string) => void
   options: DataType
   value?: string
+  helperId: string
 }) => (
   <SelectInput
+    aria-describedby={isHorizontal ? helperId : undefined}
```
jsulpis, line 32: I see that the error handling is quite complex on the OptionSelector : 
- the two selects can each have their own error
- there can be a "global" error which I guess applies to both selectors
- one of the errors is only displayed in horizontal mode (which looks like a bug)

I think ideally we want to have aria-describedby attributes on the selects which link to their error message, and also to the global error message if it's displayed. The attribute can accept a space-separated list of ids [as per the doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-describedby#id_reference_list).

Image

Image

packages/ui/src/components/Description/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import { useMemo } from 'react'
+
+import { canDisplay } from '../../helpers/hasHelperText'
+import { Text } from '../Text'
+
+import type { HTMLAttributes, ReactNode } from 'react'
+
+type HelperProps = {
+  helper?: ReactNode
```
jsulpis, line 9: what do you think of calling this prop "message" instead of "helper" ? Because `<Helper helper="..." />` looks a bit redundant

lisalupi: i think helper is better, that way it matches the props name

packages/ui/src/components/Slider/index.tsx
lisalupi: no this is not the correct behavior

packages/ui/src/helpers/hasHelperText.ts
lisalupi: it is not needed anywhere

packages/ui/src/helpers/hasHelperText.ts
lisalupi: because they have a different nature. I could change to `canDisplay(helper) || canDisplay(error) || typeof success === 'string'` if you prefer

packages/ui/src/components/Description/index.tsx
jsulpis: In that case it's the component that should be renamed, because it's not only the helper but also the error or success message. It's just that `<Helper helper="..." />` feels wrong

packages/ui/src/components/Slider/index.tsx
jsulpis: can you elaborate ? What's wrong with this ?

It fits in 3 lines instead of 5 and I find it more clear, unless there is an edge case that I didn't see

packages/ui/src/helpers/hasHelperText.ts
jsulpis: In fact I see that the error prop in all components is a `string | boolean` or sometimes just `string`. Why did you use `ReactNode` here ?

I think you could type the error as `string | boolean` so that it's like the success message, you don't need the `canDisplay` function and we can clearly see that only the helper is supposed to accept a ReactNode. This would solve the 2 comments that I had on these props

lisalupi: some components have a ReactNode error. At first I wanted to have error as a string | boolean but it would be breaking

packages/ui/src/components/Slider/index.tsx
lisalupi: in my case, it always uses the prop `ariaDescribedBy` for "aria-describedby"  _except_ when it is undefined _and_ the component has a helper/error/success content to display though props.
in your case, it only uses `ariaDescribedBy` when the component has an error/helper/success (props) -> this is not what we want, because users might user Helper component (instead of the prop) and link it with aria-describedby, but your code will return null (which should be undefined)

packages/ui/src/components/Description/index.tsx
lisalupi: i'm open to suggestions regarding the component name

packages/ui/src/components/Slider/index.tsx
jsulpis: oh yes I forgot that case, so maybe this should work ? still shorter 

```ts
  aria-describedby={ariaDescribedBy || (hasHelperText(helper, error) ? helperId : null)}
```

packages/ui/src/helpers/hasHelperText.ts
jsulpis: ok so maybe just type the error as `boolean | ReactNode` to make it clear that it accepts a boolean and have the type closer to the success ? And keep the rest as it is

packages/ui/src/components/Description/index.tsx
jsulpis: Maybe `StatusMessage` but it's not quite suitable for the helper, or `FieldDescription` ? `ContextualMessage` ?

packages/ui/src/helpers/hasHelperText.ts
lisalupi: it would create an [oxc error](https://oxc.rs/docs/guide/usage/linter/rules/typescript/no-redundant-type-constituents.html)

packages/ui/src/components/Description/index.tsx
lisalupi: `Description`?

packages/ui/src/helpers/hasHelperText.ts
jsulpis: ookay nevermind, let's keep it like that 😅

jsulpis:
I reverted the behavior of the error in the OptionSelector because there was a risk of losing error messages

jsulpis (approved):

mhervouet (approved):

PR #6403, opened 2026-05-06
fix: lint

lisalupi:
## Type

- Bug

#### What is expected?
Fix lint

### DIFF
--- a/.oxlintrc.jsonc
+++ b/.oxlintrc.jsonc
@@ -178,7 +178,7 @@
     "import/no-nodejs-modules": "warn",
     "import/max-dependencies": "off",
 
-    "oxc/branches-sharing-code": "off",
+    "oxc/branches-sharing-code": "warn",
 
     "react/display-name": "off",
     "react/no-react-children": "off",
--- a/.oxlintrc.jsonc
+++ b/.oxlintrc.jsonc
@@ -112,6 +112,17 @@
         "vitest/prefer-to-be-truthy": "off",
         "vitest/require-test-timeout": "off",
         "vitest/require-mock-type-parameters": "off",
+        "vitest/prefer-expect-assertions": "off",
+        "vitest/no-hooks": "off",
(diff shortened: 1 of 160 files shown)


philibea (approved):

.oxlintrc.jsonc
```diff
@@ -161,11 +172,14 @@
     ],
 
     "eslint/no-negated-condition": "off",
+    "react/forbid-component-props": "off",
 
     "import/no-unassigned-import": "off",
     "import/no-nodejs-modules": "warn",
     "import/max-dependencies": "off",
 
+    "oxc/branches-sharing-code": "off",
```
chambo-e: This could be useful no ? 🤔

utils/test/vitest.config.ts
```diff
@@ -5,6 +5,7 @@ import type { TestUserConfig, ViteUserConfig, Plugin } from 'vitest/config'
 /**
  * @deprecated https://github.com/vitest-dev/vitest/issues/9935
  */
+// oxlint-disable-next-line func-name-matching
 export const coveragePluginIssue = function fixIstanbulBabelInterop(): Plugin {
```
chambo-e: Do we still need this func ?

.oxlintrc.jsonc
philibea: yes but painfull to fix now

chambo-e: So maybe we need to keep this as a warning ? To avoid creating more

lisalupi: it could, but we often have the same variable name in branches (but not declared the same way) and it triggers the rule.
I can switch it to "warning" instead of error 
edit: did not see all the comments, but yes i can switch to warning

utils/test/vitest.config.ts
lisalupi: I don't think we do

chambo-e (approved):

PR #6406, opened 2026-05-07
fix: correct display of the close button on the tag

jsulpis:
## Type

- Bug

#### What is expected?

Fix the close button on the tags (which also fixes the multiselect)

- update the size of the close button
- update the gap and right padding accordingly
- set the right sentiment on the close button (it was always neutral)

| url  | image | image |
| url  | image | image |

### DIFF
--- a/packages/ui/src/components/Tag/index.tsx
+++ b/packages/ui/src/components/Tag/index.tsx
@@ -107,7 +107,12 @@ export const Tag = ({
         <button
           className={cn(
             className,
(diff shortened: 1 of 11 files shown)


packages/ui/src/components/Tag/index.tsx
jsulpis, line 51: I didn't update the keyValue variant because according to the types it's not supposed to be closable, even though the prop onClose is used in it

lisalupi (approved):

mhervouet (approved):

PR #6408, opened 2026-05-07
chore: use @scaleway/oxfmt-config

chambo-e:
## Type

- Enhancement

### Summarize concisely:

Sorry about that 😄 

So we can be consistent across our codebases

This should be mainly cosmetic
Most of the changes comes from the sortImport and the printWidth 

I can override them to what they were for now if you want to not have as much changes


oxfmt.config.ts
```diff
@@ -0,0 +1,30 @@
+import config from '@scaleway/oxfmt-config'
+import { defineConfig } from 'oxfmt'
+
+export default defineConfig({
+  ...config,
+  ignorePatterns: [
+    'CHANGELOG.md',
```
jsulpis: do you think we could have that in `@scaleway/oxfmt-config` ? It looks like generic patterns

jsulpis (approved):

philibea (approved):

oxfmt.config.ts
philibea: at least propose a default ignorePattern list we can use from the config
```
import { defaultIgnorePatterns } from '@scaleway/oxfmt-config'

chambo-e: Since it's a simple spread and not an extend we can't properly merge arrays and objects

I'll take a look if I can add a mergeConfig, like oxlint has, in our package :)

lisalupi (approved):
i find it weird that `ui/composition` had no changes and most `ui/components` did not either (or maybe it is not loading?)

oxfmt.config.ts
chambo-e: Here you go :)

https://github.com/scaleway/scaleway-lib/pull/3266

chambo-e:
> i find it weird that `ui/composition` had no changes and most `ui/components` did not either (or maybe it is not loading?)

@lisalupi I think it might be a github glitch, most of the changes are located in these two folders 😄 

Don't know if the ui might load better today ?

PR #6420, opened 2026-05-11
chore(storybook): rename a11y property to a11y-status 

philibea:
## Type

- Bug

### Summarize concisely:

Migrate status a11y to a11YStatus to avoid override a11y addon


.storybook/preview.tsx
jsulpis, line 1: the changes in this file add nothing to me, can it be reverted ?

jsulpis (approved):

PR #6421, opened 2026-05-11
test(config): use jsdom, simplify config

philibea:
## Type

- Enhancement

#### What is expected?

(Describe what you did)

1.

2.

3.


packages/ui/src/components/SelectInput/__tests__/index.test.tsx
```diff
@@ -1181,20 +1173,20 @@ describe('selectInput', () => {
       { timeout: 1000 },
     )
 
-    const selectAllGroupCheckBox = within(dropdown).getByRole('checkbox', {
-      name: 'TERRESTRIAL PLANETS',
+    const selectAllGroupCheckBox = screen.getByRole('checkbox', {
```
jsulpis: why did you remove the `within(dropdown)` ?

chambo-e (approved):

packages/ui/src/compositions/Navigation/__tests__/index.test.tsx
```diff
@@ -106,11 +106,21 @@ const NavigationNested = ({ pinnedFeature = true }: BasicNavigationProps & { onS
...
 
 describe('navigation', () => {
-  it('render with basic content', () => shouldMatchSnapshot(<BasicNavigation />))
+  it('render with basic content', () => {
+    const { asFragment } = renderWithTheme(<BasicNavigation />)
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  it('render without pinnedFeature', () => {
+    const { asFragment } = renderWithTheme(<BasicNavigationNoExpand pinnedFeature={false} />)
```
jsulpis:
```suggestion
    const { asFragment } = renderWithTheme(<BasicNavigation pinnedFeature={false} />)
```

packages/ui/package.json
```diff
@@ -64,7 +64,7 @@
     "build:profile": "npx vite-bundle-visualizer -c vite.config.ts",
     "lintpublish": "publint",
     "size": "size-limit",
-    "test:unit": "LC_ALL=en_US.UTF-8  pnpm vitest --run --config vitest.config.ts",
+    "test:unit": "TZ=UTC LC_ALL=en_US.UTF-8  pnpm vitest --run --config vitest.config.ts",
```
jsulpis: I guess vitest finds the config file automatically, why do we repeat it everywhere ? (same with `vite.config.ts`)
```suggestion
    "test:unit": "TZ=UTC LC_ALL=en_US.UTF-8  pnpm vitest --run",
```

packages/illustrations/vitest.config.ts
```diff
@@ -1,21 +1,19 @@
...
+  dom: true,
   test: {
-    ...createVitestConfig({
-      name: 'uv/illustration',
-      dom: false,
-      environment: 'jsdom',
-      setupFiles: ['./vitest.setup.ts'],
-    }),
+    name: 'uv/illustration',
+    environment: 'jsdom',
```
jsulpis: it's the default value

```suggestion
```

packages/illustrations/vitest.config.ts
```diff
@@ -1,21 +1,19 @@
...
 
-export const vitestConfig = {
+export const vitestConfig = createVitestConfig({
   plugins: [
     vanillaExtractPlugin({
       identifiers: ({ hash }) => `uv_${hash}`,
       unstable_mode: 'transform',
     }),
   ], // Enable unstable mode for better compatibility with Vitest
+  dom: true,
```
jsulpis: this option should be in the "test" object below, and I see that it enables "happy-dom" so I guess it's incompatible with `environment: 'jsdom'` ?

Image

utils/test/src/vitest/setup.ts
```diff
@@ -20,6 +33,16 @@ export const setup = () => {
     vi.spyOn(globalThis.Math, 'random').mockReturnValue(0.415_591_366_944_480_4)
 
     window.ResizeObserver = vi.fn().mockImplementation(MockResize)
+    window.matchMedia = vi.fn().mockImplementation(MockMatchMedia)
+
+    // Mock clipboard API for dom: true
```
jsulpis: I see only one `dom: true` in the illustrations package and I think it has no effect, so is this really related to the `dom: true` option ?

philibea: dom: true meaning we use happydom. we should avoid using this pattern yes

jsulpis (approved):

PR #6429, opened 2026-05-12
feat: add a header to the legend of the Pie chart

jsulpis:
## Type

- Enhancement

#### What is expected?

Allo to display the unit of the legend in the header instead of repeat it on each line

- add a prop `legendHeader` on the PieChart component to display the unit. It's optional for backward compatibility, so users can still have the unit inline until they decide to move it to the header
- clean and simplify the component : fewer DOM nodes, fewer styles
- update the styles to match the Figma design (things have moved a little bit)

| url  | image | image |
| url  | image | image |
| url | image | image |

### DIFF
--- a/packages/ui/src/components/PieChart/Legends.tsx
+++ b/packages/ui/src/components/PieChart/Legends.tsx
@@ -1,58 +1,70 @@
 'use client'
 
 import { assignInlineVars } from '@vanilla-extract/dynamic'
+import { Text } from '../Text'
 import { Tooltip } from '../Tooltip'
-import { Tooltip as TooltipContainer } from './Tooltip'
+import { TooltipContent } from './Tooltip'
 import type { Data } from './types'
 import { colorBullet, pieChartStyle } from './styles.css'
 
 type LegendsProps = {
   data?: Data[]
+  legendHeader?: string
   focused?: string
   onFocusChange: (index?: string) => void
   colors: string[]
 }
 
-export const Legends = ({ focused, data, onFocusChange, colors }: LegendsProps) => (
-  <ul className={pieChartStyle.list}>
-    {data?.map((item, index) => {
-      const isSegmentFocused = focused !== undefined && item.id === focused
+export const Legends = ({ focused, data, legendHeader, onFocusChange, colors }: LegendsProps) => (
+  <div className={pieChartStyle.legendContainer}>
+    {legendHeader ? (
+      <Text
+        as="span"
+        variant="captionStrong"
+        sentiment="neutral"
+        prominence="weak"
+        className={pieChartStyle.legendHeader}
+      >
+        {legendHeader}
+      </Text>
+    ) : null}
+    <ul className={pieChartStyle.list}>
+      {data?.map((item, index) => {
+        const isSegmentFocused = focused !== undefined && item.id === focused
 
-      const id = `chart-legend-${item.id}`
+        const id = `chart-legend-${item.id}`
 
-      return (
-        <Tooltip id={id} key={item.id} text={<TooltipContainer data={item} />} visible={isSegmentFocused}>
-          <li className={pieChartStyle.listItem({ isFocused: isSegmentFocused })}>
-            <div
-              className={pieChartStyle.toggleBox}
-              data-testid={id}
+        return (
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/PieChart/index.tsx
```diff
@@ -50,31 +52,9 @@ export const PieChart = ({
...
-          </Text>
-        </div>
-      ) : null,
-    [emptyLegend],
-  )
-
   // oxlint-disable-next-line typescript/no-unsafe-type-assertion
   const localColors = getLegendColor(theme as typeof UVTheme)
 
-  const LegendDisplayer = useCallback(
```
jsulpis, line 68: the `useCallback` caused React to recreate all the DOM nodes each time we hover/unhover a legend line

packages/ui/src/components/PieChart/Legends.tsx
```diff
@@ -1,58 +1,70 @@
...
+        const id = `chart-legend-${item.id}`
 
-      return (
-        <Tooltip id={id} key={item.id} text={<TooltipContainer data={item} />} visible={isSegmentFocused}>
-          <li className={pieChartStyle.listItem({ isFocused: isSegmentFocused })}>
-            <div
-              className={pieChartStyle.toggleBox}
-              data-testid={id}
+        return (
+          <Tooltip key={item.id} text={<TooltipContent data={item} />} visible={isSegmentFocused}>
```
jsulpis, line 38: I removed the id on the tooltip because it's duplicate with the id on the bullet below. I don't know why they exist so can't know if it will break something 🤷

packages/ui/src/components/PieChart/Legends.tsx
```diff
@@ -1,58 +1,70 @@
...
+            <li
+              className={pieChartStyle.listItem({
+                isFocused: isSegmentFocused,
+              })}
               onBlur={() => onFocusChange()}
```
jsulpis, line 43: the `onFocus` / `onBlur` have no effect because it's the parent (tooltip) element that receives the focus. So the legend line is not highlighted and the tooltip not shown when using tab. But the Tooltip doesn't accept onBlur/onFocus so I left that as it is

packages/ui/src/components/PieChart/Legends.tsx
```diff
@@ -1,58 +1,70 @@
...
-              id={`chart-legend-${item.id}`}
-              style={assignInlineVars({
-                [colorBullet]: colors[index],
-              })}
-            />
-            <div className={pieChartStyle.label}>
-              <span className={pieChartStyle.text({ isFocused: isSegmentFocused })}>{item.name}</span>
-              <span className={pieChartStyle.line}>
-                <span
-                  className={pieChartStyle.progressiveLine({
```
jsulpis, line 46: I removed the progressiveLine because it only carried a broken animation of the line. I just changed the color of the line when the item is highlighted

packages/ui/src/components/PieChart/Tooltip.tsx
```diff
@@ -16,27 +16,25 @@ type TooltipProps = {
   }
 }
 
-export const Tooltip = ({ data }: TooltipProps) => (
-  <div role="tooltip" tabIndex={-1}>
```
jsulpis, line 20: I removed this because the component is rendered inside the tooltip which already has the `role="tooltip"`

packages/ui/src/components/PieChart/Legends.tsx
jsulpis, line 38: note: the Tooltip component introduces an HTML element between the `ul` and `li` which is wrong (`li` should be immediately inside an `ul`). I didn't want to rework that, it's a global issue with the Tooltip

mhervouet (approved):

packages/ui/src/components/PieChart/Legends.tsx
philibea: Should we moove the Tooltip inside the content of the li instead ?

philibea (approved):

jsulpis:
I tried to move the Tooltip inside of the `<li>` but it doesn't forward the className and doesn't handle blur/focus events so it's not possible right now

chambo-e (approved):

PR #6435, opened 2026-05-13
feat(Link): update styles, sentiments, visited state, prominences

jsulpis:
### Summary

- update default sentiment: "info" if `target === '_blank'`, "primary" otherwise
- update the text decoration and animation, fix styles for the different sizes
- remove visited state for links without `target === '_blank'`
- add sentiments values
- deprecate the `variant` prop and the `weak` prominence

### Screenshots

| State |  Before   |      After |
| external  | image | image |
| prominences | image | image |
| Alert | image | image| 
| Toaster | image |  image |


packages/ui/src/components/Action/Link/index.tsx
```diff
@@ -98,45 +98,33 @@ export const Link = forwardRef(
...
-    }, [size])
     useEffect(() => {
-      if (oneLine && usedRef?.current) {
-        const { offsetWidth, scrollWidth } = usedRef.current
+      if (oneLine && elementRef?.current) {
+        const { offsetWidth, scrollWidth } = elementRef.current
         setIsTruncated(offsetWidth < scrollWidth)
       }
-    }, [oneLine, ref, usedRef])
+    }, [oneLine, ref, elementRef])
```
philibea, line 118: thought(non-blocking): 
If children changes and the text becomes longer/shorter, truncation is not re-evaluated.

packages/ui/src/components/Action/Link/styles.css.ts
```diff
@@ -1,36 +1,33 @@
...
 import { recipe } from '@vanilla-extract/recipes'
-import { PROMINENCES } from './constants'
+import type { LinkSentiment, LinkProminence } from './constants'
+import { PROMINENCES, SENTIMENTS } from './constants'
 
 const TRANSITION_DURATION = 250
 
-type ProminenceType = keyof typeof PROMINENCES
+function getLinkStyle(sentiment: LinkSentiment, prominence: LinkProminence) {
+  if (prominence === 'stronger') {
```
philibea: nitpick(non-blocking): 

The type system allows `<Link sentiment="primary" prominence="stronger" /> `
this will be weird for developer that use it ?

packages/ui/src/components/Action/Link/__stories__/Icons.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import type { ComponentProps } from 'react'
+import { Link } from '..'
+import { Stack } from '../../../Layout/Stack'
+import { SIZES } from '../constants'
+
+export const Icons = (props: ComponentProps<typeof Link>) => (
```
lisalupi, line 7: maybe rename it as IconPosition or ArrowIcon? At first glance, one might think that it applies to any icon

packages/ui/src/components/Action/Link/__stories__/Prominence.stories.tsx
```diff
@@ -1,21 +1,72 @@
...
-  ),
-] as Decorator[]
+  )
+}
+
+Prominence.parameters = {
+  docs: {
+    description: {
+      story:
+        'Set the `prominence` property to the same value as the background. `default` and `strong` are available with all sentiments, and `stronger` is available on sentiment `neutral` only.',
```
lisalupi, line 62: why not display all sentiments in default? To not repeat `Sentiment` story?

packages/ui/src/components/Action/Link/constants.ts
```diff
@@ -1,6 +1,12 @@
...
-  weak: 'weak',
-} as const
+export const PROMINENCES = ['default', 'strong', 'stronger'] as const
+/**
+ * @deprecated
+ */
+type LegacyLinkProminence = 'weak'
+export type LinkProminence = 'default' | 'strong' | 'stronger' | LegacyLinkProminence
+
+export const SENTIMENTS = ['primary', 'info', 'success', 'warning', 'danger', 'neutral'] as const
```
lisalupi, line 8: we have a SENTIMENTS variable defined elsewhere. Do we really don't want a `secondary` sentiment for links?

lisalupi (changes requested):
Image 

Image



Image



Image

Image

the link should be updated in the components using it (Toaster, GlobalAlert, Breadcrumbs, Banner)

packages/ui/src/components/Action/Link/index.tsx
jsulpis: I think it's unlikely that a link text changes so I would rather not fix that in this PR

packages/ui/src/components/Action/Link/styles.css.ts
jsulpis: This condition was a mistake, I changed it to apply the stronger prominence only on neutral sentiment. If stronger is used with another sentiment il will fallback to the default prominence

packages/ui/src/components/Action/Link/__stories__/Prominence.stories.tsx
jsulpis: yes but I forgot to display the backgrounds, so I added all sentiments in this story with their background

packages/ui/src/components/Action/Link/constants.ts
jsulpis: secondary is already used for the visited state so yes apparently there is no secondary sentiment

packages/ui/src/components/Action/Link/__stories__/Icons.stories.tsx
jsulpis: Actually there can be other icons in the link, I added an example with the Attach icon for links to anchors

jsulpis:
> Image Image Image Image Image
> the link should be updated in the components using it (Toaster, GlobalAlert, Breadcrumbs, Banner)

should be fixed, but on the GlobalAlert the links are expected to be dark, it's the textStrong colors. They don't have enough contrast but it's like the Buttons, theme issue

packages/ui/src/components/Action/Link/__stories__/Icons.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+          {size}
+        </Link>
+      ))}
+    </Stack>
+    <Stack direction="row" alignItems="end" gap="3">
+      <Link {...props}>
+        Link <DocPaperIcon />
+      </Link>
+      <Link {...props}>
+        Link <AttachIcon />
```
jsulpis: the alignment looks off, I'm waiting for the design to know what is expected

Image

packages/ui/src/components/Action/Link/styles.css.ts
```diff
@@ -42,55 +39,50 @@ function makeVariant(variant: 'captionStrong' | 'bodySmallStrong' | 'bodyStrong'
...
   base: {
-    backgroundColor: 'none',
     border: 'none',
     padding: 0,
-    textDecoration: 'underline',
-    textDecorationThickness: 1,
-    textUnderlineOffset: 2,
-    textDecorationStyle: 'dotted',
-    gap: theme.space[1],
+    display: 'inline-flex',
```
jsulpis, line 49: I switched to flex to apply the gap and alignment to icons in children like the <AttachIcon/> which would be complicated to style.

The display becomes block with the oneLine variant so it looses the flex gap and alignment of the icon, but I think that oneLine links should not have icons because they are truncated with the text. Except if a oneLine link can have an icon on the left, that would be an issue

packages/ui/src/components/Feedback/GlobalAlert/styles.css.ts
```diff
@@ -23,6 +24,18 @@ const container = styleVariants({
   promotional: [containerBase, { background: theme.colors.other.gradients.background.linear.aqua }],
 })
 
+globalStyle(`${container['danger']} ${linkStyle.defaultLink}`, {
+  color: theme.colors.danger.textStrong,
+})
+
```
lisalupi: instead of doing a globalStyle, you can put it in the link css

lisalupi:
Capture d’écran 2026-09-21 à 10 09 01
the link is still black instead of white on GlobalAlert

packages/ui/src/components/Feedback/GlobalAlert/styles.css.ts
jsulpis: ✅

jsulpis:
> Capture d’écran 2026-09-21 à 10 09 01 the link is still black instead of white on GlobalAlert

it's expected, it's the theme token danger.textStrong like on the Button

lisalupi (approved):

mhervouet (approved):

PR #6444, opened 2026-05-18
chore: remove devDependencies from changesets

jsulpis:
## Type

- Documentation

#### What is expected?

non-production dependencies don't generate a changeset

1. enable the flag on the CLI `@scaleway/changesets-renovate` to exclude devDeps
2. move packages from `dependencies` to `devDependencies` so that they don't generate a changeset

### DIFF
--- a/.github/workflows/changesets-renovate.yml
+++ b/.github/workflows/changesets-renovate.yml
@@ -31,3 +31,5 @@ jobs:
           cache: 'pnpm'
       - name: Run changesets-renovate
         run: pnpm dlx @scaleway/changesets-renovate
+        env:
+          EXCLUDE_DEVDEPS: true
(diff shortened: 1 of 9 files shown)


.github/renovate.json
```diff
@@ -35,7 +35,9 @@
     {
       "matchPackageNames": ["/^@testing-library/"],
       "groupName": "Testing library",
-      "groupSlug": "testing-library"
+      "groupSlug": "testing-library",
+      "semanticCommitType": "chore",
+      "semanticCommitScope": "devDeps"
```
jsulpis: the testing library packages are `dependencies` in the package `@utils/test` so I override the type and scope of the commit so that it's considered as devDependencies

jsulpis: outdated

philibea (approved):

PR #6454, opened 2026-05-20
feat: add API proposition for Filters

jsulpis:
## Type

- Feature

#### What is expected?

Proposed API for the Filters component.

Review the [API design](https://featfiltersstorybook.storybook.ultraviolet.scaleway.com/?path=/docs/compositions-filters--docs) and share your thoughts (missing features, bad naming etc).

This is only a proposition of API for now. The implementation will follow.

### DIFF
--- a/packages/ui/src/compositions/Filters/__stories__/index.mdx
+++ b/packages/ui/src/compositions/Filters/__stories__/index.mdx
@@ -1,0 +1,422 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="Compositions/Filters" />
+
+# Filters
+
+Filters allow you to filter offers based on various criteria. This document describes a proposed API to implement the filters.
+
+## Configuration
+
+### Built-in Filter Types
+
+1. **Text (`text`)**: Allows filtering by text with optional debouncing.
+2. **Select (`select`)**: Allows choosing an option from a list.
+3. **Multi-select (`multiSelect`)**: Allows choosing multiple options from a list.
+4. **Slider (`slider`)**: Allows defining a range of values (single or double).
+5. **Date Range (`dateRange`)**: Allows selecting a date range.
+
+### Custom Filter Types
+
+Any string can be used for a filter type, allowing to render custom filter components that will have to be provided.
+
+### Filter Configuration
+
+A configuration array is used to define all the filters to display.
+
+```typescript
+const filters: FilterConfig<FilterValues> = [
+  // Text filter with debouncing
+  {
+    type: 'text',
+    id: 'search',
+    label: 'Search',
+    debounceMs: 500,
+    placeholder: 'Type to search...',
+  },
+
+  // Select filter
+  {
+    type: 'select',
+    id: 'status',
+    label: 'Status',
+    options: [
+      { label: 'Active', value: 'active' },
+      { label: 'Inactive', value: 'inactive' },
+    ],
+  },
+
+  // Multi-select filter
+  {
+    type: 'multiSelect',
+    id: 'tags',
+    label: 'Tags',
+    options: [
+      { label: 'Production', value: 'prod' },
+      { label: 'Development', value: 'dev' },
+      { label: 'Staging', value: 'staging' },
+    ],
+    searchable: true,
+  },
+
+  // Slider filter (single value)
+  {
+    type: 'slider',
+    id: 'price',
+    label: 'Price',
+    min: 0,
+    max: 1000,
+    step: 10,
+  },
+
+  // Slider filter (range)
+  {
+    type: 'slider',
+    id: 'range',
+    label: 'Range',
+    min: 0,
+    max: 100,
+    double: true,
+  },
+
+  // Group of filters
+  {
+    type: 'group',
+    id: 'advanced',
+    label: 'Advanced',
+    expanded: true,
+    filters: [
+      {
+        id: 'gpu',
+        type: 'select',
+        label: 'GPU',
+        options: [...],
+      },
+    ],
+  },
+
+  // Custom filter type
+  {
+    id: 'zones',
+    type: 'zones',
+    label: 'Zones',
+    foo: 'bar'
+  },
+]
+```
+
+## Using the Filters Component
+
+The Filter component implements the filter UI pattern that displays a few filters (2-3), a button to open a Drawer containing all filters, and a clear button.
+
+### Basic Usage
+
+```typescript
+import { Filters, type FilterConfig } from '@ultraviolet/form'
+
+// default values when no filter is applied
+const defaultValues: FilterValues = {
+  name: '',
+  regions: [],
+  ram: [1, 64],
+  gpu: 'none',
+}
+
+// initial values applied on the first mount
+const initialValues: FilterValues = {
+  ram: [16, 32],
+}
+
+const config: FilterConfig<FilterValues> = [
+  ...
+]
+
+function MyPage() {
+  return (
+    <Filters
+      config={config}
+      defaultValues={defaultValues}
+      initialValues={initialValues}
+      onChange={(values) => {
+        // Called on every user interaction with a filter
+        updatePreview(values)
+      }}
+      onSubmit={(values) => {
+        // Called when filters are committed (button "see results" or update in the main row)
+        refetch({ filters: values })
+      }}
+      onDrawerOpen={() => {
+        trackEvent('filters_opened') // Example analytics call
+      }}
+      labels={{
+        seeAll: 'All Filters',
+        clearAll: 'Clear All',
+        submit: 'See Results',
+        clear: 'Clear',
+        drawerHeader: 'Filters',
+      }}
+    />
+  )
+}
+```
+
+### Layout Configuration
+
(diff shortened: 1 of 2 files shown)


packages/ui/src/compositions/Filters/__stories__/index.mdx
```diff
@@ -0,0 +1,422 @@
...
+
+## Configuration
+
+### Built-in Filter Types
+
+1. **Text (`text`)**: Allows filtering by text with optional debouncing.
+2. **Select (`select`)**: Allows choosing an option from a list.
+3. **Multi-select (`multiSelect`)**: Allows choosing multiple options from a list.
+4. **Slider (`slider`)**: Allows defining a range of values (single or double).
+5. **Date Range (`dateRange`)**: Allows selecting a date range.
```
chambo-e, line 17: Needs to handle NumberInput also :)

packages/ui/src/compositions/Filters/__stories__/index.mdx
```diff
@@ -0,0 +1,422 @@
...
+  config={config}
+  layout={{
+    mainFilters: ['search', 'status'],
+    templateColumns: '2fr 1fr auto'
+  }}
+  // ... other props
+/>
+```
+
+### Custom Field Components
```
chambo-e, line 180: Do we want to allow custom components ? 
It's going to be a bit la foire au jambon no ? Every designer is going to want different inputs

packages/ui/src/compositions/Filters/__stories__/index.mdx
```diff
@@ -0,0 +1,422 @@
...
+      defaultValues={defaultValues}
+      initialValues={initialValues}
+      onChange={(values) => {
+        // Called on every user interaction with a filter
+        updatePreview(values)
+      }}
+      onSubmit={(values) => {
+        // Called when filters are committed (button "see results" or update in the main row)
+        refetch({ filters: values })
+      }}
```
chambo-e, line 148: So we'd need to handle two use case for each type of interaction ? 

if an user change values via the main filters (and not the drawer) we are going to use onChange and if it's submitted inside the drawer we need to use onSubmit ?

packages/ui/src/compositions/Filters/__stories__/index.mdx
```diff
@@ -0,0 +1,422 @@
...
+}
+
+const handleDrawerClose = () => {
+  discard() // Reverts to last submitted values
+}
+```
+
+## Implementation ideas
+
+### UI agnostic hook
```
chambo-e, line 258: Same, IMO the logic should be tied to the UI, else it's gonna be used in every wrong way possible 😬

packages/ui/src/compositions/Filters/__stories__/index.mdx
```diff
@@ -0,0 +1,422 @@
...
+  defaultValues: T
+  dirtyFields: (keyof T)[]
+
+  // Actions
+  setValue: <K extends keyof T>(key: K, value: T[K]) => void
+  setValues: (values: Partial<T>) => void
+  resetField: <K extends keyof T>(key: K) => void
+  reset: () => void
+  submit: () => void
+  discard: () => void
```
chambo-e, line 354: When would we use that ? With reset we should be golden 100% of the time no ?

packages/ui/src/compositions/Filters/__stories__/index.mdx
jsulpis: OK I don't see it in the console but we can add it if it will be useful

packages/ui/src/compositions/Filters/__stories__/index.mdx
jsulpis: This allows to render components with custom render logic or types that are not in UV (the regions and zones for now). I don't know how this will be translated in the design, in my mind it was mostly a technical feature

packages/ui/src/compositions/Filters/__stories__/index.mdx
jsulpis: no, use `onSubmit` when you want to actually apply the filters, and `onChange` when you want to update something in realtime without applying the filters. For instance, fetch the number of results and display them in the button "see results". Maybe there is no usage for `onChange` in the console right now. Do you think I should rework the naming of these events ?

And changes in the main filters will trigger a submit.

packages/ui/src/compositions/Filters/__stories__/index.mdx
jsulpis: It's mostly an internal implementation, we don't have to export it (I put it here to gather feedback). The goal is to decouple the code of the logic and the UI for easier maintenance

packages/ui/src/compositions/Filters/__stories__/index.mdx
jsulpis: - `reset` reverts the values to the default value, it's the equivalent of clicking the "clear all" button
- `discard` reverts the values the the last submitted values, for instance when the user has changed filters in the drawer and close it without validating, we want to revert the values to match the applied filters. This behavior replaces the "tempFields" that are in the current console implementation

packages/ui/src/compositions/Filters/__stories__/index.mdx
philibea: I think we should add the possibility for all ours possible Input ?

packages/ui/src/compositions/Filters/__stories__/index.mdx
```diff
@@ -0,0 +1,422 @@
...
+        label: 'GPU',
+        options: [...],
+      },
+    ],
+  },
+
+  // Custom filter type
+  {
+    id: 'zones',
+    type: 'zones',
```
philibea, line 102: should we have a type "custom" instead ?

packages/ui/src/compositions/Filters/__stories__/index.mdx
philibea: Or we need to migrate the Region / Zone inside UV to avoid custom component but soon or later i think it's will be required to allow a custom component. 

If a dev or designer use this Filter with only Custom Component there is something wrong for sure but we need to keep flexibility also

philibea (changes requested):

packages/ui/src/compositions/Filters/__stories__/index.mdx
jsulpis: not all: phone, password etc won't ever be in filters. I think we should keep the current ones and add more later if needed

packages/ui/src/compositions/Filters/__stories__/index.mdx
jsulpis: currently the Regions and Zones filters contain logic and helpers that are really tied to the console, we can't easily move them to UV. The proposed API to add custom components allows to keep them in the console for now, and later if we manage to migrate these two filters to UV we can remove the override in the console.

packages/ui/src/compositions/Filters/__stories__/index.mdx
jsulpis: the proposition is to have anything in the type, and then provide the associated component in the `components` prop using the same string to link the component to its type. With a type "custom" we would need something else to link a custom component to it. 

```ts
const config = [{ type: 'zones', ... }, ...]

<Filters config={config} components={{ 
    zones: <ZoneFilter /> // <- "zones" linked to the "zones" type
}}/>
```

packages/ui/src/compositions/Filters/__stories__/index.mdx
chambo-e: Agree that not everything needs to be handled from the get go 

I mentionned NumberInput because I already have an MR pending where I add it :)

packages/ui/src/compositions/Filters/__stories__/index.mdx
chambo-e: Gotcha, but we (and the designers) should be aware of that, they should avoid as much as possible to have custom ones :)

packages/ui/src/compositions/Filters/__stories__/index.mdx
chambo-e: Every change in the main filter will trigger a submit ??

For example if I have a slider double on the main filters and I slide the cursor I'm gonna have a submit called for each values ? 🤔

packages/ui/src/compositions/Filters/__stories__/index.mdx
chambo-e: mmh okay, I've to admit that I never had this use case but it seems to be used so alright :)

packages/ui/src/compositions/Filters/__stories__/index.mdx
jsulpis: It's already the case ! See the GPU creation page. Since there is no button to apply the filters in the main row, I don't think we have other options than to submit all changes immediately. But I was just trying to replicate the current behavior, I did not challenge it

chambo-e: True, my bad

I'm wrong anyway because the slider will trigger its onChange only when the user drop the cursor so yeah that's good for me :)

chambo-e (approved):

philibea (approved):

PR #6462, opened 2026-05-22
chore(renovate): enable nvm manager and copy extends from lib

chambo-e:
Might be a good idea to have a shared config at this point 😂

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,7 +1,13 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "enabledManagers": ["dockerfile", "github-actions", "npm"],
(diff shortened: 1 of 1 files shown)


philibea (approved):

.github/renovate.json
```diff
@@ -1,7 +1,13 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "enabledManagers": ["dockerfile", "github-actions", "npm"],
```
jsulpis, line 3: why did you remove dockerfile ?

chambo-e: My bad, bad copy paste :)

jsulpis (approved):

PR #6481, opened 2026-05-26
feat(Logo): dark logos

lisalupi:
## Type

- Feature

### Summarize concisely:

`Logos` now support dark versions. When no dark version is provided (i.e. `/assets/dark/NAME.svg` does not exist while `/assets/light/NAME.svg` does), the light version is displayed as a fallback. *Light* assets are required, as components are indexed from `/assets/light/`.


utils/scripts/generate-icons-file.tsx
```diff
@@ -199,37 +246,45 @@ const main = async () => {
...
-          await appendExportToIndex(component.output, generatedName)
-        } catch (error) {
-          console.error('Error writing to file:', error)
+        if (
+          !(
+            file.includes('small') ||
+            file.includes('disabled') ||
+            (file.includes('/dark/') && component.name === 'Logo')
+          )
+        ) {
```
jsulpis, line 253: I think it's better to use the `break` instead of wrapping everything in the `if`, because it reduces the depth of the nested blocks, and in that case it also avoids a negative condition

jsulpis (commented):
- Datadog logo is not very readable in dark, and there is a white version in their [press kit](https://www.datadoghq.com/about/resources/)
- OpenAI logo is invisible in dark, there is a white version in their [brand page](https://openai.com/fr-FR/brand/)
- Owncloud logo is invisible in dark, there is a white version in their [brand page](https://owncloud.com/trademark/) (maybe also update the light version with "A Kiteworks Company" ?)

image
image
image

utils/scripts/generate-icons-file.tsx
lisalupi: won't `break` stop the loop?

jsulpis: right, but there was a `break` before so it's a different behavior no ? if "We do not generate an icon component" I guess it should `break` to generate no file ?

jsulpis (approved):

utils/scripts/generate-icons-file.tsx
lisalupi: I think it used to work before because the "ignored" folders were read before the ones needed ("default" read before "small" and "disabled") but now "dark" is read before "light" so the loop breaks before any component was created for the logos

chambo-e (approved):

PR #6482, opened 2026-05-26
feat(Modal): choose close button sentiment + fix border-radius

jsulpis:
## Type

- Bug
- Feature

#### What is expected?

- correct border-radius on the top when using an image
- ability to have a light close button on dark images

- fix styles
- add a prop `closeButtonSentiment` on the Modal
- update tests

| url  | image | image |

### DIFF
--- a/packages/ui/src/components/Modal/index.tsx
+++ b/packages/ui/src/components/Modal/index.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { useCallback, useContext, useId, useRef, useState } from 'react'
+import { Fragment, useCallback, useContext, useId, useRef, useState } from 'react'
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/Modal/index.tsx
```diff
@@ -46,6 +51,7 @@ export const Modal = ({
   hideOnClickOutside = true,
   hideOnEsc = true,
   isClosable = true,
+  closeButtonSentiment = 'black',
```
lisalupi: Image

we should maybe have white as a default value for dark/darker mode? Black is not very visible

lisalupi (approved):

mhervouet (approved):

PR #6486, opened 2026-05-27
fix(CodeEditor): fix the position of the tooltip on the CopyButton

jsulpis:
## Type

- Bug

#### What is expected?

The tooltip on the CopyButton in the CodeEditor is correctly positionned

(Describe what you did)

- add the positionning style on the div generated by the tooltip
- add a fallback style on the button to avoid breaking it if we forget to update the style when the Tooltip is fixed

| url  | image | image |

### DIFF
--- a/packages/ui/src/compositions/CodeEditor/styles.css.ts
+++ b/packages/ui/src/compositions/CodeEditor/styles.css.ts
@@ -8,19 +8,26 @@ export const disabledStack = style({ cursor: 'not-allowed' })
 const copyButton = style({
   backgroundColor: consoleDarkTheme.colors.neutral.backgroundWeak,
   color: consoleDarkTheme.colors.neutral.textStrong,
-  position: 'absolute',
-  right: theme.space[1],
   selectors: {
     '&:hover': {
       backgroundColor: consoleDarkTheme.colors.neutral.backgroundHover,
       color: consoleDarkTheme.colors.neutral.textStrongHover,
     },
   },
+})
+
+// FIXME: remove this patch when we fix the Tooltip so that it doesn't generate a div around the button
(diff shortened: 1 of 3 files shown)


philibea (approved):

packages/ui/src/compositions/CodeEditor/styles.css.ts
```diff
@@ -8,19 +8,26 @@ export const disabledStack = style({ cursor: 'not-allowed' })
...
     },
   },
+})
+
+// FIXME: remove this patch when we fix the Tooltip so that it doesn't generate a div around the button
+globalStyle(`div[aria-describedby]:has(> ${copyButton})`, {
+  position: 'absolute',
+  right: theme.space[1],
   top: theme.space[1],
 })
```
lisalupi: we tend to avoid globalStyles. You can put this code directly in the popup style (childrenContainer). 
```
[`&:has(> ${codeEditorStyle.copyButton})`]: {...},
```

lisalupi (approved):

packages/ui/src/compositions/CodeEditor/styles.css.ts
jsulpis: ✅

lisalupi: now it seems to break some styles 🫩
Capture d’écran 2026-05-27 à 10 38 47
Capture d’écran 2026-05-27 à 10 39 55

jsulpis: 🤦  thanks
the import of `codeEditorStyle` in the Popup changed the order of the stylesheets and it's a bit complicated to fix. If it's ok for you I reverted the globalStyle, the goal is to remove it anyway

PR #6490, opened 2026-05-27
chore: rename oxc to lint

chambo-e:


lisalupi (approved):

package.json
```diff
@@ -119,7 +116,7 @@
   "lint-staged": {
     "*.(j|t|mj)s?(x)": [
       "pnpm run format",
-      "pnpm run oxc:lint"
+      "pnpm run lint"
```
jsulpis, line 119:
```suggestion
      "pnpm run lint:fix"
```

jsulpis (approved):

package.json
chambo-e: @jsulpis I don't know if we want to autofix on hook, this means code can change without the dev knowing ?

philibea (approved):

package.json
jsulpis: I would say for autofix rules it's not a big deal and it's always what we want but personally I autofix on save anyway so we can leave it like that if you prefer

chambo-e: I have no preference either, I'll let @philibea give its opinion

philibea: we use to not autofix on lint, i think we can keep like that for now. 
I'm ok for an auto-fix on format, but auto-fix of lint rules can be dangerous sometime.

PR #6491, opened 2026-05-27
feat(Radio): small variant

lisalupi:
## Type

- Enhancement

#### What is expected?
`Radio`, `RadioGroup`: new prop "size"


packages/ui/src/components/Radio/styles.css.ts
jsulpis, line 107: the native input is not centered with the visible button, this is fixed with these styles:

```css
    top: 50%;
    transform: translateY(-50%);
```

Image 

Image

jsulpis (approved):
the size of the legend and the space below it doesn't change with the small size ? It looks a bit big

Image

mhervouet (approved):

packages/ui/src/components/Radio/styles.css.ts
```diff
@@ -180,6 +188,29 @@ const innerCircleRing = style({
   },
 })
 
+const focusRing = style({
```
jsulpis, line 191: the focus ring seems too subtle to me, we will have to make it more visible some day. But it's a global issue

jsulpis (approved):

philibea (approved):

PR #6492, opened 2026-05-27
fix: remove small description on inputs

lisalupi:
## Type

- Bug

#### What is expected?
`Description`: inputs should not use small description as it is hard to read

### DIFF
--- a/.changeset/famous-worlds-bow.md
+++ b/.changeset/famous-worlds-bow.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/ui": minor
+---
+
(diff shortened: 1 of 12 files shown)


philibea (approved):

jsulpis (commented):
what is the change that we want precisely:
1. the `size` prop of the `Description` component has no effect, or
2. we don't want to apply it on the inputs ?

if it's 1. we should update the `Description` component so that the size has no effect (always use `caption` font), mark the `size` prop as deprecated and remove the stories, to avoid using it elsewhere

dmzski (approved):

PR #6494, opened 2026-05-28
fix(FileInput): accept

lisalupi:
## Type

- Bug

#### What is expected?
`FileInput`: fix "accept" behavior issue with explicit types ('.mp3', '.png')


Echo-IV (approved):

jsulpis (commented):
it would be good to have a test for that because the bug isn't obvious

jsulpis (approved):

PR #6497, opened 2026-05-29
fix(FileInput): onChange behavior and fileType definition

lisalupi:
## Type

- Bug

#### What is expected?
`FileInput`:
- update FileList type to more closely match default File type. ⚠️ This is a breaking change for typing ⚠️ 
- fix onChange: returns the whole list of files (as a FileList), not just the newly added one.

### DIFF
--- a/packages/ui/src/components/FileInput/__stories__/Multiple.stories.tsx
+++ b/packages/ui/src/components/FileInput/__stories__/Multiple.stories.tsx
@@ -1,16 +1,19 @@
 import type { StoryFn } from '@storybook/react-vite'
 import { FileInput } from '..'
 import { Stack } from '../../Stack'
+import type { FilesType } from '../types'
 
+// oxlint-disable-next-line typescript/no-unsafe-type-assertion
 const defaultFile = [
   {
     file: 'https://upload.wikimedia.org/wikipedia/commons/4/41/Photo_Chat_Noir_et_blanc.jpg',
-    fileName: 'cat.png',
+    name: 'cat.png',
     lastModified: 1,
     size: 30_460,
     type: 'image/png',
   },
-]
+] as FilesType[]
+
 export const Multiple: StoryFn<typeof FileInput> = args => (
   <Stack direction="column" gap={3}>
     <FileInput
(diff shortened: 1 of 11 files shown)


packages/ui/src/components/FileInput/__stories__/Multiple.stories.tsx
```diff
@@ -1,16 +1,19 @@
 import type { StoryFn } from '@storybook/react-vite'
 import { FileInput } from '..'
 import { Stack } from '../../Stack'
+import type { FilesType } from '../types'
 
+// oxlint-disable-next-line typescript/no-unsafe-type-assertion
```
jsulpis: I guess this is because of the `File` native type, do we really need all of its properties ? Could we use `Pick<File, "name" | ...>` instead of the whole `File` in order to avoid the need of `as FilesType[]` and the lint warning ?

packages/ui/src/components/FileInput/index.tsx
```diff
@@ -128,20 +128,24 @@ const FileInputBase = ({
...
-      file: URL.createObjectURL(file),
-      fileName: file.name,
-      lastModified: file.lastModified,
-      size: file.size,
-      type: file.type,
-    }))
+    const formattedNewFiles = newFiles.map(file =>
+      Object.assign(file, {
+        file: URL.createObjectURL(file),
+        fileName: file.name,
```
jsulpis: didn't you remove the `fileName` from the type ? Should it be removed here ?

jsulpis (approved):

packages/ui/src/components/FileInput/index.tsx
lisalupi: yes my bad 🙈

packages/ui/src/components/FileInput/__stories__/Multiple.stories.tsx
lisalupi: I think it's better to have the full File type to ensure that the default API functions work as expected

jsulpis: ok so maybe just for the `defaultFiles` property, allow to pass a subset of a `File` ? This is what you do in the story

lisalupi: it will probably cause an issue with typing inside the component since some props might be undefined

Echo-IV (approved):

PR #6499, opened 2026-05-29
fix: improve guards for selectable card field components

LouisDrame:
## Summary

Improve guards on the selectable card fields components.

## Type

- Enhancement

### Summarize concisely:

In some cases guards weren't strong enough to ensure correct calls.

#### What is expected?

Guards should be stronger and not use truthy values

Edited the guards from xxx ?? [] to Array.isArray and a ternary to remove truthy values that are not an array from the test.

### DIFF
--- a/.changeset/violet-lines-design.md
+++ b/.changeset/violet-lines-design.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/form": patch
+---
+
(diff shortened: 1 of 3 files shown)


jsulpis (commented):
could you add a test for the case that was broken please ?

PR #6509, opened 2026-06-01
chore: group prosemirror updates in Renovate

jsulpis:
## Type

- Refactor

#### What is expected?

tedlt

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -52,6 +52,11 @@
       "matchPackageNames": ["/^@testing-library/"],
       "groupName": "Testing library",
       "groupSlug": "testing-library"
+    },
+    {
+      "matchPackageNames": ["/^prosemirror-/", "@handlewithcare/react-prosemirror"],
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -52,6 +52,11 @@
       "matchPackageNames": ["/^@testing-library/"],
       "groupName": "Testing library",
       "groupSlug": "testing-library"
+    },
+    {
+      "matchPackageNames": ["/^prosemirror-/", "@handlewithcare/react-prosemirror"],
```
jsulpis, line 57: not sure if `@handlewithcare/react-prosemirror` should be in the same group because it's not from the same author, WYT ?

lisalupi (approved):

.github/renovate.json
lisalupi: i'd group everything, that way we only have 1 PR for the whole component

philibea (approved):

PR #6511, opened 2026-06-02
fix(GlobalAlert): close button placement

lisalupi:
## Type

- Bug

#### What is expected?

`GlobalAlert`: fix layout (remove absolute position on close button and fix paddings)


Echo-IV (approved):

jsulpis (approved):
👌

PR #6512, opened 2026-06-02
fix(TextArea): padding and clear button

lisalupi:
## Type

- Bug

#### What is expected?
`TextArea`: fix padding right on content and disable clear button when input is read-only or disabled


jsulpis (approved):
👌

chambo-e (approved):

PR #6516, opened 2026-06-05
fix(SelectInput): alignments

lisalupi:
## Type

- Bug

#### What is expected?
`SelectInput`: 
	- add a transparent border when disabled to avoid sizing differences between read-only, default, and disabled states
	- fix options alignment with checkbox and `OptionalInfo`

Before (no border when disabled):

https://github.com/user-attachments/assets/792e42e2-291e-46dd-9ffc-552324ffacb3

After (border when disabled):

https://github.com/user-attachments/assets/f85c0988-625f-4455-9c9d-839deee0b907

### DIFF
--- a/packages/ui/src/components/SelectInput/components/SelectBar/selectBar.css.ts
+++ b/packages/ui/src/components/SelectInput/components/SelectBar/selectBar.css.ts
@@ -65,6 +65,7 @@ export const selectBarBase = style({
   alignItems: 'center',
   background: theme.colors.neutral.background,
   borderRadius: theme.radii.default,
+  border: `1px solid transparent`,
   boxShadow: 'none',
   cursor: 'pointer',
   display: 'grid',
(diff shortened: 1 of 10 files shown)


jsulpis (commented):
the alignment is fixed for the "left" placement but not for the "right" : the elements are pushed at the top and not vertically aligned

Image

Image

lisalupi:
> the alignment is fixed for the "left" placement but not for the "right" : the elements are pushed at the top and not vertically aligned
> Image Image

I was told to align with the top, not center

packages/ui/src/components/SelectInput/components/SelectBar/selectBar.css.ts
```diff
@@ -107,7 +108,7 @@ export const selectBar = recipe({
     {
       style: {
         selectors: {
-          "&:not([data-disabked='true'])": {
+          "&:not([data-disabled='true'])": {
```
jsulpis: I think the selector can be deleted and the style applied directly, since a select can't (or shouldn't) be open when disabled (idem below)

jsulpis (approved):

JulienSaguez (approved):

radhi-nasser-scaleway (approved):

mhervouet (approved):

PR #6528, opened 2026-06-08
fix: add step in numberInputField storybook

PierreBertinet:
## Type

- Bug

### Summarize concisely:

Add missing step props in NumberInputField storybook

#### What is expected?

Step should be visible in playground

(Describe what you did)

1.

2.

3.

### DIFF
--- a/.changeset/lazy-corners-burn.md
+++ b/.changeset/lazy-corners-burn.md
@@ -1,5 +1,0 @@
----
-"@ultraviolet/form": patch
----
-
-Add step in numberInputField playground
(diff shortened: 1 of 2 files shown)


.changeset/lazy-corners-burn.md
```diff
@@ -0,0 +1,5 @@
+---
```
lisalupi: a changeset is not necessary in that case :)

jsulpis (approved):

lisalupi (approved):

.changeset/lazy-corners-burn.md
PierreBertinet: removed

PR #6529, opened 2026-06-09
refactor: breaking: change NumberInputField isInteger handling

PierreBertinet:
## Type

- Refactor

### Summarize concisely:

Change NumberInputField isInteger error handling to pass boolean instead of step that does not handle every cases

#### What is expected?

this is a breaking change, project using this library and using NumberInputField's isInteger form error function will have to update their codebase to use isInteger as boolean and isNumber as boolean

### DIFF
--- a/.changeset/empty-rings-relate.md
+++ b/.changeset/empty-rings-relate.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/form": patch
+---
+
+breaking: change NumberInputField isInteger error handling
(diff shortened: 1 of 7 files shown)


jsulpis:
> does not handle every cases

What cases are not handled ? Could you add tests to document the edge cases please ?

PierreBertinet:
> > does not handle every cases
> 
> What cases are not handled ? Could you add tests to document the edge cases please ?

well, in shire, we use this method to determine which text we have to send for the helper. But.. isInteger (despite starting with is) is a number or a string or undefined, and is linked to step, not the actual value. So this is a refactor to send clear informations

PierreBertinet:
duplicate of https://github.com/scaleway/ultraviolet/pull/6535

PR #6531, opened 2026-06-10
feat(EstimateCost): allow to pass children instead of label/image

chambo-e:
## Type

- Feature

### DIFF
--- a/packages/ui/src/compositions/EstimateCost/Components/Zone.tsx
+++ b/packages/ui/src/compositions/EstimateCost/Components/Zone.tsx
@@ -1,7 +1,7 @@
 'use client'
 
 import { memo } from 'react'
-import type { CSSProperties } from 'react'
+import type { CSSProperties, ReactNode } from 'react'
 import { useEstimateCost } from '../EstimateCostProvider'
 import type { BareEstimateProduct, EstimateProduct, Iteration } from '../types'
 import { Item } from './Item'
@@ -10,7 +10,7 @@ import { estimateCostStyle } from '../styles.css'
 
 type RegionProps = {
   shouldBeHidden?: boolean
-  priceText?: string
+  priceText?: ReactNode
   animated?: boolean
   isFirstElement?: boolean
   isLastElement?: boolean
@@ -20,12 +20,18 @@ type RegionProps = {
   }
   iteration?: Iteration
   discount?: number
-  label: string
-  image: string
   noBorder?: boolean
   noPrice?: boolean
   style?: CSSProperties
-}
+} & XOR<
+  [
+    {
+      label: string
+      image: string
+    },
+    { children: ReactNode },
+  ]
+>
 
 export const Zone = memo(
   ({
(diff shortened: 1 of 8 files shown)


lisalupi (approved):

packages/ui/src/compositions/EstimateCost/__tests__/Zone.test.tsx
```diff
@@ -19,10 +19,19 @@ describe('estimateCost - Zone', () => {
       </EstimateCost>,
     ))
 
-  it('render region component, with animation', () =>
```
jsulpis, line 22: I think the associated snapshot is still present, it should be deleted

packages/ui/src/compositions/EstimateCost/Components/Zone.tsx
```diff
@@ -10,7 +10,7 @@ import { estimateCostStyle } from '../styles.css'
 
 type RegionProps = {
```
jsulpis:
```suggestion
type ZoneProps = {
```

jsulpis (commented):
could you update the Region story to use the children instead of image/label so we can get a preview ? The img lost its style so I think it might be broken

radhi-nasser-scaleway (approved):

packages/ui/src/compositions/EstimateCost/__tests__/Zone.test.tsx
chambo-e: I've checked and it was already properly removed by vitest :)

chambo-e:
> could you update the Region story to use the children instead of image/label so we can get a preview ? The img lost its style so I think it might be broken

@jsulpis I've completed the Region story with the children.

Wdym with "the img lost its style" ? I don't see any changes on my side, do you have a screen ? 🤔

jsulpis:
> > could you update the Region story to use the children instead of image/label so we can get a preview ? The img lost its style so I think it might be broken
> 
> @jsulpis I've completed the Region story with the children.
> 
> Wdym with "the img lost its style" ? I don't see any changes on my side, do you have a screen ? 🤔

I meant that the `img` inside of the `Region` component has a style applied to it, so if we put the same `img` in the children it won't have this style applied. But I forgot that it should be used with the `Flag` component.

Could you still update the `Region` story to replace at least one item with a children using the `Flag` component, so that we can see how to achieve the same visual as the image/label props ?

chambo-e:
> > > could you update the Region story to use the children instead of image/label so we can get a preview ? The img lost its style so I think it might be broken
> > 
> > 
> > @jsulpis I've completed the Region story with the children.
> > Wdym with "the img lost its style" ? I don't see any changes on my side, do you have a screen ? 🤔
> 
> I meant that the `img` inside of the `Region` component has a style applied to it, so if we put the same `img` in the children it won't have this style applied. But I forgot that it should be used with the `Flag` component.
> 
> Could you still update the `Region` story to replace at least one item with a children using the `Flag` component, so that we can see how to achieve the same visual as the image/label props ?

I noticed by reading your comment that I can directly achieve what I want to do with EstimateCost.Item ... I'm gonna close this pr 😄

PR #6532, opened 2026-06-11
refactor: extract useDebouncedCallback hook

jsulpis:
## Type

- Refactor

### Summarize concisely:

I need this hook for the filters so I extracted what I found in the repo. There is a debounce in the Popup component but it cannot be quickly refactored so I didn't update it.

#### What is expected?

hook `useDebouncedCallback` available in `@ultraviolet/utils`

- create hook `useDebouncedCallback` 
- use it in `InfiniteScroll` component


packages/utils/src/debounceCallback.ts
```diff
@@ -0,0 +1,32 @@
+import { useRef, useEffect, useCallback } from 'react'
```
chambo-e, line 1: Shouldn't this be in scw-lib ? (sorry) 😄 

If yes this can be done after, I can add it to my todolist :)

chambo-e: (we use https://www.npmjs.com/package/use-debounce everywhere, we don't want to use it here also ?)

jsulpis: either way, in UV I would say if we just need the debounceCallback we can avoid adding this dependency

chambo-e (approved):

lisalupi (approved):

PR #6533, opened 2026-06-11
feat: add Filters component

jsulpis:
## Type

- Feature

### Summarize concisely:

Add the Filters component with documentation. Tests will be added in another PR so that this one isn't even bigger.

#### What is expected?

Usable Filters

- add the component
- add docs


packages/ui/src/compositions/Filters/types.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+export type FilterConfigItemSlider = FilterConfigItemBase &
+  Pick<ComponentProps<typeof Slider>, 'options' | 'step' | 'unit' | 'min' | 'max' | 'double'> & {
+    type: 'slider'
+  }
+
+type RelativePreset = {
+  label: string
+  value: string
```
lisalupi, line 58: can't it be undefined?

packages/ui/src/compositions/Filters/types.ts
```diff
@@ -0,0 +1,110 @@
...
+  expanded?: boolean
+}
+
+export type FilterConfig<V extends AnyObject = AnyObject> = FilterConfigGroup<V> | FilterConfigItem<V>
+
+export type FilterComponentProps<
+  // oxlint-disable-next-line typescript/no-explicit-any
+  Value = any,
+  // oxlint-disable-next-line typescript/no-explicit-any
+  Config = any,
```
lisalupi, line 103: we can't do better than any?

packages/ui/src/compositions/Filters/filterTypes/FilterDatetimeRange/FilterDatetimeRange.tsx
```diff
@@ -0,0 +1,65 @@
+import { useMemo, useState } from 'react'
```
lisalupi, line 1: Image

issue when selecting dates (probably missing a max-width)

packages/ui/src/compositions/Filters/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+      'keyboard-focus': false,
+      'contrast-visuals': false,
+      'semantics-screen-reader': false,
+      'pointer-touch': false,
+      'specific-patterns': false,
+    },
+  },
+} as Meta
+
+export { Demo } from './Demo.stories'
```
lisalupi: we usually have a "Playground" story with default props first

packages/ui/src/compositions/Filters/__stories__/Demo.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+          clear: 'Clear',
+          clearAll: 'Clear all',
+          seeAll: 'All filters',
+          drawerHeader: 'Filters',
+          submit: 'See results',
+        }}
+        onChange={setLiveValues}
+        onSubmit={setSubmittedValues}
+      />
+      <Stack direction="row" gap="4">
```
lisalupi: maybe use the Snippet component to display this data, like we do in form stories ? (for instance [here](https://storybook.ultraviolet.scaleway.com/?path=/story/form-components-fields-textinputfield--playground)) It might be easier to read

lisalupi (changes requested):

packages/ui/src/compositions/Filters/parts/FiltersDrawer.tsx
```diff
@@ -0,0 +1,111 @@
...
+      hideOnClickOutside={false}
+      onClose={() => {
+        filters.discard()
+        closeDrawer()
+      }}
+      open={isDrawerOpen}
+      separator
+      size="small"
+    >
+      <form
```
philibea, line 56: question: 
should we provide this filter inside ours ultraviolet/form package ?

packages/ui/src/compositions/Filters/types.ts
jsulpis: no the presets have to define a value, ex:
```ts

relativePresets: [
  { relativeTime: 60 * 15 * 1000, value: 'last_15min', label: 'Last 15 minutes' },
  { relativeTime: 60 * 60 * 1000, value: 'last_hour', label: 'Last hour' },
  { relativeTime: 60 * 60 * 24 * 1000, value: 'last_24hours', label: 'Last 24 hours' },
  { relativeTime: 60 * 60 * 24 * 30 * 1000, value: 'last_30days', label: 'Last 30 days' },
  { relativeTime: 60 * 60 * 24 * 90 * 1000, value: 'last_90days', label: 'Last 90 days' },
],
```

packages/ui/src/compositions/Filters/types.ts
jsulpis: If I use `unknown` we have to do a type cast on the project side when passing custom components to the `Filters`, it's not very user-friendly. The custom components can be defined using real types for Value and Config so I think it's ok to have any by default

packages/ui/src/compositions/Filters/__stories__/index.stories.tsx
jsulpis: ✅

packages/ui/src/compositions/Filters/__stories__/Demo.stories.tsx
jsulpis: ✅

packages/ui/src/compositions/Filters/parts/FiltersDrawer.tsx
jsulpis: I would say that while the `Filters` could use a form internally or behave like a form, it's not a "field" component like all the others that are in the form package. We can use the Filters outside a form so I would keep it in the ui package

packages/ui/src/compositions/Filters/filterTypes/FilterDatetimeRange/FilterDatetimeRange.tsx
jsulpis: ✅  I updated the layout for this story

lisalupi: Capture d’écran 2026-06-15 à 15 54 00
it is less visible now but i still have the issue

packages/ui/src/compositions/Filters/__stories__/Context.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+                filters.reset()
+                filters.submit(filters.defaultValues)
+              }}
+            >
+              Reset from outside of the component
+            </Button>
+          </>
+        )}
+      </FiltersProvider>
+      <Stack direction="row" gap="4">
```
lisalupi, line 35: same here, I think a Snippet could be helpful

packages/ui/src/compositions/Filters/__stories__/Dates.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+      startAt: null,
+      endAt: null,
+    },
+    name: 'John Doe',
+  },
+  labels: {
+    clear: 'Clear',
+    clearAll: 'Clear all',
+    seeAll: 'All filters',
+    drawerHeader: 'Filters',
```
lisalupi: I see that a text for drawer header is defined but we cannot open a drawer in the story :/

packages/ui/src/compositions/Filters/parts/FiltersMainRow.tsx
```diff
@@ -0,0 +1,100 @@
...
+    filters.submit(filters.defaultValues)
+  }
+
+  return (
+    <Row
+      className={className}
+      style={{
+        background: theme.colors.neutral.backgroundWeak,
+        borderRadius: theme.radii.default,
+      }}
```
lisalupi: Image

not sure about contrast in dark/darker mode (labels/background)

\+ we do not use style, only classNames + vanilla extract

packages/ui/src/compositions/Filters/filterTypes/FilterMultiSelect.tsx
```diff
@@ -0,0 +1,70 @@
+import { useMemo } from 'react'
+import { CheckboxGroup } from '../../../components/CheckboxGroup'
+import { Label } from '../../../components/Label'
+import { SelectInput } from '../../../components/SelectInput'
+import type { FilterConfigItemMultiSelect, AnyObject, FilterComponentProps } from '../types'
+
+const DISPLAY_THRESHOLD = 10
```
lisalupi: maybe put all consts in a constants.ts file? I see that DISPLAY_THRESHOLD is re-defined somewhere else

packages/ui/src/compositions/Filters/__stories__/demo.config.ts
```diff
@@ -0,0 +1,76 @@
...
+    items: [
+      {
+        type: 'select',
+        name: 'status',
+        label: 'Status',
+        placeholder: 'Select status',
+        options: [
+          { label: 'Active', value: 'active' },
+          { label: 'Inactive', value: 'inactive' },
+        ],
```
lisalupi, line 43: issue in dark mode, the text doesn't change color in the drawer: probably a `Text` without `sentiment = "neutral"` somewhere (Also, it is probably only a problem in storybook, but you might need to check to be sure)

Image

packages/ui/src/compositions/Filters/filterTypes/FilterSlider.tsx
```diff
@@ -0,0 +1,22 @@
+import { Slider } from '../../../components/Slider'
+import type { FilterConfigItemSlider, FilterComponentProps } from '../types'
+
+type FilterSliderProps = FilterComponentProps<number | number[], FilterConfigItemSlider>
+
+export const FilterSlider = ({ value, onChange, config, hideLabel }: FilterSliderProps) => (
+  // @ts-expect-error the Slider is a bit too strict with the type of the value for this use case
```
philibea: question: should we throw an error in case the config is not align with the value.


```typescript
if(Array.isArray(value) && !config.double || config.double && !Array.isArray(value)){
throw new Error('config issue') 
}
```

philibea: or at least we can 
```typescript
if (Array.isArray(value) && config.double) {
return ( <Slider
             double={true}
              value={value}
               ....{props}
               />
}
if(!Array.isArray(value) && !config.double{
  return ( <Slider
                   double={false} 
                   value={value}
                   ...{props}
                    />
              )
}

return null 
```

lisalupi (changes requested):

packages/ui/src/compositions/Filters/parts/FiltersMainRow.tsx
lisalupi: Capture d’écran 2026-06-15 à 16 22 48
same issue with the sliders (dark mode in the screenshot)

packages/ui/src/compositions/Filters/parts/AbstractFilter.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onChange,
+  value,
+  values,
+  customComponents = {},
+  size,
+}: AbstractFilterProps) => {
+  const shouldHideLabel = directionContext === 'column' && hideLabel
+  const availableComponents = { ...defaultFieldComponents, ...customComponents }
+
+  function isAvailableType(configItem: unknown): configItem is { type: keyof typeof availableComponents } {
```
lisalupi:
```suggestion
  const isAvailableType = (configItem: unknown): configItem is { type: keyof typeof availableComponents } => {
```
we tend to prefer arrow fucntion in the code
(same comment everywhere, I will not comment on every single function you did)

packages/ui/src/compositions/Filters/parts/ExpandableFilterGroup.tsx
```diff
@@ -0,0 +1,52 @@
+import { ArrowDownIcon, ArrowUpIcon } from '@ultraviolet/icons'
+import type { ReactNode } from 'react'
+import { useState } from 'react'
+import { Button } from '../../../components/Button'
+import { Label } from '../../../components/Label'
+import { Row } from '../../../components/Row'
+import { Stack } from '../../../components/Stack'
+import { Text } from '../../../components/Text'
+
+export const ExpandableFilterGroup = ({
```
lisalupi, line 10: we already have an `Expandable` component. Was it not working correctly?

packages/ui/src/compositions/Filters/parts/FiltersDrawer.tsx
```diff
@@ -0,0 +1,111 @@
...
+        id={formId}
+        onSubmit={(e: SubmitEvent) => {
```
lisalupi:
```suggestion
        onSubmit={(event: SubmitEvent) => {
```

packages/ui/src/compositions/Filters/parts/FiltersMainRow.tsx
```diff
@@ -0,0 +1,100 @@
...
+      padding={theme.space[2]}
```
lisalupi: maybe put in in the css.ts file with other custom style, so you don't have to import and use useTheme in this file

packages/ui/src/compositions/Filters/parts/helpers.ts
```diff
@@ -0,0 +1,75 @@
...
+    // Typical cse for string | number
+    if (valueA === valueB) {
+      return false
+    }
+
+    if (Array.isArray(valueA) && Array.isArray(valueB)) {
+      if (valueA.length !== valueB.length) {
+        return true
+      }
+      // Tuples for Slider, in case of slect input it can cause a rerender if they are not sorted the same way but I prefer to keep 1 useless rerender
```
lisalupi:
```suggestion
      // Tuples for Slider, in case of select input it can cause a rerender if they are not sorted the same way but I prefer to keep 1 useless rerender
```
and maybe avoid "I" ^^

packages/ui/src/compositions/Filters/parts/helpers.ts
```diff
@@ -0,0 +1,75 @@
...
+      // Instead of making this functions more complicated
+      if (valueA.length === 2 && valueB.length === 2 && (valueA[0] !== valueB[0] || valueA[1] !== valueB[1])) {
+        return true
+      }
+      // Multiple select input check all match
+      if (!valueA.every(aValueInValueA => valueB.includes(aValueInValueA))) {
+        return true
+      }
+      return false
```
lisalupi:
```suggestion
      return (!valueA.every(aValueInValueA => valueB.includes(aValueInValueA)))
```

packages/ui/src/compositions/Filters/parts/reducer.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export type FiltersAction<V extends AnyObject> =
+  | { type: 'SET_VALUE'; payload: { name: keyof V; value: unknown } }
+  | { type: 'SET_VALUES'; payload: { values: Partial<V> } }
+  | { type: 'RESET_FIELD'; payload: { name: keyof V; defaultValue?: V[keyof V] } }
+  | { type: 'RESET_FIELDS'; payload: { names: (keyof V)[] } }
+  | { type: 'RESET'; payload: { values?: Partial<V> } }
+  | { type: 'SUBMIT' }
+  | { type: 'DISCARD' }
```
lisalupi, line 17: I am not a big fan of single-letter types

lisalupi (changes requested):

packages/ui/src/compositions/Filters/filterTypes/FilterDatetimeRange/FilterDatetimeRange.tsx
jsulpis: I updated it but it's broken on mobile anyway, because of a div added by the `Popup` that can't be styled

image

packages/ui/src/compositions/Filters/filterTypes/FilterSlider.tsx
jsulpis: It's just the value `number | number[]` that causes a type issue, but the config prop is correctly typed. 

And I actually wanted to remove the duplication that we had before (`if (condition) return ... else return ...`)

packages/ui/src/compositions/Filters/__stories__/Context.stories.tsx
jsulpis: oops missed this one, fixed ✅

packages/ui/src/compositions/Filters/__stories__/Dates.stories.tsx
jsulpis: ✅  I updated the types to make all labels optional except `clearAll`. I used a type guard in the `Filters` component to render the Drawer only if its labels are present (otherwise there was a type issue)

packages/ui/src/compositions/Filters/parts/FiltersMainRow.tsx
jsulpis: ✅ it looks like the theme from `useTheme` always rendered in light mode, it's fixed when using vanilla extract 

image

packages/ui/src/compositions/Filters/filterTypes/FilterMultiSelect.tsx
jsulpis: I just exported the const from the `FilterSelect` and use it in `FilterMultiSelect` if it's just one constant

packages/ui/src/compositions/Filters/__stories__/demo.config.ts
jsulpis: ✅  there was a missing sentiment on the label of the Radio component and the `dialog` element applied a dark color, I added the sentiment so now our neutral text color is used

packages/ui/src/compositions/Filters/parts/AbstractFilter.tsx
jsulpis: with the formatting rule that expects no block statement I find this less readable but I will follow the habits
```ts
const isAvailableType = (configItem: unknown): configItem is { type: keyof typeof availableComponents } =>
    (configItem as FilterConfig).type in availableComponents
```

packages/ui/src/compositions/Filters/parts/ExpandableFilterGroup.tsx
jsulpis: I just extracted the components from shire, I didn't rework everything as it would be too much work.

We could update this component later to use the `Expandable` with an animation maybe, but without the animation the `Expandable` component is currently useless (and not accessible because doesn't accept an `id` to use with `aria-controls`)

packages/ui/src/compositions/Filters/parts/FiltersDrawer.tsx
jsulpis: ✅

packages/ui/src/compositions/Filters/parts/FiltersMainRow.tsx
jsulpis: ✅

packages/ui/src/compositions/Filters/parts/helpers.ts
jsulpis: ✅

packages/ui/src/compositions/Filters/parts/helpers.ts
jsulpis: ✅  `I` is not even me, it's whoever wrote this code that I just copied/pasted 😅

packages/ui/src/compositions/Filters/filterTypes/FilterSlider.tsx
philibea: you can write this  to avoid the `@ts-expect-error` which can hide later issue on type here.

```typescript
type FilterSliderProps =
  | FilterComponentProps<number[], FilterConfigItemSlider & { double: true }>
  | FilterComponentProps<number[], FilterConfigItemSlider & { double: false | undefined }>
```

packages/ui/src/compositions/Filters/parts/reducer.ts
philibea: it's a standard inside type generic to avoid too much noise on the code, but here we only have `V` which can be `Values`

jsulpis: you mean `V` ? It's a convention for generic types in all programming languages afaik. I used `V` everywhere for `Values`. In places where there are multiple generic types like `FilterComponentProps<Value, Config, Values>` I used full words but even there I think it's common to have single-letter types

packages/ui/src/compositions/Filters/index.ts
```diff
@@ -0,0 +1,17 @@
+export { Filters, type FiltersProps } from './Filters'
```
philibea:
```suggestion
export { Filters } from './Filters'
export type { FiltersProps } from './Filters'
```

jsulpis: ✅

packages/ui/src/compositions/Filters/filterTypes/FilterSlider.tsx
jsulpis: the type doesn't work, I updated the component with if/else

philibea (approved):

lisalupi (approved):

PR #6534, opened 2026-06-11
fix(Tabs): add focus ring

lisalupi:
## Type

- Enhancement

#### What is expected?
`Tabs`: add focus ring

### DIFF
--- a/packages/ui/src/components/Tabs/styles.css.ts
+++ b/packages/ui/src/components/Tabs/styles.css.ts
@@ -63,19 +63,21 @@ const button = style({
   outline: 'none',
   padding: `${theme.space[1]} ${theme.space[2]}`,
   selectors: {
-    '&:focus-visible': {
-      outline: 'auto',
-    },
     '&:hover, &:active, &:focus': {
       outline: 'none',
       textDecoration: 'none',
     },
-    '&[aria-disabled="false"]:not(:disabled):focus, &[aria-disabled="false"]:not(:disabled):hover, &[aria-disabled="false"]:not(:disabled):active':
-      {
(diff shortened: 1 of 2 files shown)


jsulpis (changes requested):
there is still the issue that when clicking on a tab and releasing outside, two tabs are in active state. Did you check if that can be fixed without too much work ?

lisalupi:
> there is still the issue that when clicking on a tab and releasing outside, two tabs are in active state. Did you check if that can be fixed without too much work ?

fixed

radhi-nasser-scaleway (approved):

JulienSaguez (approved):

packages/ui/src/components/Tabs/styles.css.ts
```diff
@@ -63,19 +63,21 @@ const button = style({
...
-        outline: 'none',
-      },
+    '&[aria-disabled="false"]:not(:disabled):hover': {
+      borderBottomColor: theme.colors.primary.border,
+      color: theme.colors.primary.text,
+      outline: 'none',
+    },
+    '&[aria-disabled="false"]:not(:disabled):focus-visible': {
+      boxShadow: `inset ${theme.shadows.focusPrimary}`,
+      borderBottomColor: theme.colors.primary.border,
```
jsulpis, line 77: it's out of the scope here but I think that the border-bottom should only be visible on the `active` state, otherwise we can't differentiate the active tab when we hover/focus another  one

Image

jsulpis (approved):

philibea (approved):

mhervouet (approved):

PR #6535, opened 2026-06-11
fix(NumberInputField): correct validator functions

jsulpis:
## Type

- Bug

### Summarize concisely:

- Fix the `isInteger` validator function
- add `isNumber` validator

#### What is expected?

- `isInteger` correctly checks that the value is an integer
- another validator `isNumber` checks that the value is a decimal number

Duplicate #6529


packages/form/src/components/Form/defaultErrors.ts
```diff
@@ -14,6 +14,8 @@ export const defaultErrors: RequiredErrors = {
   required: () => '',
   setValueAs: () => '',
   shouldUnregister: () => '',
+  isInteger: () => 'isInteger error',
+  isNumber: () => 'isNumber error',
```
philibea: question: all others fields are initialize to empty string, should we do the same for isInteger and isNumber. Or should we fix the empty value ?

jsulpis: oops I think I did that when debugging, I can reset to an empty string like the others

lisalupi (approved):

packages/form/src/components/Form/defaultErrors.ts
jsulpis: FYI these `defaultErrors` are used in the `Form` component itself but in the stories and tests we use `mockErrors` which contain real messages

PierreBertinet (approved):

philibea (approved):

philibea (approved):

PR #6536, opened 2026-06-11
fix(rich-text-input): spread changement value for richtextinput compo…

JulienSaguez:
## Type

- Bug

### Summarize concisely:
- Fix the spread prop value

#### What is expected?

- Value change in RichTextInput if the form reset the value

### DIFF
--- a/packages/ui/src/compositions/RichTextInput/index.tsx
+++ b/packages/ui/src/compositions/RichTextInput/index.tsx
@@ -6,7 +6,7 @@ import { theme } from '@ultraviolet/themes'
 import { cn } from '@ultraviolet/utils'
 import { assignInlineVars } from '@vanilla-extract/dynamic'
 import type { EditorState } from 'prosemirror-state'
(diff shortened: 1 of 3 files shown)


jsulpis (changes requested):
please add a test to document the expected behavior because the description is not clear

packages/ui/src/compositions/RichTextInput/index.tsx
```diff
@@ -81,6 +81,18 @@ export const RichTextInput = ({
 
   const [editorState, setEditorState] = useState<EditorState>(() => createEditorState(docFromHtml(value, editorSchema)))
 
+  useEffect(() => {
```
philibea, line 84: please can you add a test which will explain the why ?

jsulpis (approved):

lisalupi (approved):

radhi-nasser-scaleway (approved):

PR #6537, opened 2026-06-11
fix(FileInputField): label and aria label

lisalupi:
## Type

- Bug

#### What is expected?

`FileInput`: fix label type (`label` was mandatory even with aria-label and for overlay variant which do not display the label)

### DIFF
--- a/packages/form/src/components/FileInputField/index.tsx
+++ b/packages/form/src/components/FileInputField/index.tsx
@@ -11,9 +11,15 @@ type FileInputFieldProps<TFieldValues extends FieldValues, TFieldName extends Fi
   TFieldValues,
   TFieldName
 > &
-  Omit<ComponentProps<typeof FileInput>, 'error' | 'name' | 'onChangeFiles' | 'label' | 'aria-label'> & {
-    label: string
-  }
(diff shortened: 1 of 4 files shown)


jsulpis (approved):

packages/form/src/components/FileInputField/index.tsx
```diff
@@ -36,10 +42,13 @@ const FileInputFieldBase = <
   bottom,
   children,
   errorLabel,
+  'aria-label': ariaLabel,
   ...props
 }: FileInputFieldProps<TFieldValues, TFieldName>) => {
   const { getError } = useErrors()
 
+  const computedLabel = label ?? ariaLabel ?? name
```
philibea, line 50: suggestion: 
we can maybe create a computedLabel helper which will allow us to have the same pattern for all ours Inputs ?

There is something to share with the error props also as it's used the computed label.

Also i think the XOR type for label and aria-label should be centralized to avoid duplication

NB: we can create a ticket for this.

lisalupi: we already have a Label type but in ultraviolet ui and label is defined as a ReactNode,  but in form it should be a string
I can create one for all of ultraviolet form, but maybe in a new pr so i can merge this first?

radhi-nasser-scaleway (approved):

chambo-e (approved):

PR #6541, opened 2026-06-12
feat(TextArea): prop size

lisalupi:
## Type

- Feature

`TextArea`: new prop `size`


packages/ui/src/components/TextArea/index.tsx
jsulpis, line 241: missing prop `size` to forward to the `Description` (and probably the `Text` in the `Notice`)

jsulpis (approved):

packages/ui/src/components/TextArea/index.tsx
lisalupi: no we don't change description size in inputs

chambo-e (approved):

PR #6551, opened 2026-06-15
fix(OfferList): banner and border

lisalupi:
## Type

- Bug

#### What is expected?
`OfferList`: fix issues with banner and borders

| url  |Capture d’écran 2026-06-15 à 17 38 44 |Capture d’écran 2026-06-15 à 17 39 03 |

### DIFF
--- a/packages/ui/src/compositions/OfferList/styles.css.ts
+++ b/packages/ui/src/compositions/OfferList/styles.css.ts
@@ -37,7 +37,7 @@ const bannerText = recipe({
   base: {
     border: `1px solid ${theme.colors.neutral.border}`,
     borderRadius: `0 0 ${theme.radii.default} ${theme.radii.default}`,
-    borderTopColor: 'transparent',
+    borderTopWidth: 0,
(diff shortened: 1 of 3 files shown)


etienne-scaleway (approved):

packages/ui/src/compositions/OfferList/styles.css.ts
```diff
@@ -97,6 +95,11 @@ const bannerCell = style({
   width: '100%',
   borderBottom: '1px solid transparent',
   transition: 'box-shadow 200ms ease, border-color 200ms ease',
+  selectors: {
+    '&&': {
+      borderLeft: 'none', // Need to add specificity to overload List.Cell style
```
jsulpis: there is still an unwanted border on the right, this can be fixed with `borderInline`. I added a note in the issue to mutualize List and OfferList so that we fix that properly

Image

```suggestion
      borderInline: 'none', // Need to add specificity to overload List.Cell style
```

jsulpis (approved):

PR #6552, opened 2026-06-15
feat(SelectInput): export OptionType and DataType

lisalupi:
## Type

- Feature

#### What is expected?
Export `OptionType` and `DataType` from `SelectInput`

### DIFF
--- a/packages/ui/src/components/index.ts
+++ b/packages/ui/src/components/index.ts
@@ -75,3 +75,5 @@ export { Tooltip } from './Tooltip'
 export { TreeMapChart } from './TreeMapChart'
 export { UnitInput } from './UnitInput'
 export { VerificationCode } from './VerificationCode'
+
+export type { OptionType, DataType } from './SelectInput/types'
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/index.ts
```diff
@@ -75,3 +75,5 @@ export { Tooltip } from './Tooltip'
 export { TreeMapChart } from './TreeMapChart'
 export { UnitInput } from './UnitInput'
 export { VerificationCode } from './VerificationCode'
+
+export type { OptionType, DataType } from './SelectInput/types'
```
philibea, line 79: you should probably need to export inside `packages/ui/src/index.ts`

export type * from './components'

lisalupi: Isn't `export * from './components'` enough? I tried to use this type in a example (next) and it seemed ok

jsulpis (approved):

philibea (approved):

PR #6558, opened 2026-06-16
chore: add a11y-audit skill

jsulpis:
## Type

- Feature

### Summarize concisely:

Add a11y-audit skill to audit components

#### What is expected?

Usable skill to conduct accessibility audits

- add a11y-audit skill
- run it on the Alert component


philibea (approved):

.agents/skills/a11y-audit/resources/audit-template.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+{/* Rules that can be enforced by the component implementation */}
+
+To apply when using the component:
+
+{/* Rules that depend on how the component is used by developers */}
+
+## Accessibility issues
+
+{/* prettier-ignore */}
```
lisalupi, line 36: we don't use prettier in the codebase

lisalupi (approved):

.agents/skills/a11y-audit/resources/audit-template.mdx
jsulpis: `oxfmt-ignore` doesn't work and oxfmt escapes a character which breaks the comment. With `prettier-ignore` it effectively disables oxfmt. They mention prettier in their docs, it's not very clear when to use which 
https://oxc.rs/docs/guide/usage/formatter/ignore-comments.html#inline-ignore-comments

PR #6578, opened 2026-06-18
feat(code-editor): add onclick prop for code editor component

JulienSaguez:
## Type

- Feature

### Summarize concisely:

Add onclick prop for code editor component

### What is expected?
We need this callback for to display a modal window when the code editor is clicked


jsulpis:
Hi, could you explain why you need this callback please ?

JulienSaguez:
> Hi, could you explain why you need this callback please ?

Yes, i add a description why we need this callback

jsulpis:
> > Hi, could you explain why you need this callback please ?
> 
> Yes, i add a description why we need this callback

ok, don't forget keyboard users, I think it should be `onFocus` instead. Would that work for you ?

PS: this is why we ask that you tell us about the changes in Slack first, we would have spared two round-trips 😉

JulienSaguez:
> > > Hi, could you explain why you need this callback please ?
> > 
> > 
> > Yes, i add a description why we need this callback
> 
> ok, don't forget keyboard users, I think it should be `onFocus` instead. Would that work for you ?
> 
> PS: this is why we ask that you tell us about the changes in Slack first, we would have spared two round-trips 😉

Yes this is perfect. Sorry, we ask you in Slack first for the next time

lisalupi (approved):

packages/ui/src/compositions/CodeEditor/__tests__/index.test.tsx
```diff
@@ -99,4 +99,20 @@ describe('codeEditor', () => {
...
+      <CodeEditor
+        data-testid="code-editor"
+        onFocus={onFocus}
+        value="configuration: 1/ntest: 'ok'"
+        extensions="yaml"
+        onChange={newValue => newValue}
+      />,
+    )
+    const codeEditor = screen.getByTestId('code-editor')
+    fireEvent.focus(codeEditor)
```
philibea: avoid fireEvent if possible and use userEvent with a click on this input should trigger focus.

philibea (approved):

PR #6579, opened 2026-06-18
feat(font): add bundled fonts

philibea:
## Type

- Feature

- Gives users the option to self-host fonts for better privacy/control
- Reduces external dependencies to Scaleway's CDN
- No additional configuration needed for Vite/Next.js projects


packages/fonts/package.json
```diff
@@ -13,34 +13,38 @@
...
   "os": [
     "darwin",
     "linux"
   ],
   "type": "module",
+  "sideEffects": [
+    "**/*.css",
+    "**/*.css.ts"
+  ],
+  "types": "./index.d.ts",
```
jsulpis: there is no `.css.ts` or `index.d.ts`
```suggestion
  "sideEffects": [ "**/*.css" ],
```

packages/fonts/vite.config.ts
```diff
@@ -1,15 +1,20 @@
...
       rolldownOptions: {
-        input: path.resolve(import.meta.dirname, 'src/fonts.css'),
+        input: {
+          'fonts-cdn': path.resolve(import.meta.dirname, 'src/fonts-cdn.css'),
+          'fonts-bundled': path.resolve(import.meta.dirname, 'src/fonts-bundled.css'),
+        },
         output: {
           assetFileNames: '[name][extname]',
-          dir: path.resolve(import.meta.dirname, 'dist'),
+          dir: outputDir,
```
jsulpis: could be inlined like the other paths instead of using a variable
edit: just revert ?

examples/vite/src/main.tsx
```diff
@@ -1,6 +1,7 @@
 import { StrictMode } from 'react'
 import { createRoot } from 'react-dom/client'
 import { App } from './App.tsx'
+import '@ultraviolet/fonts/fonts.css'
```
jsulpis: I think we want to check the bundling here ?

```suggestion
import '@ultraviolet/fonts/fonts-bundled.css'
```

examples/vite/src/main.tsx
```diff
@@ -1,6 +1,7 @@
 import { StrictMode } from 'react'
 import { createRoot } from 'react-dom/client'
 import { App } from './App.tsx'
+import '@ultraviolet/fonts/fonts-bundled.css'
```
philibea, line 4: Image

packages/fonts/package.json
philibea: Yes fixed.

packages/fonts/vite.config.ts
philibea: revert

packages/fonts/vite.config.ts
```diff
@@ -1,15 +1,22 @@
 import path from 'node:path'
+import { defaultConfig } from '@repo/config/vite/vite.config'
 import { defineConfig, mergeConfig } from 'vite'
 
 export const config = mergeConfig(
   {},
   defineConfig({
+    base: './',
     build: {
+      license: true,
```
jsulpis: there is no license, could we remove it ?

chambo-e (approved):

jsulpis (approved):
quick note: 
the comment in `fonts-cdn.css` is for UV devs and the comment in `fonts-bundled.css` is for users. I think they should either be both for UV devs and in that case we can minify the files since there is nothing useful for users, or have the two comments made for users and keep minifying to false. Either option is ok to me

PR #6608, opened 2026-06-22
fix(NumberInputField): improve type safety and null value handling

philibea:
## Summary

**fix(NumberInputField): improve type safety and null value handling**

**What this MR does:**

- Refactors `NumberInputField` to use `ComponentProps` extraction instead of `Omit`, improving type inference
- Adds type constraint to ensure the `name` prop only accepts fields with `number | null` values
- Updates validation functions to properly handle `null` values (in addition to finite numbers)
- Removes unnecessary type casting in the `onChange` handler


packages/form/src/components/NumberInputField/__tests__/index.test-d.tsx
```diff
@@ -0,0 +1,98 @@
+import { useForm } from 'react-hook-form'
```
lisalupi: why create a new file?

packages/form/src/components/NumberInputField/__tests__/index.test-d.tsx
```diff
@@ -0,0 +1,98 @@
...
+      requiredString: string
+      optionalString: string | null
+      undefinedString?: string
+    }
+
+    it('should accept number field names', () => {
+      const { control } = useForm<FormData>()
+
+      // These should all compile without errors
+      ;<NumberInputField name="requiredNumber" control={control} aria-label="Required Number" />
```
lisalupi: renderWithForm(...) ?

lisalupi (approved):

packages/form/src/components/NumberInputField/__tests__/index.test-d.tsx
philibea: renderWithForm doesn't let me play with the control which handle the type i'm trying to test here.

packages/form/src/components/NumberInputField/__tests__/index.test.tsx
```diff
@@ -198,27 +198,145 @@ describe('numberInputField', () => {
...
+      await userEvent.type(input, '1')
+
+      expect(onChange).toHaveBeenCalledExactlyOnceWith(1)
+
+      await userEvent.clear(input)
+
+      expect(onChange).toHaveBeenLastCalledWith(null)
+
+      // Will not trigger onChange event.
       await userEvent.type(input, 'e')
```
lisalupi, line 229: "e" should be an accepted value since we can create number with it (eg typeof 1e10 is number)

packages/form/src/components/NumberInputField/index.tsx
```diff
@@ -3,18 +3,19 @@
...
+import type { FieldPath, FieldValues } from 'react-hook-form'
 import { useErrors } from '../../providers'
 import type { BaseFieldProps } from '../../types'
 
-type NumberInputProps<TFieldValues extends FieldValues, TFieldName extends FieldPath<TFieldValues>> = BaseFieldProps<
-  TFieldValues,
-  TFieldName
-> &
-  Omit<ComponentProps<typeof NumberInput>, 'onChange'>
+type NumberInput = ComponentProps<typeof NumberInput>
```
lisalupi: we have both type NumberInput and component NumberInput. Isn't it a bit confusing?

packages/form/src/types.ts
```diff
@@ -11,11 +11,11 @@ import type {
...
 
+/**
+ * this type is used by the packages/form/src/providers/ErrorContext/index.tsx
+ * should be migrate
+ */
 export type MetaField = {
-  /**
-   * @deprecated this field is not relevant, use the `value` instead to check what you want
-   */
-  isInteger?: number | string
```
lisalupi, line 18: breaking change?

packages/ui/src/components/NumberInput/index.tsx
```diff
@@ -191,6 +191,7 @@ export const NumberInput = forwardRef(
                     onBlur?.(event)
                   }}
                   onChange={
+                    //  onChange is not trigger with a string when type=number, so you can't validate "string" value
```
lisalupi:
```suggestion
                    //  onChange is not triggered with a string when type=number, so you can't validate "string" value
```

utils/test/src/vitest/helpers/index.tsx
```diff
@@ -19,8 +19,8 @@ export const ComponentWrapper = ({
 )
 
 export const mockFormErrors: FormErrors = {
-  isNumber: () => 'This field should be a number',
-  isInteger: () => 'This field should be an integer',
+  isNumber: ({ value }) => `this field should be a number: ${value} `,
```
lisalupi:
```suggestion
  isNumber: ({ value }) => `This field should be a number: ${value} `,
```

packages/form/src/types.ts
philibea: was not used and it's not exposed to our user.

packages/ui/src/components/NumberInput/__stories__/MinMax.stories.tsx
```diff
@@ -4,18 +4,22 @@ import { NumberInput } from '..'
 
 export const Template: StoryFn<typeof NumberInput> = props => {
   const [value, setValue] = useState<number | null>(10)
+  console.debug(props.helper)
```
lisalupi, line 7: to remove

packages/ui/src/components/NumberInput/__stories__/Template.stories.tsx
```diff
@@ -1,4 +1,8 @@
 import type { StoryFn } from '@storybook/react-vite'
 import { NumberInput } from '..'
 
-export const Template: StoryFn<typeof NumberInput> = props => <NumberInput {...props} />
+export const Template: StoryFn<typeof NumberInput> = props => (
+  <>
+    <NumberInput {...props} />
+  </>
+)
```
lisalupi:
```suggestion
    <NumberInput {...props} />
)
```

packages/ui/src/components/NumberInput/components/Controls.tsx
```diff
@@ -2,6 +2,7 @@ import { MinusIcon } from '@ultraviolet/icons/MinusIcon'
 import { PlusIcon } from '@ultraviolet/icons/PlusIcon'
 import { useCallback } from 'react'
 import type { RefObject } from 'react'
+import type { NumberInputProps } from '..'
```
lisalupi: doesn't it create circular dependencies?

packages/ui/src/components/NumberInput/index.tsx
```diff
@@ -182,19 +175,41 @@ export const NumberInput = forwardRef(
...
+                        onChange?.(max)
+                      }
+                      if (!isNullOrUndefined(min) && valueAsNumber < min) {
+                        onChange?.(min)
+                      }
                     }
                     onBlur?.(event)
                   }}
                   onChange={
+                    // onChange is not triggered with a string when type=number, so you can't validate "string" value.
```
lisalupi:
```suggestion
                    // onChange is not triggered with a string, so you can't validate "string" value.
```
since it is always of type number

packages/ui/src/components/NumberInput/components/Controls.tsx
philibea: as it's only types, it's ok, but i create a types.ts file for this.

PR #6615, opened 2026-06-25
chore: add display name

lisalupi:
## Type

- Enhancement
- Documentation

#### What is expected?
Add a `displayName` to every component

This helps to improve code snippets on Storybook


chambo-e (commented):
Are we sure we want to do this ? This adds an overhead we don't really need 

What is the source of the issue ? It was working a couple of days ago, what changed since ?

lisalupi:
> Are we sure we want to do this ? This adds an overhead we don't really need
> 
> What is the source of the issue ? It was working a couple of days ago, what changed since ?

The issue appeared with typescript 7 across all components, but it was already present for some components, specifically the ones with subcomponent(s) : Storybook displays the internal component's name instead of the exported one. For instance, if "InfoTable.Row" uses a compoenent "InfoTableRow", in the snippet storybook show <InfoTableRow> instead of <InfoTable.Row>. This fix mitigates the issue

chambo-e:
@lisalupi got it for the subcomponents

Do we know what plugin/storybook internals does not supports ts v7 ?

lisalupi:
@chambo-e  @storybook/addon-docs is the one generating the snippets

jsulpis:
@chambo-e could you precise the overhead you are worried about ? Runtime performance ? bundle size ? 

The DisplayName could allow to properly see UV components in the React devtools, currently with the forwardRef they are Anonymous

chambo-e:
> @chambo-e could you precise the overhead you are worried about ? Runtime performance ? bundle size ?
> 
> The DisplayName could allow to properly see UV components in the React devtools, currently with the forwardRef they are Anonymous

@jsulpis @lisalupi I'm worried that it's gonna add a dev overhead. We might forget to add it or to update the name of the component, or somehting else.

If the issue is with `@storybook/addon-docs` which breaks with typescript v7 maybe we need to create an issue there ? 

For components behind a forwardRef I agree, I don't think there is another solution 😄 

I'm just raising a warning here, we used to do this (specifying displayName) in earlier uv versions and we've removed it because there were no added value at the time, maybe this has changed :)

jsulpis:
We don't add or rename components very often so IMHO this won't be a huge pain to maintain. And it's trivial to remove. I would say let's merge this to fix the issue of the code snippets, and we will see how it goes

jsulpis (approved):

chambo-e (approved):

PR #6617, opened 2026-06-25
docs: button a11y audit

lisalupi:
## Type

- Documentation

#### What is expected?
Audit `Button`  accessibility

### DIFF
--- a/packages/ui/src/components/Button/__stories__/A11y.mdx
+++ b/packages/ui/src/components/Button/__stories__/A11y.mdx
@@ -1,0 +1,70 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="UI/Action/Button/A11y" />
+
+# Button - Accessibility
+
+## Description
+
+A fundamental interactive component used to trigger actions or events (submitting forms, opening dialogs, canceling actions, etc.). Supports multiple variants (filled, outlined, ghost), sentiments, sizes, and can render as a native button, anchor tag, or custom element via render prop.
+
+## Summary
+
+1. ⚠️ **Perceivable**: Focus indicator implementation unclear, relies on browser default or theme shadows, labelling must be improved, text contrast fails AA in some theme/sentiment combinations
+2. ⚠️ **Operable**: Keyboard support present but focus visibility needs verification
+3. ✅ **Understandable**: Content is readable and labels are supported
+4. ❌ **Robust**: Critical ARIA attributes (`aria-describedby`, `aria-disabled`, `aria-pressed`, `aria-roledescription`, `aria-keyshortcuts`) not passed to button element
+
+## ARIA Pattern
+
+[ARIA Button Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/button/)
+
+## Rules
+
+Enforced by the component:
+
+- ❌ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) - Uses semantic `<button>` and `<a>` elements with proper ARIA attributes support **on anchor only**; button element missing `aria-describedby`, `aria-disabled`, `aria-pressed`, `aria-roledescription`, `aria-keyshortcuts`
+- ⚠️ [WCAG 2.4.7 - Focus Visible (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html) - Focus ring implemented via `boxShadow` on `:active` state only; no explicit `:focus-visible` style defined
+- ⚠️ [WCAG 1.4.11 - Non-text Contrast (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html) - Focus indicator contrast depends on theme shadow values; needs visual verification
+- ⚠️ [WCAG 1.1.1 - Non-text Content (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html) - `aria-label` prop is supported for accessible naming, but correct labelling must be ensured for icon-only buttons and buttons with both icon and text.
+- ❌ [WCAG 1.4.3 - Contrast (Minimum) (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html) - Text color contrast against button background has known failures:
+  - Dark theme + Danger + Filled: `4.29:1`
+  - Darker theme + Success + Filled: `4.47:1`
+  - Darker theme + Danger + Filled: `3.86:1`
+  - Darker theme + Info + Filled: `4.49:1`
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Button/__stories__/A11y.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+## Accessibility issues
+
+**Critical:**
+
+- Missing ARIA attributes on button element - The following ARIA props are defined but NOT passed to the native `<button>` element: `aria-describedby`, `aria-disabled`, `aria-pressed`, `aria-roledescription`, `aria-keyshortcuts`. These work correctly on anchor elements but are silently ignored on buttons. This breaks accessibility for toggle buttons, disabled state communication, keyboard shortcut announcements, and accessible descriptions.
+
+**High:**
+
+- Missing explicit `:focus-visible` styles - Focus ring only appears on `:active` state via `boxShadow`. The component relies on browser default focus outline or theme shadows, which may not be visible in all contexts. Should add explicit `:focus-visible` styles for consistent focus indication across browsers.
```
iManu: This is true, focus should be added

https://github.com/scaleway/ultraviolet/blob/main/packages/ui/src/components/Button/styles.css.ts

packages/ui/src/components/Button/__stories__/A11y.mdx
```diff
@@ -0,0 +1,70 @@
...
+- Text contrast does not meet WCAG 2.1 AA (4.5:1) for certain theme/sentiment combinations:
+  - Dark theme + Danger + Filled: 4.29:1
+  - Darker theme + Success + Filled: 4.47:1
+  - Darker theme + Danger + Filled: 3.86:1
+  - Darker theme + Info + Filled: 4.49:1
+
+**Medium:**
+
+- `aria-disabled` prop available but not coordinated with native `disabled` - When `isLoading` is true, button uses native `disabled` attribute. The `aria-disabled` prop is separate and could create confusion if both are used inconsistently.
```
iManu: you should make a choice, and should prefer native against aria

packages/ui/src/components/Button/__stories__/A11y.mdx
```diff
@@ -0,0 +1,70 @@
...
+- No built-in support for toggle button pattern - While `aria-pressed` is supported, there's no built-in state management for toggle buttons. Developers must manually manage pressed state.
```
iManu: The catch-all aspect of the component create hard to manage situations, like this. For my point of view there are too much aria role and maybe useless in many cases, when the component transforms to anchor versus button)

philibea (approved):

packages/ui/src/components/Button/__stories__/A11y.mdx
```diff
@@ -0,0 +1,86 @@
...
+- [WCAG 1.4.1 - Use of Color (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html): Don't rely on color alone; provide additional text or icons
+- [WCAG 3.2.4 - Consistent Identification (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/consistent-identification.html): Use consistent button labels for same functionality across pages
+- [WCAG 1.1.1 - Non-text Content (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html): Icon-only: use `aria-label` (current) or `tooltipLabel`/`accessibleLabel` (not implemented yet). Icon+text: no extra label needed
+- [WCAG 1.4.11](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html): Ensure icon contrast is >3:1
+
+## Accessibility issues
+
+**Critical:**
+
+- Missing ARIA props on button: The following ARIA props are defined but NOT passed to the native `<button>` element: `aria-describedby`, `aria-disabled`, `aria-pressed`, `aria-roledescription`, `aria-keyshortcuts`. These work correctly on anchor elements but are silently ignored on buttons. This breaks accessibility for toggle buttons, disabled state communication, keyboard shortcut announcements, and accessible descriptions.
```
jsulpis, line 59: could we deprecate aria-pressed, aria-roledescription and aria-disabled ? I see that they have been added for [dnd kit](https://dndkit.com/) which is not installed anymore 

Image

packages/ui/src/components/Button/__stories__/A11y.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+**High:**
+
+- No `:focus-visible` styles. Focus ring only on `:active` via `boxShadow`; relies on browser defaults
+- Contrast failures
+
+**Medium:**
+
+- `aria-disabled` not coordinated with native `disabled`: When `isLoading`, button uses native `disabled`; separate `aria-disabled` may cause confusion
+- No toggle button state management: `aria-pressed` supported but no built-in state handling
```
jsulpis, line 72: Maybe the Button is not supposed to be used as a toggle button so we could deprecate aria-pressed ?

packages/ui/src/components/Button/__stories__/A11y.mdx
```diff
@@ -0,0 +1,86 @@
...
+**Low:**
+
+- Icon contrast (WCAG 1.4.11) must be validated per context (icons use `currentColor`/theme colors)
+- `tooltip` prop adds extra DOM node, may interfere with accessibility patterns
+
+## Dependencies
+
+- Icon component: Must default to `aria-hidden="true"`; currently requires manual prop and renders `<title>` with technical name
+- `VisuallyHidden` component: Required for `accessibleLabel`; doesn't exist yet
+- `.sr-only` CSS class: Required for visually hidden content; doesn't exist yet
```
jsulpis: not necessary if we use the `VisuallyHidden` component ?

jsulpis (approved):

packages/ui/src/components/Button/__stories__/A11y.mdx
lisalupi: It is used as a toggle button in `RichTextInput` (toolbar: bold, italic, etc buttons) I think, but without aria-pressed (or aria- at all) 🤔

packages/ui/src/components/Button/__stories__/A11y.mdx
lisalupi: we never installed it, but some of our users do use it

packages/ui/src/components/Button/__stories__/A11y.mdx
jsulpis: ok so let's say we keep it and fix it later then

packages/ui/src/components/Button/__stories__/A11y.mdx
jsulpis: oh ok... maybe it would be simpler to just forward all native button attributes 🤷

PR #6618, opened 2026-06-26
fix(FileInput): issue with paths

lisalupi:
## Type

- Bug

#### What is expected?
`FileInput`: keep file path when dropping folders

### DIFF
--- a/packages/ui/src/components/FileInput/helpers.ts
+++ b/packages/ui/src/components/FileInput/helpers.ts
@@ -61,10 +61,25 @@ const isDirectoryEntry = (entry: FileSystemEntry): entry is FileSystemDirectoryE
 
 const convertEntryToFile = (entry: FileSystemFileEntry): Promise<File> =>
   new Promise((resolve, reject) => {
-    entry.file(resolve, reject)
+    entry.file(file => {
+      const cleanPath = entry.fullPath?.startsWith('/') ? entry.fullPath.slice(1) : entry.fullPath
+
+      // If cleanPath === file.name => the file was NOT dropped from a folder, so we don't need to correct the webkitRelativePath
+      // If cleanPath === file.webkitRelativePath => no issue in webkitRelativePath so no need to correct it
+      if (cleanPath !== file.webkitRelativePath && cleanPath !== file.name) {
+        Object.defineProperty(file, 'webkitRelativePath', {
+          value: cleanPath,
+          writable: true,
+          enumerable: true,
+          configurable: true,
+        })
+      }
+
+      resolve(file)
+    }, reject)
   })
 
-async function readEntriesInDirectory(directory: FileSystemDirectoryEntry) {
+const readEntriesInDirectory = async (directory: FileSystemDirectoryEntry) => {
   const reader = directory.createReader()
   const nestedEntries = []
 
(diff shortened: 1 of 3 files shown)


PierreBertinet (approved):

packages/ui/src/components/FileInput/helpers.ts
```diff
@@ -1,5 +1,9 @@
 import type { MimeType } from './types'
 
+type FileWithPath = {
+  webkitRelativePath: string
+} & File
+
```
philibea: File already give this type

 */
interface File extends Blob {
    /**
     * The **`lastModified`** read-only property of the File interface provides the last modified date of the file as the number of milliseconds since the Unix epoch (January 1, 1970 at midnight). Files without a known last modified date return the current date.
     *
     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/File/lastModified)
     */
    readonly lastModified: number;
    /**
     * The **`name`** read-only property of the File interface returns the name of the file represented by a File object. For security reasons, the path is excluded from this property.
     *
     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/File/name)
     */
    readonly name: string;
    /**
     * The **`webkitRelativePath`** read-only property of the File interface contains a string which specifies the file's path relative to the directory selected by the user in an <input> element with its webkitdirectory attribute set.
     *
     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/File/webkitRelativePath)
     */
    readonly webkitRelativePath: string;
}

packages/ui/src/components/FileInput/helpers.ts
```diff
@@ -59,12 +63,22 @@ export const fileIsAccepted = (file: File, accept?: string) => {
...
 const isDirectoryEntry = (entry: FileSystemEntry): entry is FileSystemDirectoryEntry => !!entry.isDirectory
 
-const convertEntryToFile = (entry: FileSystemFileEntry): Promise<File> =>
+// Manually add webkitRelativePath because sometimes it is not added to the File object (Chrome)
+const convertEntryToFile = (entry: FileSystemFileEntry): Promise<FileWithPath> =>
   new Promise((resolve, reject) => {
-    entry.file(resolve, reject)
+    entry.file(file => {
+      const cleanPath = entry.fullPath?.startsWith('/') ? entry.fullPath.slice(1) : entry.fullPath
+      Object.defineProperty(file, 'webkitRelativePath', {
```
philibea: we should check if the webkitRelativePath is already set as it's the normal behavior ?

packages/ui/src/components/FileInput/helpers.ts
lisalupi: i know but when i convert to a File (entry.file(resolve, reject)) chrome somehow drops it

packages/ui/src/components/FileInput/helpers.ts
lisalupi: I can add a check to use webkitRelativePath if the `""` value chrome always returns is correct. 
It should be `""` only when the file is not inside a dir

lisalupi:
```js
// If cleanPath === file.name => the file was not droped from a folder, so we don't need to correct the webkitRelativePath
// If cleanPath === file.webkitRelativePath => no issue in webkitRelativePath so no need to correct it

if (cleanPath !== file.webkitRelativePath && cleanPath !== file.name) {
```

what do you think?

packages/ui/src/components/FileInput/helpers.ts
lisalupi: I can remove it my bad, it is defined, just always set to `""`

packages/ui/src/components/FileInput/helpers.ts
```diff
@@ -1,5 +1,9 @@
 import type { MimeType } from './types'
 
+type FileWithPath = {
+  webkitRelativePath: string
```
philibea: can be remove

philibea (approved):

PR #6626, opened 2026-06-29
fix(OrderSummary): prices overflow

lisalupi:
## Type

- Enhancement

#### What is expected?

`OrderSummary`: fix overflow with long (sub-)category name and prices

### DIFF
--- a/packages/ui/src/compositions/OrderSummary/styles.css.ts
+++ b/packages/ui/src/compositions/OrderSummary/styles.css.ts
@@ -122,14 +122,19 @@ const totalPrice = styleVariants({
     display: 'flex',
     alignItems: 'center',
     gap: theme.space[1],
+    minWidth: 'fit-content',
   },
-  default: {},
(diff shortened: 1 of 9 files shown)


lisalupi:
waiting for design for more details on expected behavior

packages/ui/src/compositions/OrderSummary/styles.css.ts
```diff
@@ -132,6 +133,10 @@ const compactTotalPrice = style({
   gap: theme.space[1],
 })
 
+const categoryPrice = style({
+  minWidth: 'fit-content',
```
jsulpis, line 137: quick note: it also works and is simpler IMHO with a `display: grid` on the parent with a `grid-template-columns: 1fr auto;` but maybe it would break something else so that's no big deal (just FYI)

jsulpis (approved):

philibea (approved):

PR #6631, opened 2026-06-30
fix(SelectInput): searchable behavior

lisalupi:
## Type

- Enhancement

#### What is expected?
`SelectInput`: deprecate `searchable` prop; searchable behavior is completely automatic

### DIFF
--- a/packages/ui/src/components/SelectInput/__tests__/index.test.tsx
+++ b/packages/ui/src/components/SelectInput/__tests__/index.test.tsx
@@ -467,7 +467,7 @@ describe('selectInput', () => {
     expect(asFragment()).toMatchSnapshot()
   })
 
-  it('should be searchable when there are less more than 6 items', async () => {
+  it('should be searchable when there are more than 6 items', async () => {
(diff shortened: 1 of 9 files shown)


Echo-IV (approved):

packages/ui/src/components/SelectInput/__tests__/index.test.tsx
```diff
@@ -429,25 +424,65 @@ describe('selectInput', () => {
...
     await userEvent.click(input)
     const dropdown = screen.getByRole('dialog')
     await waitFor(() => {
       expect(dropdown).toBeVisible()
+      expect(screen.getByRole('textbox', { name: 'search-bar' })).toBeInTheDocument()
+    })
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  it('should be searchable when there are less more than 6 items', async () => {
```
jsulpis:
```suggestion
  it('should be searchable when there is more than 6 items', async () => {
```

jsulpis (approved):

PR #6633, opened 2026-07-01
feat: deprecate themes in ui

lisalupi:
## Type

Deprecate `ThemeProvider`, `darkTheme`, `darkerTheme`, `theme`, and `useTheme` from @ultraviolet/ui. Use the equivalents from @ultraviolet/themes instead.

⚠️⚠️ **Future breaking change** ⚠️⚠️  
The `ThemeProvider` from `@ultraviolet/ui` automatically injects styles from `@ultraviolet/icons`.  
The replacement `ThemeProvider` in `@ultraviolet/themes` does **not** include this side effect.

When these exports are removed in the next major version, you must manually import the icon styles to avoid visual regressions:

```js
import '@ultraviolet/icons/styles'
```

### DIFF
--- a/examples/next/src/pages/advanced/AdvancedUseCases.tsx
+++ b/examples/next/src/pages/advanced/AdvancedUseCases.tsx
@@ -2,8 +2,8 @@ import { Stack, Text } from '@ultraviolet/ui'
 import CopyBox from '../../components/CopyBoxCommand'
 import styles from '../../../styles/advanced.module.scss'
 
-const useCase1 = `import { theme as lightTheme, dark as darkTheme, Button, Text } from '@ultraviolet/ui'
-import { ThemeProvider } from '@ultraviolet/themes'
+const useCase1 = `import { theme as lightTheme, Button, Text } from '@ultraviolet/ui'
+import { ThemeProvider, consoleDarkTheme as darkTheme } from '@ultraviolet/themes'
 import React, { useCallback, useState } from 'react'
 
 const App = () => {
--- a/examples/next/src/pages/advanced/AdvancedUseCases.tsx
+++ b/examples/next/src/pages/advanced/AdvancedUseCases.tsx
@@ -2,8 +2,8 @@ import { Stack, Text } from '@ultraviolet/ui'
 import CopyBox from '../../components/CopyBoxCommand'
 import styles from '../../../styles/advanced.module.scss'
 
-const useCase1 = `import { theme as lightTheme, Button, Text } from '@ultraviolet/ui'
-import { ThemeProvider, consoleDarkTheme as darkTheme } from '@ultraviolet/themes'
+const useCase1 = `import { Button, Text } from '@ultraviolet/ui'
+import { ThemeProvider, consoleDarkTheme as darkTheme, consoleLightTheme as lightTheme } from '@ultraviolet/themes'
 import React, { useCallback, useState } from 'react'
 
 const App = () => {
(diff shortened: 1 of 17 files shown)


packages/themes/src/extendTheme.ts
```diff
@@ -0,0 +1,15 @@
...
+
+type RecursivePartial<T> = {
+  [P in keyof T]?: RecursivePartial<T[P]>
+}
+
+/**
+ * Will extend theme with new theme properties
+ * @param {UltravioletUITheme} baseTheme the theme you want to extend from, by default it is set to light theme
+ * @param {RecursivePartial<UltravioletUITheme>} extendedTheme the properties of a new theme you want to apply from baseTheme
+ */
```
jsulpis: there is only one param to the function

```suggestion
 * @param {RecursivePartial<UltravioletUITheme>} extendedTheme the properties of a new theme you want to merge to the default light theme
 */
```

jsulpis (commented):
I see `extendTheme` a few times in stories, imported from `@ultraviolet/ui`, it should be updated to use `@ultraviolet/themes`

packages/themes/src/extendTheme.ts
```diff
@@ -0,0 +1,14 @@
...
+type RecursivePartial<T> = {
+  [P in keyof T]?: RecursivePartial<T[P]>
+}
+
+/**
+ * Will extend theme with new theme properties
+ * @param {RecursivePartial<UltravioletUITheme>} extendedTheme the properties of a new theme you want to apply from baseTheme
+ */
+export const extendTheme = (extendedTheme: RecursivePartial<UltravioletUITheme>) =>
+  deepmerge(consoleLightTheme, extendedTheme) as UltravioletUITheme
```
philibea:
```suggestion
export const extendTheme = (extendedTheme: RecursivePartial<UltravioletUITheme>): UltravioletUITheme =>
  deepmerge(consoleLightTheme, extendedTheme)
```

packages/ui/src/theme/ThemeProvider.tsx
```diff
@@ -32,6 +32,7 @@ type ThemeProviderProps = {
 /**
  * ThemeProvider will apply generated global CSS variables to the application in the `<head>`.
  * If no theme is provided, it will default to `lightTheme`.
+ * @deprecated use ThemeProvider from \@ultraviolet/themes
```
philibea: you can add the changeset deprecated explanation here

philibea (changes requested):

examples/next/src/pages/advanced/AdvancedUseCases.tsx
```diff
@@ -2,8 +2,8 @@ import { Stack, Text } from '@ultraviolet/ui'
 import CopyBox from '../../components/CopyBoxCommand'
 import styles from '../../../styles/advanced.module.scss'
 
-const useCase1 = `import { theme as lightTheme, dark as darkTheme, Button, Text } from '@ultraviolet/ui'
-import { ThemeProvider } from '@ultraviolet/themes'
+const useCase1 = `import { theme as lightTheme, Button, Text } from '@ultraviolet/ui'
+import { ThemeProvider, consoleDarkTheme as darkTheme } from '@ultraviolet/themes'
```
jsulpis: I think ?
```suggestion
const useCase1 = `import { Button, Text } from '@ultraviolet/ui'
import { ThemeProvider, consoleLightTheme as lightTheme, consoleDarkTheme as darkTheme } from '@ultraviolet/themes'
```

jsulpis (approved):

packages/themes/src/extendTheme.ts
lisalupi: it creates a type error

philibea (approved):

packages/themes/src/extendTheme.ts
philibea: deepmerge should keep the same type normally so what is the type error ?

lisalupi: ok my bad I retried and I don't have any error, I probably made a typo the first type

PR #6635, opened 2026-07-02
fix(NumberInput): add background

lisalupi:
## Type

- Bug

#### What is expected?
`NumberInput`, `UnitInput`: add background

| UnitInput  | Capture d’écran 2026-07-02 à 14 12 41| Capture d’écran 2026-07-02 à 14 12 56|
| NumberInput  | Capture d’écran 2026-07-02 à 14 13 38| Capture d’écran 2026-07-02 à 14 13 22|

### DIFF
--- a/packages/ui/src/components/NumberInput/styles.css.ts
+++ b/packages/ui/src/components/NumberInput/styles.css.ts
@@ -95,6 +95,7 @@ const unit = recipe({
 
 const numberinput = recipe({
   base: {
+    background: 'none',
     borderWidth: 1,
     borderStyle: 'solid',
     borderColor: 'inherit',
@@ -205,8 +206,11 @@ const container = recipe({
       '&:has(input:out-of-range), &:has(:user-invalid:focus)': {
         boxShadow: theme.shadows.focusDanger,
       },
-      '&:has(input:out-of-range), &:has(:user-invalid)': {
+      '&:has(input:out-of-range:hover), &:has(:user-invalid:hover)': {
         borderColor: theme.colors.danger.borderHover,
+      },
+      '&:has(input:out-of-range), &:has(:user-invalid)': {
+        borderColor: theme.colors.danger.border,
       },
       '&[data-controls="false"]': {
         borderWidth: 0,
--- a/packages/ui/src/components/NumberInput/styles.css.ts
+++ b/packages/ui/src/components/NumberInput/styles.css.ts
@@ -39,7 +39,14 @@ const inputContainer = recipe({
 })
 
 const unit = recipe({
-  base: { alignItems: 'center', display: 'flex', padding: theme.space[1] },
+  base: {
+    alignItems: 'center',
+    display: 'flex',
+    padding: theme.space[1],
+    borderBlockWidth: 1,
+    borderColor: 'inherit',
+    borderBlockStyle: 'solid',
+  },
   defaultVariants: {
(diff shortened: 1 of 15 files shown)


packages/ui/src/components/NumberInput/styles.css.ts
```diff
@@ -187,10 +202,15 @@ const container = recipe({
       '&:hover': {
         border: `1px solid ${theme.colors.primary.borderHover}`,
       },
-      '&:has(input:out-of-range), &:has(input:invalid), &:has(:user-invalid)': {
-        borderColor: theme.colors.danger.borderHover,
+      '&:has(input:out-of-range), &:has(:user-invalid:focus)': {
         boxShadow: theme.shadows.focusDanger,
       },
```
philibea, line 208:
```suggestion
      '&:has(input:out-of-range), &:has(:user-invalid:focus)': {
       borderColor: theme.colors.danger.borderHover
       boxShadow: theme.shadows.focusDanger,
},
```

lisalupi: it is already done on the line above

philibea (approved):

packages/ui/src/components/NumberInput/styles.css.ts
```diff
@@ -79,15 +86,18 @@ const unit = recipe({
...
     },
   },
 })
 
 const numberinput = recipe({
   base: {
-    background: 'none',
-    border: 'none',
+    borderWidth: 1,
```
jsulpis, line 99: what's wrong with the border on the container like before ? It's a bit weird to have the border on all child elements instead of the container

packages/ui/src/components/NumberInput/styles.css.ts
```diff
@@ -179,6 +193,7 @@ const container = recipe({
     display: 'flex',
     flexDirection: 'row',
     justifyContent: 'space-between',
+    background: theme.colors.neutral.background,
```
jsulpis, line 197: not sure it's this line but there is an issue with the token in dark mode 

Image

packages/ui/src/components/NumberInput/styles.css.ts
```diff
@@ -187,10 +202,15 @@ const container = recipe({
       '&:hover': {
         border: `1px solid ${theme.colors.primary.borderHover}`,
       },
-      '&:has(input:out-of-range), &:has(input:invalid), &:has(:user-invalid)': {
-        borderColor: theme.colors.danger.borderHover,
+      '&:has(input:out-of-range), &:has(:user-invalid:focus)': {
         boxShadow: theme.shadows.focusDanger,
       },
+      '&:has(input:out-of-range), &:has(:user-invalid)': {
+        borderColor: theme.colors.danger.borderHover,
```
jsulpis, line 210: I think there is a missing `:hover` state to use with this `borderHover` token (so maybe need to add the non-hover state as well)

packages/ui/src/components/NumberInput/styles.css.ts
lisalupi: fixed

packages/ui/src/components/NumberInput/styles.css.ts
lisalupi: fixed

jsulpis (approved):

PR #6638, opened 2026-07-06
refactor: create useField hook for fields components

jsulpis:
## Type

- Refactor

### Summarize concisely:

Extract a `useField` hook to centralize Field components logic

#### What is expected?

- no change in behavior (form submission, error handling etc)
- less re-renders
- centralized fields logic

- create a `useField` hook by extracting logic from the `TextInputField`
- use it in `TextInputField` and `SelectInputField`
- create a form in the example Vite app to test the usage and the re-renders

### DIFF
--- a/examples/vite/src/DemoForm.tsx
+++ b/examples/vite/src/DemoForm.tsx
@@ -12,9 +12,11 @@ type FormValues = {
 
 export const DemoForm = () => {
   const methods = useForm<FormValues>({
+    mode: 'onChange',
     defaultValues: {
-      // number: 5,
-      // text: 'test',
+      number: null,
+      name: 'Doe',
+      firstname: 'John',
     },
   })
 
@@ -30,6 +32,8 @@ export const DemoForm = () => {
         pattern: () => 'pattern',
         min: () => 'min',
         minLength: ({ minLength }) => `min ${minLength} characters`,
+        isInteger: () => 'should be an integer',
+        isNumber: () => 'should be a number',
       }}
       methods={methods}
       onSubmit={handleSubmit}
@@ -38,7 +42,13 @@ export const DemoForm = () => {
       <Stack gap="3" maxWidth="400px">
         <TextInputField<FormValues> name="name" minLength={3} label="Name" placeholder="Doe" required />
         <TextInputField<FormValues> name="firstname" minLength={3} label="FirstName" placeholder="John" required />
-        <NumberInputField<FormValues> label="Number Input" name="number" min={10} placeholder="Enter a number" />
+        <NumberInputField<FormValues>
+          label="Number Input"
+          name="number"
+          min={10}
+          placeholder="Enter a number"
+          step="1"
+        />
         <DateInputField<FormValues> label="Date Input" name="date" placeholder="YYYY-MM-DD" />
         <Button type="submit">Submit</Button>
       </Stack>
--- a/examples/vite/src/DemoForm.tsx
+++ b/examples/vite/src/DemoForm.tsx
@@ -10,13 +10,23 @@ type FormValues = {
   date: Date | null
 }
 
+const mockFormErrors: import('@ultraviolet/form').FormErrors = {
+  required: () => 'field required',
+  pattern: () => 'pattern',
+  min: () => 'min',
+  max: () => 'max',
+  minLength: ({ minLength }) => `min ${minLength} characters`,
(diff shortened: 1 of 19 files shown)


packages/form/src/components/TextInputField/index.tsx
```diff
@@ -23,73 +21,26 @@ export const TextInputField = <
...
-      label={label}
-      maxLength={maxLength}
-      minLength={minLength}
-      name={name}
+      {...fieldProps}
+      error={error}
       onBlur={event => {
         onBlur?.(event)
-        field.onBlur()
+        fieldProps.onBlur(event)
```
jsulpis: question: should it be called `fieldProps` or `field` like before ?

packages/form/src/components/Submit/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -32,7 +32,6 @@ exports[`submit > form is invalid 1`] = `
             id="_r_1_"
             name="toto"
             type="text"
-            value="4"
```
philibea, line 35: question: just wonder if ours users have some tests based on this value this will be a breaking change for them ?

NB: this is maybe juste an issue with this test where the default value was spread to the TextInput before

jsulpis: IMO the exact template is an implementation detail so I don't know what is the status quo but I would not encourage users to do snapshot tests (so I would not consider this a breaking change, but it's debatable)

packages/form/src/components/TextInputField/index.tsx
philibea: fieldProps sounds better

packages/form/src/hooks/useRegisterField.tsx
```diff
@@ -0,0 +1,82 @@
...
+      },
+      callback: ({ values }) => {
+        const error = getError({ ...errorMeta, value: values[name] }, getFieldState(name).error)
+        setError(error)
+      },
+    })
+    return unsubscribe
+  }, [subscribe, getFieldState, name, errorMeta, getError])
+
+  const registeredField = register(name, {
```
philibea, line 60: should we add disabled  props ?

philibea: https://react-hook-form.com/docs/useform/register

> Set disabled to true will lead input value to be undefined and input control to be disabled.
> disabled prop will also omit built-in validation rules.
> For schema validation, you can leverage the undefined value returned from input or context object.

packages/form/src/components/TextInputField/useField.tsx
```diff
@@ -0,0 +1,82 @@
+import { useState, useEffect, useMemo } from 'react'
+import { FieldValues, FieldPath, useFormContext } from 'react-hook-form'
+import { useErrors } from '../../providers'
+import { BaseFieldProps, MetaField } from '../../types'
+import { validateRegex } from '../../utils/validateRegex'
+
+/**
+ * Hook to use in Field components to centralize the logic linked to React-Hook-Form and other common field props.
+ */
+export const useField = <
```
philibea: should be place into ours hooks directory ?

packages/form/src/hooks/useRegisterField.tsx
jsulpis: Is it related to the changes in this PR ? If not, I would rather consider it after this one is merged

packages/form/src/components/TextInputField/useField.tsx
jsulpis: it has been moved in `packages/form/src/hooks`

examples/vite/src/DemoForm.tsx
```diff
@@ -0,0 +1,57 @@
...
+    },
+  })
+
+  const handleSubmit = (data: FormValues) => {
+    console.log('Form submitted:', data)
+  }
+
+  return (
+    <Form
+      // @ts-expect-error we don't need all errors here
```
philibea: nitpick(non-blocking):
can we avoid the ts-expect-error, even if it's just a test, not a fan to hide error

jsulpis: ✅

lisalupi (approved):

philibea (approved):

PR #6639, opened 2026-07-06
feat(KeyValueInput): new component

lisalupi:
## Type

- Feature
- Enhancement

1. New component `KeyValueInput`, inspired by `KeyValueField`
2. Use `KeyValueInput` to define `KeyValueField` (like other form components) instead of having a custom stand-alone component
3. Improvements: prop size supported, can have a `select` variant (instead of always being text)


Echo-IV (approved):

packages/ui/src/components/KeyValueInput/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+import { SelectInput } from '../SelectInput'
+import { Stack } from '../Stack'
+import { Text } from '../Text'
+import { TextInput } from '../TextInput'
+import { KeyValueInputProps } from './types'
+
+/**
+ * KeyValuenput allow user to add key-value pairs
+ */
+export const KeyValueInput = ({
```
philibea, line 18: Should we have a KeyValueInputField also ?

lisalupi: We already have `KeyValueField` and I updated it in this PR to use KeyValueInput

packages/ui/src/components/KeyValueInput/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,23 @@
+import type { Meta } from '@storybook/react-vite'
+import { KeyValueInput } from '..'
+
+export default {
+  component: KeyValueInput,
+  decorators: [StoryComponent => <StoryComponent />],
```
jsulpis: could this be deleted ?

packages/ui/src/components/KeyValueInput/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+  className,
+  size = 'large',
+  disabled,
+  required,
+  error,
+  onChange,
+  onFocus,
+  onBlur,
+  name,
+  'data-testid': dataTestid,
```
jsulpis: following the discussion on Slack, is it ok for you to avoid adding the `data-testid` ?

packages/ui/src/components/KeyValueInput/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+  return (
+    <Stack gap={3} style={style} className={className} data-testid={dataTestid}>
+      {keyvalues && keyvalues?.length > 0 ? (
+        <Stack gap={3}>
+          {keyvalues.map((_, index) => (
+            <Row alignItems="flex-end" gap={2} key={index} templateColumns="1fr 1fr auto">
+              {inputKey.inputType === 'select' ? (
+                <SelectInput
+                  options={inputKey.options}
+                  label={inputKey.label}
```
jsulpis: would it be possible to mutualize common props in an (or two) object to avoid all the duplication in Select/Text + key/value ?

packages/ui/src/components/KeyValueInput/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+                  name={`${name}.${index}.key`}
+                  readOnly={readOnly}
+                  required={inputKey.required || required}
```
jsulpis: is it on purpose that if the global `required` prop is `true`, then it's not possible to have a key or value not required ? This would be possible with `??` instead of `||`, but either option can make sense

packages/ui/src/components/KeyValueInput/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+                  size={size}
+                  onChangeValue={key => handleChange(index, 'change', key)}
+                  disabled={disabled}
+                  data-testid={`${dataTestid}-${index}.key`}
+                  aria-describedby={ariaDescribedBy}
+                  onFocus={() => onFocus?.(keyvalues, index)}
+                  onBlur={() => onBlur?.(keyvalues, index)}
+                  value={keyvalues[index].key}
```
jsulpis: the `regex` properties are missing on the two TextInputs, which means the regex tests are also missing

packages/ui/src/components/KeyValueInput/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+          sentiment="primary"
+          tooltip={canAdd ? addButton.tooltip : maxSizeReachedTooltip}
+          variant="outlined"
+          size={size}
+        >
+          <PlusIcon disabled={!editable} />
+          {addButton.name}
+        </Button>
+      </Stack>
+      {typeof error === 'string' ? (
```
jsulpis: - I think we should use the `Description` component for consistency, with an id used in the `aria-describedby` attributes of the fields like we do in the other components
- could you add a story that displays the error, or add the error param in the storybook controls so that we can check the display ?

packages/ui/src/components/KeyValueInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,337 @@
...
+      />,
+    )
+    expect(screen.getByRole('button', { name: /add/i })).toBeDisabled()
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  it('renders readOnly', () => {
+    const { asFragment } = renderWithTheme(
+      <KeyValueInput
+        disabled
```
jsulpis:
```suggestion
        readOnly
```

packages/ui/src/components/KeyValueInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,337 @@
...
+        inputValue={{
+          label: 'value',
+        }}
+        size={size}
+      />,
+    )
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  it('renders readOnly', () => {
```
jsulpis: there is the same test above

packages/ui/src/components/KeyValueInput/__tests__/index.test.tsx
```diff
@@ -0,0 +1,337 @@
...
+    const keyInput = screen.getByRole('textbox', { name: 'key' })
+    const valueInput = screen.getByRole('textbox', { name: 'value' })
+    expect(keyInput).toBeInTheDocument()
+
+    await userEvent.type(keyInput, '1')
+    expect(onChange).toHaveBeenCalledExactlyOnceWith([{ key: '1', value: '' }], 0, 'change')
+
+    await userEvent.type(valueInput, '2')
+    expect(onChange).toHaveBeenCalledTimes(2)
+    expect(onChange).toHaveBeenCalledWith([{ key: '', value: '2' }], 0, 'change')
```
jsulpis, line 161: why is it not `{ key: '1', value: '2' }` ? 🤔

packages/ui/src/components/KeyValueInput/index.tsx
lisalupi: no problem

packages/ui/src/components/KeyValueInput/index.tsx
lisalupi: I tried to mutualize the component (`MyComponent = select ? SelectInput : TextInput` something like that) but I had type errors (typescript struggled to get the right types), so I went back to duplicate the code. Did not think of mutualizing the component haha, I'll try that

packages/ui/src/components/KeyValueInput/index.tsx
lisalupi: yes i was on purpose, I thought that it made more sense to enforce it on every single input when the whole "field" is required. But I am not 100% set on this choice, we can change that

packages/ui/src/components/KeyValueInput/__tests__/index.test.tsx
lisalupi: i did not bother to control the component in the test so the changes are not applied to it

packages/ui/src/components/KeyValueInput/index.tsx
lisalupi: regex doesn't exist on `TextInput`, only `TextInputField`

packages/ui/src/components/KeyValueInput/index.tsx
jsulpis: yes ok I think it's better to be too strict and remove the constraint later if needed, than the other way around

packages/ui/src/components/KeyValueInput/index.tsx
jsulpis: Oh ok but the regex is now broken on the `KeyValueField` (see the [regex story](https://ultravioletjxvhzygy-featkeyvaluefieldstorybook.functions.fnc.fr-par.scw.cloud/?path=/story/form-components-fields-keyvaluefield--regex), so the missing test is on the `KeyValueField`). I don't know how the fieldArray works, could we give it the regex to restore the previous behavior ?

packages/ui/src/components/KeyValueInput/__tests__/index.test.tsx
jsulpis: could you delete it if it's the same as above ?

packages/ui/src/components/KeyValueInput/index.tsx
lisalupi: should be fixed now

packages/form/src/components/KeyValueField/__tests__/index.test.tsx
```diff
@@ -73,4 +73,41 @@ describe('keyValueField', () => {
...
+    const input = screen.getByRole('textbox', { name: 'value' })
+    expect(input).toBeInTheDocument()
+
+    await userEvent.type(input, '2')
+    expect(input).toHaveAccessibleDescription(/This field should match the regex/i)
+    expect(input).toBeInvalid()
+
+    await userEvent.clear(input)
+    await userEvent.type(input, 'a')
+    screen.debug()
```
jsulpis: 🗑️ 
```suggestion
```

packages/ui/src/components/KeyValueInput/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+            const errorKey = fieldErrors?.[index]?.key
+            const errorValue = fieldErrors?.[index]?.value
+
+            return (
+              <Row alignItems="flex-end" gap={2} key={index} templateColumns="1fr 1fr auto">
+                {inputKey.inputType === 'select' ? (
+                  <SelectInput
+                    options={inputKey.options}
+                    onChange={(key: string) => handleChange(index, 'change', key)} // can't properly infer type of onChange
+                    {...commonProps(index, 'key', !!errorKey)}
```
jsulpis: you could pass the error string directly and in the function `commonProps`, use it instead of `inputError` and `hasError`

```suggestion
                    {...commonProps(index, 'key', errorKey)}
```

packages/form/src/components/KeyValueField/index.tsx
```diff
@@ -1,147 +1,115 @@
...
   }
 
-  const handleFieldBlur = () => {
-    onBlur?.(getValues(name))
-  }
+  const currentValues = (getValues(name) ?? []) as KeyValueType
 
-  const canAdd = fields.length !== undefined && fields.length < maxSize
-  const maxSizeReachedTooltip = addButton.maxSizeReachedTooltip ?? `Cannot add more than ${maxSize} elements`
+  const fieldErrors = fields.reduce<Record<number, KeyValueFieldError>>((acc, _, index) => {
```
jsulpis: could this be an array instead of an object with indexes as properties ? I think it would be easier to work with

packages/form/src/components/KeyValueField/index.tsx
```diff
@@ -1,147 +1,115 @@
...
+}: KeyValueFieldProps<TFieldValues, TFieldArrayName, TFieldName>) => {
+  const regexKey = inputKey.inputType !== 'select' ? inputKey.regex : undefined
+  const regexValue = inputValue.inputType !== 'select' ? inputValue.regex : undefined
+  const { errors } = useErrors()
+  const {
+    fieldState: { error },
+  } = useController<TFieldValues, TFieldName>({
+    control,
+    name,
+    rules: {
```
jsulpis, line 47: If I understand correctly [the doc](https://react-hook-form.com/docs/usefieldarray) (and Gemini), I think you can move this `rules` object in the `useFieldArray`. And then you can read the "global" error with `formState.errors?.[name]?.root `. This would avoid the extra `useController` used just for the error

packages/form/src/components/KeyValueField/index.tsx
```diff
@@ -1,147 +1,115 @@
...
+      validate:
```
jsulpis, line 48: the `disabled` prop could be added in the `validate` object so that if the whole KeyValue field is required, the global error is shown and the field is invalid

Image

packages/ui/src/components/KeyValueInput/styles.css.ts
```diff
@@ -0,0 +1,8 @@
+import { theme } from '@ultraviolet/themes'
+import { styleVariants } from '@vanilla-extract/css'
+
+const errorDescription = styleVariants({
+  key: { gridColumn: '1', marginTop: `calc(-1 * ${theme.space[1]})` },
```
jsulpis: you could probably split the `gap` on the parent grid to `column-gap` + smaller `row-gap` to avoid the negative margin patch here

packages/form/src/components/KeyValueField/__tests__/index.test.tsx
lisalupi: oupsi

packages/form/src/components/KeyValueField/index.tsx
```diff
@@ -1,147 +1,119 @@
...
-                onChange={handleFieldChange}
-                placeholder={inputValue.placeholder}
-                readOnly={readOnly}
-                regex={inputValue.regex}
-                required={inputValue.required}
-                type={inputValue.type}
-              />
+    if (keyError || valueError) {
+      return {
+        key: keyError ? (errorLabel ?? errors.pattern({ label: label ?? '', regex: regexKey })) : '',
```
jsulpis, line 96: the `fieldErrors` contain only the patterns error, so in the `Regex` story of the `KeyValueInputField` when I add a pair, type in the key and erase it there is no "required" error. I think the `fieldErrors` should return required errors too

Image

packages/form/src/components/KeyValueField/index.tsx
```diff
@@ -1,147 +1,119 @@
...
+                if (!keyValueArray) return !required
+
+                return keyValueArray.every(item => {
+                  const keyValid =
+                    (required || inputKey.required) && item?.key && (!regexKey || validateRegex(item?.key, regexKey))
+                  const valueValid =
+                    (required || inputValue.required) &&
+                    item?.value &&
+                    (!regexValue || validateRegex(item?.value, regexValue))
+                  return keyValid && valueValid
```
jsulpis, line 64: little bug here, with `(required || inputKey.required) &&`, the key/value must be required to be valid. Which I can verify by removing the "required" key in the regex story, the field becomes invalid.

Image

this seems to fix it:
```ts
const keyValidRegex = !regexKey || validateRegex(item?.key, regexKey)
const keyValidRequired = !(required || inputKey.required) || item?.key
const keyValid = keyValidRegex && keyValidRequired

const valueValidRegex = !regexValue || validateRegex(item?.value, regexValue)
const valueValidRequired = !(required || inputValue.required) || item?.value
const valueValid = valueValidRegex && valueValidRequired

return keyValid && valueValid
```

you can add a test to check that with nothing required and a regex on the key, when typing on the key the field is valid (submit button enabled, no need for another snapshot)

jsulpis (commented):
almost there 💪

packages/form/src/components/KeyValueField/index.tsx
jsulpis: edit: this was the previous behavior, we should fix it in another PR

jsulpis (approved):
well done 👏 👏

PR #6664, opened 2026-07-09
fix(Text): prominence style

lisalupi:
## Type

- Bug

#### What is expected?

- Fix `Key`style when prominence="strong"
- Fix `Text` style when prominence !== "default"
- Doc: add a "ShowCase" story (similar to the one for `Button`) to showcase all possible (non-disabled) variants of `Text`

| Text  |Capture d’écran 2026-07-09 à 17 32 34 |Capture d’écran 2026-07-09 à 17 32 00 |
| Key  |Capture d’écran 2026-07-09 à 17 33 13 |Capture d’écran 2026-07-09 à 17 33 11 |


packages/ui/src/components/Text/__stories__/Showcase.stories.tsx
```diff
@@ -0,0 +1,57 @@
+import type { StoryFn } from '@storybook/react-vite'
+import { HelpCircleIcon } from '@ultraviolet/icons'
+import { cn } from '@ultraviolet/utils'
+import { Text } from '..'
+import { Stack, Table, Tooltip } from '../..'
+import { SENTIMENTS } from '../../../theme'
+import { showCase } from './style.css'
+
+const textProminences = ['weak', 'default', 'strong', 'stronger'] as const
```
jsulpis: would that work ? To keep a single definition of the prominences

```suggestion
import { PROMINENCES } from '../../../theme'
...
const textProminences = Object.keys(PROMINENCES)
```

packages/ui/src/components/Text/style.css.ts
jsulpis:
```suggestion
   * "weak" and "stronger" prominences are only defined inside neutral sentiment.
```

jsulpis (approved):

packages/ui/src/components/Text/style.css.ts
jsulpis: I think this is equivalent
```suggestion
const definedProminence = isStrong(prominence) ? 'Strong' : ''
    const text = `text${definedProminence}${disabled ? 'Disabled' : ''}` as const
```

jsulpis: the diff doesn't work but it's a replacement for lines 71-78

philibea (approved):

packages/ui/src/components/Text/__stories__/Showcase.stories.tsx
lisalupi: it is typed as a `string[]`, which means that I have to add as (keyof typeof PROMINENCES)[] at the end. Since it is only an example, is it really necessary?

jsulpis: ok no problem 👌

philibea: in that case it's should be done inside the constants files ? 
inside the constants file 

```typescript
export const PROMINENCES = {
  default: '',
  strong: 'strong',
  stronger: 'stronger',
  weak: 'weak',
} as const

export type Prominence = keyof typeof PROMINENCES

export const PROMINENCE_ARRAY = Object.keys(PROMINENCES) as readonly Prominence[]
```

lisalupi: ok done

PR #6666, opened 2026-07-13
fix(NumberInputField): better management of NaN values + fix `isNumber` validator

jsulpis:
## Type

- Bug

#### What is expected?

NumberInputField correctly handles special characters and empty values

- update the `isNumber` validator
- update the components to fix little bugs and console warnings
- add an e2e test

### DIFF
--- a/e2e/tests/numberInputField/render.tsx
+++ b/e2e/tests/numberInputField/render.tsx
@@ -15,12 +15,7 @@ const Render = () => {
   return (
     <Form errors={mockErrors} methods={methods} onSubmit={console.log}>
       <Stack gap={2}>
-        <Stack
-          gap={2}
-          style={{
-            width: '250px',
-          }}
-        >
+        <Stack gap={2} width="250px">
           <NumberInputField min={0} max={10} name="example" label="Test" />
           <Submit>Submit</Submit>
         </Stack>
--- a/e2e/tests/numberInputField/render.tsx
+++ b/e2e/tests/numberInputField/render.tsx
@@ -1,0 +1,53 @@
+import { useForm, Submit, Form, NumberInputField } from '@ultraviolet/form'
+import { Snippet, Stack } from '@ultraviolet/ui'
+import { mockErrors } from '../../mocks/mockErrors'
+
+const Render = () => {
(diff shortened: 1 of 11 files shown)


packages/form/src/components/NumberInputField/__tests__/index.test.tsx
```diff
@@ -196,46 +196,6 @@ describe('numberInputField', () => {
       })
     })
 
-    it('should fail validation when value is not a decimal number', async () => {
```
jsulpis, line 199: moved in e2e test

packages/ui/src/components/NumberInput/index.tsx
```diff
@@ -175,7 +161,7 @@ export const NumberInput = forwardRef(
                   step={step}
                   style={style}
                   type="number"
-                  value={value?.toString()}
+                  value={value === null || Number.isNaN(value) ? '' : value}
```
jsulpis, line 164: empty string to avoid a React warning because the component alternates between controlled and uncontrolled

etienne-scaleway: So no way to have this component uncontrolled ? Won't  this have some side effects ? 🤔

jsulpis: the uncontrolled version to use with the `register` function of react-hook-form will come in another PR. 

For the controlled version I think there is no side effect

e2e/tests/numberInputField/render.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const { errors, isDirty, isValid } = methods.formState
+
+  return (
+    <Form errors={mockErrors} methods={methods} onSubmit={console.log}>
+      <Stack gap={2}>
+        <Stack
+          gap={2}
+          style={{
+            width: '250px',
+          }}
```
lisalupi:
```suggestion
            width="250px"
```

philibea (approved):

lisalupi (approved):

e2e/tests/numberInputField/render.tsx
jsulpis: ✅

PR #6675, opened 2026-07-15
chore(package): use new dev engine metadata

philibea:
## Type

- Enhancement

devEngines seems to be prefered over packageManager

https://github.com/orgs/pnpm/discussions/9817#discussioncomment-13973040

Will add the pnpm version to the pnpm lock
Capture d’écran 2026-07-15 à 16 00 58

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,11 +143,22 @@
       "path": "./node_modules/cz-conventional-changelog"
     }
   },
+  "devEngines": {
+    "packageManager": {
+      "name": "pnpm",
+      "version": "11.11.0",
+      "onFail": "download"
+    },
+    "runtime": {
+      "name": "node",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -143,6 +143,18 @@
       "path": "./node_modules/cz-conventional-changelog"
     }
   },
+  "devEngines": {
+    "packageManager": {
```
jsulpis, line 147: is it a replacement for the `packageManager` field l.162 ? (should we remove it or is it kept for compatibility with something)

jsulpis (approved):

package.json
philibea: we can deleted it  now i think, it's should be supported by renovate, corepack

lisalupi (approved):

PR #6677, opened 2026-07-16
feat(Tabs): hover state

lisalupi:
## Type

- Enhancement

#### What is expected?
`Tabs`: improve hover/focus state

| url  |Capture d’écran 2026-07-16 à 10 27 42 | Capture d’écran 2026-07-16 à 10 28 01|

### DIFF
--- a/packages/ui/src/components/Tabs/styles.css.ts
+++ b/packages/ui/src/components/Tabs/styles.css.ts
@@ -67,16 +67,18 @@ const button = style({
       outline: 'none',
       textDecoration: 'none',
     },
-    '&[aria-disabled="false"]:not(:disabled):hover': {
+    '&[aria-disabled="false"]:not(:disabled):active': {
       borderBottomColor: theme.colors.primary.border,
       color: theme.colors.primary.text,
       outline: 'none',
     },
+    '&[aria-disabled="false"]:not(:disabled):hover': {
+      background: theme.colors.neutral.backgroundHover,
+    },
     '&[aria-disabled="false"]:not(:disabled):focus-visible': {
       boxShadow: `inset ${theme.shadows.focusPrimary}`,
-      borderBottomColor: theme.colors.primary.border,
-      color: theme.colors.primary.text,
       outline: 'none',
+      background: theme.colors.neutral.backgroundHover,
     },
     '&[aria-disabled="true"], &:disabled': {
       cursor: 'not-allowed',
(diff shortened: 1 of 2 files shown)


philibea (approved):

packages/ui/src/components/Tabs/styles.css.ts
```diff
@@ -116,18 +118,19 @@ const badge = style({
...
   padding: `0 ${theme.space[1]}`,
   selectors: {
-    [`${button}[aria-selected="false"]:hover &, ${button}[aria-selected="false"]:active &`]: {
-      backgroundColor: theme.colors.primary.background,
-      borderColor: theme.colors.primary.background,
-      color: theme.colors.primary.text,
-    },
+    [`${button}:not(:disabled)[aria-selected="false"]:hover &, ${button}[aria-selected="false"]:active &, ${button}[aria-selected="false"]:focus-visible &`]:
+      {
+        backgroundColor: theme.colors.primary.background,
```
jsulpis: the badge in primary on hover is not very readable, I think it would be better to keep it neutral 

Image

Image

packages/ui/src/components/Tabs/styles.css.ts
```diff
@@ -67,16 +67,18 @@ const button = style({
...
     },
+    '&[aria-disabled="false"]:not(:disabled):hover': {
+      background: theme.colors.neutral.backgroundHover,
+    },
     '&[aria-disabled="false"]:not(:disabled):focus-visible': {
       boxShadow: `inset ${theme.shadows.focusPrimary}`,
-      borderBottomColor: theme.colors.primary.border,
-      color: theme.colors.primary.text,
       outline: 'none',
+      background: theme.colors.neutral.backgroundHover,
```
jsulpis: on keyboard focus the neutral background in the primary border looks a bit weird to me, I think the border alone is enough. WDYT ?

Image

Image

lisalupi: @mhervouet what do you think?

jsulpis (approved):

packages/ui/src/components/Tabs/styles.css.ts
jsulpis: I let you check with Mathieu, it's ok if you prefer to keep the background

mhervouet: Can we have a background for hover but no background when it's in focus? If that's okay, I'd actually prefer not to have a background when it's in focus.

PR #6680, opened 2026-07-17
fix(TextArea): fix automatic resize when rows="auto"

jsulpis:
## Type

- Bug

#### What is expected?

Textarea can be resized and keep its size and scroll position when typing in it.

- use the new CSS property field-sizing in supported browsers
- use a fallback with a less good behavior on older browsers (the textarea won't shrink back because it causes scroll issues)

#### To check the fallback behavior
- open Firefox
- go to about:config
- search for "field-sizing"
- disable `layout.css.field-sizing.enabled` using the arrow on the right

### DIFF
--- a/.changeset/lovely-stamps-battle.md
+++ b/.changeset/lovely-stamps-battle.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/ui": patch
+---
+
+`TextArea`: fix automatic resize when rows="auto"
--- a/packages/form/src/components/TextAreaField/__tests__/__snapshots__/index.test.tsx.snap
+++ b/packages/form/src/components/TextAreaField/__tests__/__snapshots__/index.test.tsx.snap
@@ -27,7 +27,7 @@ exports[`textAreaField > should render correctly 1`] = `
             id="_r_0_"
             name="test"
             rows="3"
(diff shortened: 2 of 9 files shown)


lisalupi (commented):
Image You should update this story, its description is not correct anymore with the changes you made. Why did you remove the automatic expand of the component?
```
textArea.style.height = 'auto'
textArea.style.resize = 'none'
```

philibea (approved):

jsulpis:
> Image You should update this story, its description is not correct anymore with the changes you made. Why did you remove the automatic expand of the component?
> 
> ```
> textArea.style.height = 'auto'
> textArea.style.resize = 'none'
> ```

- I updated the docs on the story
- the textarea should still automatically expand: either thanks to the native property `field-sizing`, or with the line that sets the height in inline style using the scrollHeight. The line `textArea.style.height = 'auto'` is what allowed the textarea to shrink, but I removed it since it causes a scroll jump (we can save and restore the scroll programmatically but it doesn't work well if the html has `scroll-behavior: smooth`, so I haven't found a good solution). So it's a bit less good but it's the fallback behavior for older browsers so I think it's acceptable.
- I removed the `textArea.style.resize = 'none'` because I have seen no conflit with the automatic resize behavior so I think we can keep the textarea always resizable (it can be useful even with automatic resize)

lisalupi (approved):

PR #6684, opened 2026-07-17
docs(Bullet): accessibility audit

lisalupi:
## Type

- Documentation

#### What is expected?
Audit `Bullet` accessibility

### DIFF
--- a/packages/ui/src/components/Bullet/__stories__/A11y.mdx
+++ b/packages/ui/src/components/Bullet/__stories__/A11y.mdx
@@ -13,7 +13,7 @@ A decorative component that displays content (icon or text) within a colored cir
 1. ❌ **Perceivable**: Text-to-background contrast ratios fall below 4.5:1 for several sentiment/prominence combinations
 2. ✅ **Operable**: Component is not interactive, no keyboard requirements
 3. ✅ **Understandable**: Content is readable when provided
-4. ❌ **Robust**: Uses generic `<div>` without semantic role support; lacks `aria-hidden` or `role` props for proper assistive technology integration
+4. ❌ **Robust**: Uses generic `<div>` without semantic role support; lacks `aria-hidden` or `role='img'` props for proper assistive technology integration for icon-only bullets
 
 ## ARIA Pattern
 
--- a/packages/ui/src/components/Bullet/__stories__/A11y.mdx
+++ b/packages/ui/src/components/Bullet/__stories__/A11y.mdx
@@ -1,0 +1,68 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="UI/Badges/Bullet/A11y" />
+
+# Bullet - Accessibility
+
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Bullet/__stories__/A11y.mdx
```diff
@@ -0,0 +1,68 @@
...
+## Description
+
+A decorative component that displays content (icon or text) within a colored circular background. Supports multiple sizes, sentiments, prominence variants, and optional tooltip. Used for status indicators, badges, or visual markers.
+
+## Summary
+
+1. ❌ **Perceivable**: Text-to-background contrast ratios fall below 4.5:1 for several sentiment/prominence combinations
+2. ✅ **Operable**: Component is not interactive, no keyboard requirements
+3. ✅ **Understandable**: Content is readable when provided
+4. ❌ **Robust**: Uses generic `<div>` without semantic role support; lacks `aria-hidden` or `role` props for proper assistive technology integration
```
jsulpis: I'm not sure there is a role for that if it's just a content 🤔 or maybe in the case of icons ?

packages/ui/src/components/Bullet/__stories__/index.stories.tsx
```diff
@@ -7,9 +7,9 @@ export default {
   parameters: {
     a11yStatus: {
       perceivable: false,
-      operable: false,
-      understandable: false,
-      robust: false,
+      operable: true,
+      understandable: true,
+      robust: true,
```
jsulpis: to be consistent with the other file
```suggestion
      robust: false,
```

jsulpis (approved):

packages/ui/src/components/Bullet/__stories__/A11y.mdx
lisalupi: in case of icon and maybe when it is done as a list (in the [`StepList`](https://storybook.ultraviolet.scaleway.com/?path=/docs/ui-data-display-steplist--docs&globals=theme:dark) component) but i am not sure about that

philibea (approved):

packages/ui/src/components/Bullet/__stories__/A11y.mdx
iManu: same, I think this is for icon only. It should be resolved by the icons RFC current work 

- `role="img"` with `aria-label` (or an other accessible name) when representing a status icon with meaning, or aria-hidden if a label visually exist (in StepList for instance)

iManu (approved):
The icon only case may be resolved after icon refacto I guess

packages/ui/src/components/Bullet/__stories__/A11y.mdx
lisalupi: fixed, i specified that the aria-hidden/role props apply only to icon-only bullets

PR #6691, opened 2026-07-19
fix(test): form relative path import inside 'utils/test' break resolu…

philibea:
…tion with turbo

## Type

- Bug


utils/test/src/vitest/helpers/index.tsx
```diff
@@ -1,9 +1,7 @@
-import { render, renderHook } from '@testing-library/react'
+import { render } from '@testing-library/react'
 import type { RenderOptions } from '@testing-library/react'
 import { consoleLightTheme, ThemeProvider } from '@ultraviolet/themes'
-import type { ComponentProps, ReactElement, ReactNode } from 'react'
-import type { FormErrors, UseFormProps } from '../../../../../packages/form/src'
-import { Form, useForm } from '../../../../../packages/form/src/index'
```
philibea, line 6: this relative import break.it's resolved by vitest which will try to run some test not relative to the current package.

To check, delete all */.turbo and ./dist and just run test:unit:coverage inside icons packages.

jsulpis (approved):

PR #6695, opened 2026-07-20
docs: fix Storybook docgen

jsulpis:
## Type

- Documentation

#### What is expected?

Storybook playgrounds show relevant documentation for the components properties

**Still broken**:
- props defined with `XOR` (Button, Link, Breadcrumb...). Only those part are broken, the others props are fine
- props using a `ReceipeVariants` type from Vanilla Extract (Button, Badge, Alert...)
- other complex props types

- remove [react-docgen-typescript](https://github.com/styleguidist/react-docgen-typescript) which has no commit since more than 1 year and doesn't support typescript 7
- use the default [react-docgen](https://github.com/reactjs/react-docgen) which seems more maintained and seems to work fine

| GlobalAlert  | image | image |

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -38,16 +38,6 @@ const main = defineMain({
     '../utils/stories/src/**/*.mdx',
     '../utils/stories/src/**/*.stories.tsx',
   ],
-
-  typescript: {
-    reactDocgen: 'react-docgen-typescript',
-    reactDocgenTypescriptOptions: {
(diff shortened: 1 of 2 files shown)


lisalupi (approved):
maybe we can change the broken props that get their type from vanilla extract to better fix the component stories? It’s not done consistently across the codebase anyway, so this would be a great opportunity to have a more coherent prop definition "schema".

PR #6696, opened 2026-07-21
feat(Theme): font fallback

lisalupi:
## Type

- Enhancement

#### What is expected?

Add font fallbacks directly in the theme, with fallback similar to `--font-inter`, `--font-jetbrains`, and `--font-space-grotesk` (from `@ultraviolet/fonts`).

Fallback on heading text: 
Before: 

https://github.com/user-attachments/assets/5042db8a-8b5f-439a-b04c-f05b1bc6da1f

After: 

https://github.com/user-attachments/assets/5079ac83-329a-4c65-b5c8-51652b0d59fd


philibea (approved):

philibea (approved):

utils/config/tsconfig/base.json
```diff
@@ -15,7 +15,7 @@
     "module": "esnext",
     "noEmit": true,
     "noFallthroughCasesInSwitch": true,
-
+    "allowImportingTsExtensions": true,
```
philibea, line 18: why do you need this ?
this has already been had to the root config tsconfig.json which is used for 

```"include": [
    "utils/**/*.ts",
    "utils/**/*.tsx",
    "**/*.spec.ts",
    ".storybook/**/*.ts",
    ".storybook/**/*.tsx",
    "**/__stories/*.stories.tsx"
  ]
```

lisalupi: without it I had a typecheck error:
```js
@utils/stories:typecheck: ../scripts/figma-synchronise-token/generatePalette.ts:3:32 - error TS5097: An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.
@utils/stories:typecheck: 
@utils/stories:typecheck: 3 import { FONTS_FALLBACK } from './constants.ts'
@utils/stories:typecheck:                                  ~~~~~~~~~~~~~~~~
```

And if I removed the `.ts` it broke the script ^^'

jsulpis (approved):

PR #6699, opened 2026-07-22
docs(Modal): a11y audit

lisalupi:
## Type

- Documentation

#### What is expected?
Audit `Modal` accessibility

### DIFF
--- a/packages/ui/src/components/Modal/__stories__/A11y.mdx
+++ b/packages/ui/src/components/Modal/__stories__/A11y.mdx
@@ -1,0 +1,68 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="UI/Overlay/Modal/A11y" />
+
+# Modal - Accessibility
+
+## Description
+
+Modal is a dialog component that displays content on top of other page content, commonly used for dialogs with additional information or confirmations. It supports focus trapping, keyboard navigation (Escape to close), and various placements.
+
+## Summary
+
+1. ✅ **Perceivable**: Uses `aria-label` for accessible naming, close button labeled, semantic `<dialog>` element.
+2. ⚠️ **Operable**: Keyboard support (Escape, focus trap) present, but focus management has issues: initial focus on dialog element (not inside), incomplete focus restoration for programmatic modals (with prop `open` without disclosure).
+3. ✅ **Understandable**: Clear labeling via `ariaLabel`, consistent behavior.
+4. ⚠️ **Robust**: Native `<dialog>` provides role, but missing `aria-labelledby` which may be less optimal for modals with visible titles, not using native `showModal()`/`close()` methods. The `preventBodyScroll` prop should be removed as modals should always prevent background scrolling.
+
+## ARIA Pattern
+
+[Dialog (Modal) Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)
+
+## Rules
+
+Enforced by the component:
+
+- ✅ [WCAG 2.1.2 - No Keyboard Trap (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/no-keyboard-trap.html) - Focus trap via `handleFocusTrap`
+- ⚠️ [WCAG 2.4.3 - Focus Order (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html) - Initial focus on dialog element (not inside), incomplete focus restoration: it only works when a `disclosure` element is provided. Programmatic modals don't restore focus to a logical element on close
+- ⚠️ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) - Missing `aria-labelledby` and `aria-describedby` (optional)
+
+To apply when using the component:
+
+- [WCAG 1.3.1 - Info and Relationships (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) - Ensure semantic structure (headings, lists) within modal content
+- [WCAG 1.4.11 - Non-text Contrast (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html) - Ensure 3:1 contrast for interactive elements within modal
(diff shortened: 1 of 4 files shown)


lisalupi:
I added a story to test focus, I left it but it can be removed

packages/ui/src/components/Modal/__stories__/A11y.mdx
```diff
@@ -0,0 +1,68 @@
...
+# Modal - Accessibility
+
+## Description
+
+Modal is a dialog component that displays content on top of other page content, commonly used for dialogs with additional information or confirmations. It supports focus trapping, keyboard navigation (Escape to close), and various placements.
+
+## Summary
+
+1. ✅ **Perceivable**: Uses `aria-label` for accessible naming, close button labeled, semantic `<dialog>` element.
+2. ⚠️ **Operable**: Keyboard support (Escape, focus trap) present, but focus management has issues: initial focus on dialog element (not inside), incomplete focus restoration for programmatic modals (with prop `open` without disclosure).
```
iManu, line 14: The component use Dialog, which is already a dialog native element but it do not use the features like showModal() functions etc, or ::backdrop css selector etc. 
The support is good in 2026, even using fallbacks or polyfills, it can be beneficial to clean large parts of js handlers and cover conformity better

packages/ui/src/components/Modal/__stories__/A11y.mdx
```diff
@@ -0,0 +1,68 @@
...
+3. ✅ **Understandable**: Clear labeling via `ariaLabel`, consistent behavior.
+4. ⚠️ **Robust**: Native `<dialog>` provides role, but `aria-modal` always `true` (even when background not inert), missing `aria-labelledby` which may be less optimal for modals with visible titles, not using native `showModal()`/`close()` methods.
```
iManu: A lot a aria can be cleaned and/or optimised for this component and related components

iManu (approved):

packages/ui/src/components/Modal/__stories__/A11y.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+[Dialog (Modal) Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)
+
+## Rules
+
+Enforced by the component:
+
+- ✅ [WCAG 2.1.2 - No Keyboard Trap (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/no-keyboard-trap.html) - Focus trap via `handleFocusTrap`
+- ⚠️ [WCAG 2.4.3 - Focus Order (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html) - Initial focus on dialog element (not inside), incomplete focus restoration: it only works when a `disclosure` element is provided. Programmatic modals don't restore focus to a logical element on close
+- ⚠️ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) - Missing `aria-labelledby`, `aria-modal` always `true` even when the background content is not truly inert (`preventBodyScroll={false}`)
```
jsulpis: Unless I miss something, the Modal component is always modal (i.e. blocks interaction with the rest of the page and hides it with a backdrop), so it makes sense to always have `aria-modal="true"`. Have you seen some guidelines that say otherwise ? 

Also, IMHO the scroll should always be locked behind the modal (the content is supposed to be hidden behind the backdrop), so this should not affect the `aria-modal` attribute

packages/ui/src/components/Modal/__stories__/A11y.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+## Accessibility issues
+
+### Medium
+
+- **Incorrect initial focus placement**: Focus lands on `<dialog>` element itself instead of an element **inside** the dialog: screen reader users hear container focus instead of meaningful content.
+
+> Focus should go on the first focusable element inside (unless it causes a scroll in the page), an element with `tabindex="-1"` (title/paragraph) at the start for complex content, the least destructive action of irreversible actions (prefer `Dialog` for this use case), or the most frequently used element for simple _OK/Continue_ dialogs
+
+- **Missing `title` prop and `aria-labelledby`**: Only supports `aria-label`, not `aria-labelledby` for visible titles. Dialogs should have `aria-labelledby` referencing a visible title (preferred) or `aria-label` fallback. Adding a `title` prop that renders visible heading and sets `aria-labelledby` automatically would ensures that screen reader users hear the same title as sighted users without requiring developers to manually wire up IDs.
```
jsulpis: maybe `heading` instead of `title` to avoid a confusion with the native `title` attribute ?

packages/ui/src/components/Modal/__stories__/A11y.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+- **`aria-modal` always true**: Set unconditionally, even when `preventBodyScroll={false}`. Only set to true when background is truly inert.
+
+- **Not using native `<dialog>` methods**: Uses a native `<dialog>` element but doesn't leverage `showModal()` and `close()` methods. Using native methods would provide better built-in accessibility: automatic focus management & focus trap, native Escape key handling, and proper focus return on close. The current custom implementation may have edge cases not covered by the manual focus trap.
```
jsulpis: Note: I agree that native dialog methods would be ideal but we will have to be careful with the integration of all other components, because `showModal` promotes the dialog on the top layer, above everything else (even elements at the root of the body with high z-index). So the Modal would be above the SelectInput popup for instance. We will have to work on all components that have a popup before using the `showModal` method

jsulpis:
> I added a story to test focus, I left it but it can be removed

we could keep it until we fix the focus management in the component and then remove it

jsulpis (approved):

packages/ui/src/components/Modal/__stories__/A11y.mdx
lisalupi: I was referring to the fact that `preventBodyScroll={false}` allows scrolling on the page, so in my mind the content wasn't fully inert. But I can note to remove this prop (since the content shouldn't be scrollable in a "real" modal).

packages/ui/src/components/Modal/__stories__/A11y.mdx
lisalupi: I updated the audit to warn about those methods

PR #6702, opened 2026-07-23
fix(FileInput): onChange behavior

lisalupi:
## Type

- Bug

#### What is expected?

`FileInput`: add more file check to ensure an `undefined` object is not added to dataTransfer (onChange behavior).

There were instances where `[formattedNewFiles[0]` returned `[undefined]` instead of `[]` which then caused an error when doing `dataTransfer.items.add()`

### DIFF
--- a/.changeset/six-jars-design.md
+++ b/.changeset/six-jars-design.md
@@ -1,0 +1,5 @@
+---
+"@ultraviolet/ui": patch
+---
+
+`FileInput`: add more file check to ensure an `undefined` object is not added to dataTransfer (onChange behavior)
--- a/packages/ui/src/components/FileInput/index.tsx
+++ b/packages/ui/src/components/FileInput/index.tsx
@@ -138,7 +138,10 @@ const FileInputBase = ({
       }),
     )
 
-    const formattedFiles = computedMultiple ? [...files, ...formattedNewFiles] : [formattedNewFiles[0]]
+    const formattedFiles = (computedMultiple ? [...files, ...formattedNewFiles] : [formattedNewFiles[0]]).filter(
+      Boolean,
+    )
+
     setFiles(formattedFiles)
     onChangeFiles?.(formattedFiles)
 


Echo-IV (approved):

.changeset/six-jars-design.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/ui": patch
+---
+
+`FileInput`: add more file check to ensure an `undefined` object is not added to dataTrasnfer (onChange behavior)
```
philibea:
```suggestion
`FileInput`: add more file check to ensure an `undefined` object is not added to dataTransfer (onChange behavior)
```

packages/ui/src/components/FileInput/index.tsx
```diff
@@ -138,7 +138,10 @@ const FileInputBase = ({
       }),
     )
 
-    const formattedFiles = computedMultiple ? [...files, ...formattedNewFiles] : [formattedNewFiles[0]]
+    const formattedFiles = (computedMultiple ? [...files, ...formattedNewFiles] : [formattedNewFiles[0]]).filter(
```
philibea, line 141: question(non-blocking): should we add this check also here
 ```
[...files, ...formattedNewFiles]
```

philibea (approved):

jsulpis (commented):
is it possible to add a test or too complicated ?

packages/ui/src/components/FileInput/index.tsx
lisalupi: `.filter` is done on both `[...files, ...formattedNewFiles]`and `[formattedNewFiles[0]]`

lisalupi:
> is it possible to add a test or too complicated ?

We already have a test for the onchange with accept, and I don't really know how to test what I did

jsulpis (approved):

PR #6726, opened 2026-07-28
docs(Icon): a11y audit

jsulpis:
## Type

- Documentation

#### What is expected?

Audit page aligned with the decision in the RFC [Accessible Icons and Hidden Labels](https://github.com/scaleway/ultraviolet/discussions/6585)

Audit the Icon component using the skill

### DIFF
--- a/packages/icons/src/components/Icon/__stories__/A11y.mdx
+++ b/packages/icons/src/components/Icon/__stories__/A11y.mdx
@@ -1,0 +1,68 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="Icons/Icon/A11y" />
+
+# Icon - Accessibility
+
+## Description
+
+Base SVG icon component for the design system. It renders an `<svg>` with configurable size, sentiment color, prominence and disabled state. Every generated icon (e.g. `ArrowRightIcon`, `AddressIcon`) is a thin wrapper around this base component, passing its SVG paths as children and a `title` prop set to the icon's technical name.
+
+## Summary
+
+1. ❌ **Perceivable**: icons are not hidden from assistive technologies by default and generated icons expose a meaningless technical name (e.g. `ArrowRightIcon`) as their accessible name.
+2. ✅ **Operable**: the icon is a non-interactive element; operability depends on the interactive parent it is placed inside (button, link, etc.).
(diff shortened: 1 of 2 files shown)


packages/icons/src/components/Icon/__stories__/A11y.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+**High:**
+
+- No explicit `role="img"`. The `<svg>` element does not declare `role="img"`. While some browsers map an SVG containing a `<title>` to the `img` role, this mapping is inconsistent across browsers and assistive technologies. Explicitly setting `role="img"` (only when the icon is informative and not `aria-hidden`) would make identification robust. This affects WCAG 4.1.2.
+- `<title>` is always rendered and the `title` prop is required. Because `title` is a required prop, a `<title>` element is always inserted inside the SVG. When the icon is decorative (`aria-hidden="true"`), the title is ignored by assistive technologies but still present in the DOM, which is contradictory. Consumers are also forced to provide a value even for decorative icons. The `title` prop should be optional and the `<title>` element should only be rendered when the icon is informative.
```
lisalupi: `Consumers are also forced to provide a value even for decorative icons`
No, since the `Icon` component is not exposed, only the generated ones with the default title. The prop on the consumer components (`InformationIcon`, `FolderIcon` etc etc) do not have a "title" prop (it is explicitly omitted). 
Currently: 
1. All generated icon have a default title (corresponding to the icon name), added during the generation phase
2. It is **not** possible to change it (their type is `Omit<IconProps, 'children' | 'title'>`)

I think that it would be clearer to say something along the lines of 

> `<title>` is always rendered and cannot be changed. When the icon is decorative (`aria-hidden="true"`), the title is ignored by assistive technologies but still present in the DOM. The *title* prop should removed from`Icon` and added to generated component (i.e. `InformationIcon`, `FolderIcon` etc.) as an optional property to provide an adequate value for informative icons.

I'm not sure if I'm being clear ^^'

jsulpis: yes you're right, I updated the paragraph. I suggested to name the property `accessibleLabel` to be consistent with the RFC, instead of title. Since the prop is currently not exposed we can safely rename it without breaking change. wdyt ?

lisalupi (approved):

PR #6727, opened 2026-07-29
fix(InfoTable): style and overflow

lisalupi:
## Type

- Bug
- Enhancement

#### What is expected?

`InfoTable.Cell`: 
- should not override the custom style of its children.
- when children are an array of strings or numbers, the same style is applied as for a single string.

`Text`: updated tooltip detection logic to handle overflowing content when children are of type `(string | number)[]`. Such arrays are now converted to strings for tooltip display. For example:
```js
<Text as="p" variant="body">Ready?{isReady ? "Yes!" : "No..."}</Text>
```
Previously, no tooltip appeared when overflowing since the computed child was `['Ready?', 'Yes!']` (or `['Ready?', 'No...']`). Now, the tooltip correctly displays `Ready? Yes!`.

### DIFF
--- a/packages/ui/src/helpers/isStringNumberArray.ts
+++ b/packages/ui/src/helpers/isStringNumberArray.ts
@@ -1,0 +1,6 @@
+export const isStringNumberArray = (value: unknown): value is (string | number)[] => {
+  return (
+    Array.isArray(value) &&
+    value.every((item): item is string | number => typeof item === 'string' || typeof item === 'number')
+  )
+}
--- a/packages/ui/src/helpers/isStringNumberArray.ts
+++ b/packages/ui/src/helpers/isStringNumberArray.ts
@@ -1,6 +1,0 @@
-export const isStringNumberArray = (value: unknown): value is (string | number)[] => {
-  return (
-    Array.isArray(value) &&
-    value.every((item): item is string | number => typeof item === 'string' || typeof item === 'number')
-  )
-}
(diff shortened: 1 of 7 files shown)


packages/ui/src/helpers/recursivelyGetChildrenString.ts
```diff
@@ -1,12 +1,24 @@
...
   if (Array.isArray(children)) {
-    return ''
-  } // We can't determine which string to display in tooltip
+    let finalString = ''
+    for (const child of children) {
+      finalString = finalString.concat(' ', recursivelyGetChildrenString(child))
+    } // Recursively add every string or number of the array
+
+    return finalString
+  }
```
jsulpis: suggestion:
```ts
  if (typeof children === 'string' || typeof children === 'number') {
    return String(children)
  }

  if (Array.isArray(children)) {
    return children
      .map(child => recursivelyGetChildrenString(child))
      .filter(Boolean)
      .join(' ')
  }
```

`filter(Boolean)` avoids a double space if there is a `null` somewhere in the array (which could happen I guess with ternary expressions)

packages/ui/src/helpers/isStringNumberArray.ts
```diff
@@ -0,0 +1,6 @@
+export const isStringNumberArray = (value: unknown): value is (string | number)[] => {
```
jsulpis: more clear ?
```suggestion
export const isStringOrNumberArray = (value: unknown): value is (string | number)[] => {
```

jsulpis (approved):

philibea (approved):

packages/ui/src/helpers/recursivelyGetChildrenString.ts
lisalupi: done

PR #6729, opened 2026-07-29
feat: improve Modal/Dialog/Drawer animations

jsulpis:
## Type

- Enhancement

#### What is expected?

Smooth opening/closing animations

- update the Modal component to delay the unmount when closing
- update styles
- update the Dialog action button to trigger a different animation than when "canceling" the Dialog


packages/ui/src/compositions/Navigation/constants.ts
```diff
@@ -8,4 +8,4 @@ export const NAVIGATION_MAX_WIDTH = 320 // in px
  * collapse and expand animation duration of the whole nav in ms
  */
 export const ANIMATION_DURATION = 250
-export const ANIMATION_EASING = 'cubic-bezier(0.5, 0, 0.2, 1)'
+export const ANIMATION_EASING = 'cubic-bezier(0.5, 0, 0.1, 1)'
```
jsulpis, line 11: I updated the curve on the Navigation component to be the same as the Drawer, and the duration is the same too, so that we can mutualize them later in the animations package. I'm not doing it now because I don't really know in what format to export them (JS constant, vanilla extract variables, CSS variables...)

packages/ui/src/components/Modal/components/Dialog.tsx
```diff
@@ -35,34 +35,33 @@ export const Dialog = ({
...
+  const { registerModal, unregisterModal, openedModals } = useModal()
 
   // register/unregister the modal to handle nested modals
   useEffect(() => {
     // a drawer should not be registered since it does not stack with other modals
-    if (!isDrawer) {
+    if (open && !isDrawer) {
       registerModal({ id, ref: dialogRef })
     }
+    if (!open) {
```
lisalupi:
```suggestion
    if (!open && !isDrawer) {
```
i'm not sure if it necessary here. Without it, there might be an issue when you open/close a modal from a drawer

packages/ui/src/components/Modal/constants.css.ts
```diff
@@ -0,0 +1,88 @@
...
+export const animationDuration = createVar()
+export const animationDurationTranslation = createVar()
+export const offscreenTranslation = createVar()
+
+export const MODAL_PLACEMENT: Record<ModalPlacement, StyleRule> = {
+  bottom: {
+    margin: 'auto',
+    marginBottom: 0,
+    vars: {
+      [offscreenTranslation]: 'translateY(8px)',
```
lisalupi:
```suggestion
      [offscreenTranslation]: `translateY(`theme.space.[1]`)`,
```
🤔?

packages/ui/src/components/Modal/ModalContent.tsx
```diff
@@ -89,4 +102,21 @@ export const ModalContent = ({
...
+    if (open && !shouldRender) {
+      setShouldRender(true)
+    } else if (!open && shouldRender) {
+      timeoutId = setTimeout(() => setShouldRender(false), delayTime)
+    }
+    return () => clearTimeout(timeoutId)
+  }, [open, delayTime, shouldRender])
+
+  return shouldRender
+}
```
lisalupi:
```suggestion
const useDelayUnmount = (open?: boolean, delayTime?: number) => {
  const [shouldRender, setShouldRender] = useState(open)

  useEffect(() => {
    let timeoutId: ReturnType<typeof setTimeout>
    if (open && !shouldRender) {
      setShouldRender(true)
    } else if (!open && shouldRender) {
      timeoutId = setTimeout(() => setShouldRender(false), delayTime)
    }
    return () => clearTimeout(timeoutId)
  }, [open, delayTime, shouldRender])

  return shouldRender
}
```

I'd put it before the component or in another file

packages/ui/src/components/Modal/constants.css.ts
jsulpis: ✅

packages/ui/src/components/Modal/ModalContent.tsx
jsulpis: I moved it in another file

packages/ui/src/components/Modal/components/Dialog.tsx
jsulpis: I checked with nested modals triggered from the Drawer:
- the "registerModal" thing only handles the position and scale of nested modals using CSS variables
- those CSS variables are not used (or overridden) in the Drawer
- the position/scale of a nested modal is based on the modal _on top of it_, so since the Drawer is behind, it's not involved in the calculations
- so nothing changes if I keep or remove the condition

I think it would only have an effect if we triggered a Drawer _from_ a Modal. I guess we don't and will never want to deal with that ?

If it's the same, I think it's more consistent to keep the condition `!isDrawer` since the Drawer should not be considered in the list of nested modals

lisalupi: > If it's the same, I think it's more consistent to keep the condition !isDrawer since the Drawer should not be considered in the list of nested modals

But you did not keep it to unregister, which was my point

lisalupi (approved):

packages/ui/src/components/Modal/components/Dialog.tsx
jsulpis: Oh right I thought you talked about the first `if` above 🤦 
I fixed it

mhervouet (approved):

PR #6732, opened 2026-07-30
docs(TextInput): audit accessibility

lisalupi:
## Type

- Documentation

#### What is expected?

Audit `TextInput` accessibility

### DIFF
--- a/packages/ui/src/components/TextInput/__stories__/A11y.mdx
+++ b/packages/ui/src/components/TextInput/__stories__/A11y.mdx
@@ -1,0 +1,63 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="UI/Data Entry/TextInput/A11y" />
+
+# TextInput - Accessibility
+
+## Description
+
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/TextInput/__stories__/A11y.mdx
```diff
@@ -0,0 +1,63 @@
...
+## Accessibility issues
+
+**High:**
+
+- Show/hide password button missing `aria-pressed` state to indicate current visibility state. The button toggles between show/hide but screen readers cannot determine the current state. Should use `aria-pressed={isPasswordVisible}` or `aria-expanded={isPasswordVisible}`.
+- Uses deprecated `aria-label` pattern. Per [RFC #6585](https://github.com/scaleway/ultraviolet/discussions/6585), should migrate to `accessibleLabel` prop for hidden labels or `tooltipLabel` for tooltip-based labels. Current `aria-label` prop should be deprecated in favor of the new pattern.
+
+**Medium:**
+
+- Password visibility toggle buttons lack accessible labels that indicate the action. "show" and "hide" are ambiguous without context. Should be "Show password" and "Hide password", or use `tooltipLabel` prop.
```
jsulpis: I think the `tooltipLabel` prop should define the label of the field itself and not some subcomponents, right ? Maybe we could have a `labels` prop as an object with different labels used in the component. Same for the clear button below

jsulpis (approved):

PR #6733, opened 2026-07-31
docs(NumberInput): audit accessibility

lisalupi:
## Type

- Documentation
#### What is expected?
Audit `NumberInput` accessibility

### DIFF
--- a/packages/ui/src/components/NumberInput/__stories__/A11y.mdx
+++ b/packages/ui/src/components/NumberInput/__stories__/A11y.mdx
@@ -28,7 +28,7 @@ Enforced by the component:
 - ⚠️ [WCAG 2.1.1 - Keyboard (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html): Native type="number" provides arrow key increment/decrement; buttons are clickable, but shouldn't be focusable.
 - ✅ [WCAG 3.3.1 - Error Identification (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/error-identification.html): Error state communicated via error prop and linked description
 - ✅ [WCAG 3.3.2 - Labels or Instructions (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/labels-or-instructions.html): Labels provided via label prop with optional helper text
-- ⚠️ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html): Native input provides role/value but missing aria-invalid for error state; icons should follow RFC #6585
+- ⚠️ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html): Native input provides role/value but missing aria-invalid for error state; icons should follow RFC #6585; buttons should have aria-controls pointing to input
 
 To apply when using the component:
 
@@ -43,6 +43,7 @@ To apply when using the component:
 - When increment/decrement buttons are clicked, focus remains on the button rather than returning to the input, which can be disorienting for keyboard users.
 - When the `error` prop is set, the input should have `aria-invalid="true"` to inform assistive technologies of the invalid state. Currently the error is only communicated visually and via `aria-describedby`, but not programmatically as invalid.
 - Icons in Controls should be aria-hidden. The accessible name should be provided on the interactive parent (Button).
+- The increment/decrement buttons should have `aria-controls` pointing to the input's ID to explicitly indicate which element they control, following ARIA best practices for spinbutton patterns.
 
 **Medium:**
 
--- a/packages/ui/src/components/NumberInput/__stories__/A11y.mdx
+++ b/packages/ui/src/components/NumberInput/__stories__/A11y.mdx
@@ -1,0 +1,55 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="UI/Data Entry/NumberInput/A11y" />
+
+# NumberInput - Accessibility
+
+## Description
+
+A form control for entering numeric values with optional increment/decrement buttons, min/max constraints, step values, and validation states (error, success, disabled, read-only). Uses native `<input type="number">` which provides built-in spinbutton semantics and keyboard support.
+
+## Summary
+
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/NumberInput/__stories__/A11y.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+- [WCAG 1.3.5 - Identify Input Purpose (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/identify-input-purpose.html): Consider autocomplete for user-specific numeric data.
+- [WCAG 2.4.6 - Headings and Labels (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/headings-and-labels.html): Ensure labels are descriptive. Do **not** use placeholder as a label.
+- **When using external `Label` or `Notice` components**: If you render `Label` and/or `Notice` (for helper/error/success messages) externally instead of using the built-in props, you **must** manually manage ARIA associations.
+
+## Accessibility issues
+
+**High:**
+
+- When increment/decrement buttons are clicked, focus remains on the button rather than returning to the input, which can be disorienting for keyboard users.
```
jsulpis, line 43: I am not sure about that, in the [ARIA example](https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/examples/quantity-spinbutton/) the +/- buttons keep the focus 🤔  but in [Base UI](https://base-ui.com/react/components/number-field) the focus is given to the input, so I don't know. Maybe we can try to do it and see the complexity of implementing it (and always check with a screen reader the difference between implementations).

The two examples however have `aria-controls` attributes on the +/- buttons so I think we can add that as well.

jsulpis (approved):

packages/ui/src/components/NumberInput/__stories__/A11y.mdx
lisalupi: no it doesn't? they have a tabIndex -1 and are not focusable

lisalupi: in the aria example

lisalupi: but you do lose the focus in the input that is correct

jsulpis: yes you can see the focused element in the console using a live expression and `document.activeElement` (I have this one permanently on my devtools). 

`tabindex="-1"` only prevents keyboard focus, but the element is still focusable with the pointer or JavaScript

image

lisalupi: they say "Focus remains on the text field during operation" so I think we should try to have focus only on the input, even if they don't do it in their example

jsulpis: They say that in the "Keyboard Interaction" section so I guess it's when pressing the arrow keys etc ? 

But either way is fine to me, you can resolve this conversation when you want

packages/ui/src/components/NumberInput/__stories__/A11y.mdx
```diff
@@ -0,0 +1,56 @@
...
+# NumberInput - Accessibility
+
+## Description
+
+A form control for entering numeric values with optional increment/decrement buttons, min/max constraints, step values, and validation states (error, success, disabled, read-only). Uses native `<input type="number">` which provides built-in spinbutton semantics and keyboard support.
+
+## Summary
+
+1. ✅ **Perceivable**: labels and error/helper text properly associated with input
+2. ⚠️ **Operable**: native arrow key support from type="number"; increment/decrement buttons should not be focusable (tabindex="-1")
```
iManu, line 14: Si ça parle des boutons + et - je ne suis pas trop d'accord, ils fonctionnent au clavier 

Thought: modulo que le nouveau nombre n'est pas restitué par VoiceOver, on doit pouvoir ajouter un comportement aria live (et/ou role region) pour que la valeur pour chaque increment/decrement soit restitué

packages/ui/src/components/NumberInput/__stories__/A11y.mdx
```diff
@@ -0,0 +1,56 @@
...
+3. ✅ **Understandable**: labels, error/success states, and helper text provide clear instructions
+4. ⚠️ **Robust**: missing aria-invalid for error state; icons should follow RFC #6585 (aria-hidden with accessible name on parent)
```
iManu, line 16: je peux entrer la lettre "e" à la main à la place du nombre, plutôt étrange, mais effectivement ça provoque un feedback error avec un border rouge uniquement, ce qui n'est pas explicite en plus de l'absence d'aria-invalid

packages/ui/src/components/NumberInput/__stories__/A11y.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+[WAI-ARIA Spinbutton Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/)
+
+## Rules
+
+Enforced by the component:
+
+- ⚠️ [WCAG 1.1.1 - Non-text Content (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html): Icons in controls have aria-label on buttons, but per RFC #6585 should be aria-hidden="true" with label on parent via `VisuallyHidden`
+- ✅ [WCAG 1.3.1 - Info and Relationships (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html): Label associated via htmlFor/id; error/helper linked via aria-describedby
+- ⚠️ [WCAG 2.1.1 - Keyboard (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html): Native type="number" provides arrow key increment/decrement; buttons are clickable, but shouldn't be focusable.
```
iManu, line 28: à creuser, il a peut–être raison ou alors la règle est mal interprétée ?

iManu (approved):

packages/ui/src/components/NumberInput/__stories__/A11y.mdx
lisalupi: I tested, voice over says the new number both  when using the arrow keys and typing in the component

PR #6734, opened 2026-08-03
fix(Modal): don't close the modal when clicking on the dialog element and releasing on the backdrop

jsulpis:
## Type

- Bug

#### What is expected?

When pressing the mouse on the dialog and releasing on the backdrop, the Modal shouldn't close.

- save the pointer position on pointerdown event
- close the modal if the pointer position was outside the dialog when pointer down AND up
- add and update tests

### DIFF
--- a/packages/ui/src/components/Modal/__tests__/index.test.tsx
+++ b/packages/ui/src/components/Modal/__tests__/index.test.tsx
@@ -288,42 +288,117 @@ describe('modal', () => {
     expect(mockOnClose).not.toHaveBeenCalled()
   })
 
-  it('test hideOnClickOutside is true', async () => {
-    const mockOnClose = vi.fn(() => {})
+  it('should close when clicking on the backdrop with hideOnClickOutside', async () => {
+    const mockOnBeforeClose = vi.fn(() => {})
     renderWithTheme(
       <Modal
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Modal/__tests__/index.test.tsx
```diff
@@ -288,42 +288,117 @@ describe('modal', () => {
...
       >
-        <div> test</div>
+        <div>test</div>
       </Modal>,
     )
-    await userEvent.click(screen.getByRole('dialog'))
+
+    await userEvent.click(screen.getByRole('button', { name: 'Open' }))
+
+    const dialog = screen.queryByRole('dialog')
```
lisalupi, line 307: why not `screen.getByRole('dialog')` here?

lisalupi (approved):

packages/ui/src/components/Modal/__tests__/index.test.tsx
jsulpis: I reused it to check that it's not in the DOM later, but `getByRole` fails if the element in not in the DOM, whereas `queryByRole` doesn't fail. I could have used two queries, `getByRole` and then `queryByRole`, but I thought it would be simpler this way

PR #6738, opened 2026-08-03
fix(Key): prominence strong style issues

lisalupi:
## Type

- Bug

#### What is expected?
`Key`: remove border on prominence strong and fix disabled style when prominence strong and sentiment primary

### DIFF
--- a/packages/ui/src/components/Key/styles.css.ts
+++ b/packages/ui/src/components/Key/styles.css.ts
@@ -16,16 +16,9 @@ const key = recipe({
     {
       style: {
         backgroundColor: theme.colors.neutral.backgroundStronger,
-        borderColor: theme.colors.neutral.border,
+        border: 'none',
       },
       variants: { prominence: 'strong', sentiment: 'neutral' },
-    },
-    {
-      style: {
-        backgroundColor: theme.colors.neutral.backgroundStrongerDisabled,
-        borderColor: theme.colors.neutral.borderDisabled,
-      },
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Key/styles.css.ts
```diff
@@ -38,17 +37,9 @@ const key = recipe({
     {
       style: {
         backgroundColor: theme.colors.primary.backgroundStrong,
-        border: theme.colors.neutral.border,
```
jsulpis, line 41: I can see that there is still a border on the Tags with prominence strong (both sentiments): the `border: theme.colors.neutral.border` actually removed the border because this is the shorthand property and it doesn't contain "solid". In order to override the color, it should be `border-color`. So if you want to really remove the border, it should be `border: none` in the two styles of the prominence strong

packages/ui/src/components/Key/styles.css.ts
```diff
@@ -57,6 +48,13 @@ const key = recipe({
       },
       variants: { disabled: true, sentiment: 'primary' },
     },
+    {
+      style: {
+        backgroundColor: theme.colors.primary.backgroundStrongDisabled,
+        border: theme.colors.neutral.borderDisabled,
```
jsulpis: same thing here, the shorthand property removes the border. If this is intentional, `border: none` would be more explicit, otherwise use `border-color` (and I guess it should not be "neutral" on the primary sentiment

jsulpis (approved):

philibea (approved):

PR #6760, opened 2026-08-04
feat(Table): new prop size

lisalupi:
## Type

- Enhancement

#### What is expected?
`Table`: new prop "size"


jsulpis (commented):
I see a few diffs from the design:
- the header cells should have the same padding as body cells so that they stay aligned

Image

- on the large size, the vertical padding (top/bottom) is 12px and not 16px
- on the stories with a checkbox column on the left, the padding-left of the checkbox is too big and there is an empty space on the right

Image

jsulpis (approved):

mhervouet (approved):

PR #6762, opened 2026-08-05
fix(form): omit doesn't respect xor types

philibea:
- Bug

Use A distributive version of `Omit` that preserves union types (including XOR patterns).


.changeset/tricky-hornets-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ultraviolet/form": minor
+---
+
+Use A distributive version of `Omit` that preserves union types (including XOR patterns).
```
jsulpis: maybe announce more precisely the impacts here ? (label / aria-label are now required on components XXX)

packages/form/src/components/TimeInputField/__stories__/TimeFormat.stories.tsx
```diff
@@ -12,4 +12,4 @@ export const TimeFormat: StoryFn<ComponentProps<typeof TimeInputField>> = args =
   </Stack>
 )
 
-TimeFormat.args = { ...Template.args, label: '12-hour format', timeFormat: 12 }
+TimeFormat.args = { ...Template.args, timeFormat: 12 }
```
jsulpis: why no label ?

packages/form/src/components/KeyValueField/index.tsx
```diff
@@ -21,8 +21,8 @@ type KeyValueFieldProps<
   TFieldName extends FieldPath<TFieldValues>,
 > = {
   name: TFieldArrayName
-} & Omit<ComponentProps<typeof KeyValueInput>, 'value' | 'onChange'> &
-  BaseFieldProps<TFieldValues, TFieldName>
+} & DistributiveOmit<ComponentProps<typeof KeyValueInput>, 'value' | 'onChange'> &
+  DistributiveOmit<BaseFieldProps<TFieldValues, TFieldName>, 'onBlur'>
```
jsulpis, line 25: why Omit the onBlur ?

packages/form/src/components/RadioField/__stories__/Template.stories.tsx
```diff
@@ -1,5 +1,8 @@
 import type { StoryFn } from '@storybook/react-vite'
-import type { ComponentProps } from 'react'
+import type { ComponentProps, ComponentType } from 'react'
 import { RadioField } from '..'
 
-export const Template: StoryFn<ComponentProps<typeof RadioField>> = args => <RadioField {...args} label="Option 1" />
+export const Template: StoryFn<ComponentProps<typeof RadioField>> = args => {
+  const Component = RadioField as ComponentType<Record<string, unknown>>
+  return <Component {...args} label="Option 1" />
```
jsulpis: there was a type error without this refactoring ?

packages/form/src/components/KeyValueField/index.tsx
philibea: The KeyValueInput is a little bit different and the type of the onBlur is 
```
onBlur?: (keyValues: KeyValuePair[], index?: number) => void
```
which doesn't not respect the default 
```
 onBlur?: (event?: FocusEvent) => void
``` 
we can have on other fields.

If we let the default "onBlur" we have a type like this at the end 
```
onBlur: 
(((keyValues: KeyValuePair[], index?: number) => void)
 & 
((event?: React.FocusEvent) => void)) | undefined
```

packages/form/src/components/RadioField/__stories__/Template.stories.tsx
philibea: there was an issue with the type if we let the default label="Option 1" .
I think it's should be define be other stories instead of the Template so i removed the label by default.

packages/form/src/components/TimeInputField/__stories__/TimeFormat.stories.tsx
philibea: Isse with the time coming from Template.Arg which is retype by storybook.

.changeset/tricky-hornets-poke.md
jsulpis: still not very explicit 😅
suggestion:
"Fix a type constraint that was ignored on components using XOR patterns: some XxxField components now require either a `label` or an `aria-label` property"

(I removed the components list because I'm not sure it's exhaustive)

packages/ui/src/components/TimeInput/index.tsx
```diff
@@ -54,7 +54,7 @@ type TimeInputProps = {
   'aria-describedby'?: string
 } & (
   | {
-      label?: string
+      label: string
```
jsulpis: could we keep this change for later ? It's another breaking type not related to Omit (the only in package ui)

philibea: this component is experimental actually so i changed the type as the type as it's written doesn't work actually, but ok i revert.

/**
 * A time input component that allows users to type a time in a 24 or 12-hour format.
 * @experimental This component is experimental and may be subject to breaking changes in the future.
 */

jsulpis (approved):

PR #6767, opened 2026-08-06
feat(VisuallyHidden): new component

lisalupi:
## Type

- Feature

#### What is expected?
New component `VisuallyHidden`: use this component to visually hide content while leaving it available to screen readers. It is preferred to `aria-label` for accessible labels.
Capture d’écran 2026-08-06 à 16 56 03
Capture d’écran 2026-08-06 à 16 56 09

### DIFF
--- a/packages/ui/src/components/VisuallyHidden/__stories__/Role.stories.tsx
+++ b/packages/ui/src/components/VisuallyHidden/__stories__/Role.stories.tsx
@@ -1,0 +1,17 @@
+import { Template } from './Template.stories'
+
+export const As = Template.bind({})
+
+As.args = {
+  children: 'Hidden text',
+  as: 'div',
+}
+
+As.parameters = {
+  docs: {
+    description: {
+      story:
+        'You can change the element type using prop `as`. Do not use any interactive element, as the element is not visible.',
+    },
+  },
+}
--- a/packages/ui/src/components/VisuallyHidden/__stories__/Role.stories.tsx
+++ b/packages/ui/src/components/VisuallyHidden/__stories__/Role.stories.tsx
@@ -11,7 +11,7 @@ As.parameters = {
   docs: {
     description: {
       story:
-        'You can change the element type using prop `as`. Do not use any interactive element, as the element is not visible.',
+        'You can change the element type using prop `as`. Be careful when using interactive elements (button, a, etc.), as the element is not visible unless focused (via Tab).',
     },
   },
 }
--- a/packages/ui/src/components/VisuallyHidden/__stories__/index.stories.tsx
+++ b/packages/ui/src/components/VisuallyHidden/__stories__/index.stories.tsx
@@ -1,0 +1,20 @@
+import type { Meta } from '@storybook/react-vite'
+import { VisuallyHidden } from '..'
+
+export default {
+  component: VisuallyHidden,
+  title: 'UI/Other/VisuallyHidden',
+  parameters: {
+    a11yStatus: {
+      perceivable: true,
+      operable: true,
+      understandable: true,
+      robust: true,
+    },
+  },
+} as Meta
+
+export { Playground } from './Playground.stories'
+export { As } from './Role.stories.tsx'
+export { Example } from './Example.stories.tsx'
+export { FocusableChildren } from './FocusableChildren.stories.tsx'
--- a/packages/ui/src/components/VisuallyHidden/__tests__/index.test.tsx
+++ b/packages/ui/src/components/VisuallyHidden/__tests__/index.test.tsx
@@ -1,0 +1,135 @@
+import { screen, waitFor } from '@testing-library/react'
+import { userEvent } from '@testing-library/user-event'
+import { DownloadIcon } from '@ultraviolet/icons'
+import { renderWithTheme } from '@utils/test'
+import { describe, expect, it, vi } from 'vitest'
+import { VisuallyHidden } from '..'
+
+type TestElement = HTMLElement | SVGElement
+
+/**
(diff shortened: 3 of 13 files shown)


packages/ui/src/components/VisuallyHidden/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+type VisuallyHiddenProps<T extends ElementType = 'span'> = Omit<ComponentProps<T>, 'as'> & {
+  /**
+   * The element type for the container. By default `span`.
+   * Any focusable element will be displayed when focused
+   */
+  as?: T
+}
+
+export const VisuallyHidden = <T extends ElementType = 'span'>({
```
philibea: should we handle the forwardRef for react 18 ?

packages/ui/src/components/VisuallyHidden/__stories__/index.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import type { Meta } from '@storybook/react-vite'
+import { VisuallyHidden } from '..'
+
+export default {
+  component: VisuallyHidden,
+  title: 'UI/Other/VisuallyHidden',
+  parameters: {
+    a11yStatus: {
```
philibea, line 8: by default any new component should be accessible and an audit should be done before ?

packages/ui/src/components/VisuallyHidden/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+  children,
+  className,
+  as: Component,
+  ...props
+}: VisuallyHiddenProps<T>) => {
+  const ComponentElement = (Component ?? 'span') as ElementType
```
philibea: if by default ElementType = span, should we have `as: Component,` default on span.
The idea is to avoid the casting with as ElementType if possible.

packages/ui/src/components/VisuallyHidden/__tests__/index.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+    const element = screen.getByText('hidden', { selector: 'span' })
+    expectVisuallyHidden(element)
+
+    await userEvent.tab()
+    await waitFor(() => expect(element).toHaveFocus())
+
+    expect(asFragment()).toMatchSnapshot()
+  })
+
+  it('should be visible when it is a focusable element', async () => {
```
jsulpis, line 61: I think this test is redundant with the previous one, you could choose one of them

packages/ui/src/components/VisuallyHidden/__tests__/index.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+ * - Is present in the DOM;
+ * - Doesn’t have the display CSS property set to none;
+ * - Doesn’t have the opacity CSS property set to 0;
+ * - Doesn’t have the visibility CSS property set to hidden/collapse;
+ * -Doesn’t have the aria-hidden attribute set to true;
+ *
+ * None of this is true in `VisuallyHidden`. So we check if the style is correctly applied instead.
+ *
+ * For visibility, after focus we assert the element receives focus; CSS :focus rules
+ * are not reliably reflected in jsdom computed styles, so testing focus is simpler
```
jsulpis, line 21: are you saying that when focusing an element, the visually hidden style still applies ? 

I was thinking it would be good to have a [custom matcher](https://vitest.dev/api/expect.html#expect-extend) `.toBeVisuallyHidden()` instead of `expectVisuallyHidden`, for now in this test suite but we could make it available everywhere later. I guess we could use `.not.toBeVisuallyHidden()` but I'm not sure with your comment

packages/ui/src/components/VisuallyHidden/__tests__/index.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+        <VisuallyHidden id="example_desc">desc</VisuallyHidden>
+      </>,
+    )
+
+    const button = screen.getByRole('button')
+    expect(button).toHaveAccessibleName('download')
+
+    const input = screen.getByRole('textbox')
+    expect(input).toHaveAccessibleName('label')
+    expect(input).toHaveAccessibleDescription('desc')
```
jsulpis, line 119: perfect 👌

packages/ui/src/components/VisuallyHidden/__stories__/Role.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+As.args = {
+  children: 'Hidden text',
+  as: 'div',
+}
+
+As.parameters = {
+  docs: {
+    description: {
+      story:
+        'You can change the element type using prop `as`. Do not use any interactive element, as the element is not visible.',
```
jsulpis: this comment is probably outdated since you show focusable elements for skip links in the FocusableChildren story ?

packages/ui/src/components/VisuallyHidden/__tests__/index.test.tsx
lisalupi: no, the opposite. I was saying that I consider that the selector `:not(:focus)` on the css class works correctly, making the element visible.
I can see to create a custom matcher

packages/ui/src/components/VisuallyHidden/__stories__/Role.stories.tsx
lisalupi: yes my bad, i changed the bahavior but forgot to update this comment 🤐

jsulpis (approved):

PR #6770, opened 2026-08-07
feat(Filters): label description

lisalupi:
## Type

- Feature

#### What is expected?
`Filters`: add prop `labelDescription` to all filter items (date, multiselect, search, etc.)

### DIFF
--- a/packages/ui/src/compositions/Filters/types.ts
+++ b/packages/ui/src/compositions/Filters/types.ts
@@ -1,4 +1,4 @@
-import type { ComponentProps } from 'react'
+import type { ComponentProps, ReactNode } from 'react'
 import type { DateInput } from '../../components/DateInput'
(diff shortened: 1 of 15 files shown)


packages/ui/src/compositions/Filters/types.ts
```diff
@@ -117,6 +118,7 @@ export type FilterConfigItem<V extends AnyObject = AnyObject> =
 export type FilterConfigGroup<V extends AnyObject = AnyObject> = {
   type: 'group'
   label: string
+  labelDescription?: ReactNode
```
jsulpis: in `FiltersDrawer.tsx:78` you hide the `labelDescription` if it's a config group, so I guess the `labelDescription` should not be present in the type `FilterConfigGroup` ?

jsulpis (approved):

PR #6772, opened 2026-08-10
docs(Radio): accessibility audit

lisalupi:
## Type

- Documentation

#### What is expected?
Audit `Radio` accessibility

### DIFF
--- a/packages/ui/src/components/Radio/__stories__/A11y.mdx
+++ b/packages/ui/src/components/Radio/__stories__/A11y.mdx
@@ -46,6 +46,7 @@ To apply when using the component:
 **High:**
 
 - Tooltip wrapper around the radio may interfere with screen reader announcements and keyboard navigation.
+- Hover state management is placed on the parent `div` (`radioStyle.container`), which is a non-interactive element that cannot receive focus. The component should manage all interaction states on the `input` field only, and the CSS should target those input states (e.g. `:hover`, `:focus-visible`) relative to the `svg` sibling instead of relying on hover selectors on the parent container.
 
 **Medium:**
 
--- a/packages/ui/src/components/Radio/__stories__/A11y.mdx
+++ b/packages/ui/src/components/Radio/__stories__/A11y.mdx
@@ -1,0 +1,59 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="UI/Data Entry/Radio/A11y" />
+
+# Radio - Accessibility
+
+## Description
+
+A form input component used to select a single option from a group of options. Supports checked/unchecked states, error states with helper text, disabled state, required field indication, tooltip support, and two sizes (small/medium).
+
+## Summary
+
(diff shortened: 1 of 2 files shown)


jsulpis (approved):

packages/ui/src/components/Radio/__stories__/A11y.mdx
```diff
@@ -0,0 +1,58 @@
...
+- Focus ring contrast ratio is insufficient: 1.32:1 (light theme), 1.63:1 (dark theme), 1.34:1 (darker theme). All are below the WCAG 1.4.11 minimum of 3:1. Keyboard users with low vision cannot see which element has focus. Accessibility recommendations state to never disable `outline`. Use `:focus-visible` to target only keyboard interactions without visual pollution for mouse users.
+- SVG `<title>radio</title>` exposes a non-meaningful title to screen readers. The SVG is decorative and should have `aria-hidden="true"` or remove the title element entirely.
+
+**High:**
+
+- Tooltip wrapper around the radio may interfere with screen reader announcements and keyboard navigation.
+
+**Medium:**
+
+- Tooltip wrapper around the radio may interfere with keyboard navigation and screen reader announcements depending on implementation.
```
jsulpis: I forgot, the "tooltip" prop will have to be migrated to tooltipDescription and aria-label to accessibleLabel

(same for all components, maybe we should just log that somewhere)

packages/ui/src/components/Radio/__stories__/A11y.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+# Radio - Accessibility
+
+## Description
+
+A form input component used to select a single option from a group of options. Supports checked/unchecked states, error states with helper text, disabled state, required field indication, tooltip support, and two sizes (small/medium).
+
+## Summary
+
+1. ❌ **Perceivable**: Focus indicator contrast is insufficient (1.32:1 light, 1.63:1 dark, 1.34:1 darker themes)
```
iManu, line 13: + 1 : keyboard navigation should be improved, especially in Dark mode which the focus is mainly invisible

packages/ui/src/components/Radio/__stories__/A11y.mdx
```diff
@@ -0,0 +1,59 @@
...
+2. ❌ **Operable**: Focus indicator contrast fails WCAG 1.4.11 (all themes below 3:1 minimum); tooltip wrapper may interfere with keyboard navigation
+3. ✅ **Understandable**: Labels and helper text are properly associated via htmlFor and aria-describedby
+4. ❌ **Robust**: SVG contains `<title>radio</title>` which is confusing; missing radiogroup semantics at component level
```
iManu, line 16: the svg should be aria-hidden

iManu (approved):
General note: The hover management is set on the parent div which is a non-interactive element that can not receive focus. The component should manage all the states on the input field only and the css should set the designs related to the input states, perhaps on the svg element, as it is sibling of input

PR #6783, opened 2026-08-12
feat(Tooltip): redesign using Floating UI

jsulpis:
## Type

- Enhancement

#### What is expected?

`Tooltip`: rebuilt on Floating UI instead of the `Popup`, fixing multiple positioning and accessibility bugs.
  - The tooltip can be used as a label or a description
  - The tooltip can be rendered next to its reference element or in a React Portal
  - The trigger wrapper `tabIndex` defaults to `-1` when the children is already a focusable element, removing the double tab stop
  - The redundant `aria-controls` is no longer set, and `role` is hard-pinned to `tooltip`
  - `aria-describedby` is now forwarded to render-function children and is only set while the tooltip is open
  - The tooltip is now hoverable
  - Multiple props were added or deprecated, staying backward-compatible

- rewrite the Tooltip using floating-ui
- update the accessibility audit

A few components tooltips are a bit broken in this PR, the fixes are made in the next PR.


packages/ui/src/components/Tag/__tests__/index.test.tsx
```diff
@@ -66,26 +66,27 @@ describe('tag', () => {
...
   it('works correctly with copiable', async () => {
-    const { asFragment } = renderWithTheme(<Tag copiable>test</Tag>)
+    renderWithTheme(<Tag copiable>test</Tag>)
 
     const tag = screen.getByRole('button', { name: 'test' })
 
     await userEvent.click(tag)
     expect(writeTextMock).toHaveBeenCalledExactlyOnceWith('test')
-
-    expect(asFragment()).toMatchSnapshot()
```
lisalupi, line 76: why did you remove snapshots?

packages/ui/src/components/Tooltip/__stories__/A11y.mdx
```diff
@@ -51,32 +51,89 @@ To apply when using the component:
 
 **Critical:**
 
-- Missing `aria-describedby` when using function children - The `Popup` render-prop API passes `onBlur`, `onFocus`, `onPointerEnter`, `onPointerLeave` and `ref` to the children function, but does NOT forward `aria-describedby` (or the tooltip `id`). As a result, when consumers use the function form of children, the trigger element is never associated with the tooltip, and screen readers will not announce the tooltip content on focus. This is the core requirement of the WAI-ARIA tooltip pattern.
+- ✅ Missing `aria-describedby` when using function children
```
lisalupi, line 55: why not remove the "accessibility issues"? Or maybe rename the section

packages/ui/src/components/Tooltip/styles.css.ts
```diff
@@ -1,6 +1,98 @@
...
 
 const tooltip = style({
-  pointerEvents: 'none',
+  backgroundColor: theme.colors.neutral.backgroundStronger,
+  borderRadius: theme.radii.default,
+  color: theme.colors.neutral.textStronger,
+  fontSize: theme.typography.caption.fontSize,
+  lineHeight: theme.typography.caption.lineHeight,
+  overflowWrap: 'break-word',
+  padding: `${theme.space['0.5']} ${theme.space[1]}`,
```
lisalupi, line 17: padding should be 8px (0.5rem)
Image

packages/ui/src/components/Tooltip/__tests__/a11y.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { consoleThemesMap } from '@ultraviolet/themes'
+import { renderWithTheme, expectNoViolations } from '@utils/test'
+import { describe, expect, it } from 'vitest'
+import { Tooltip } from '..'
+import { TextInput } from '../../TextInput'
+
+describe('tooltip - A11y', { tags: ['a11y'] }, () => {
+  describe('axe checks', () => {
+    it.for([...consoleThemesMap.entries()])(
+      'should not have violations when visible (theme: %s)',
```
lisalupi, line 13: are we sure that the contrast checker works?

packages/ui/src/components/Tooltip/__tests__/a11y.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+        await expectNoViolations(container)
+      },
+    )
+  })
+
+  describe('interactions', () => {
+    it('should open with focus (Tab) and close with Escape', async () => {
+      renderWithTheme(
+        <Tooltip delay={{ open: 0 }} maxWidth={100} text="test success!">
+          <p data-testid="children">Hover me</p>
```
lisalupi: why add a data-testid here?

packages/ui/src/components/Tooltip/useTooltip.ts
```diff
@@ -0,0 +1,96 @@
...
+  useHover,
+  safePolygon,
+  useFocus,
+  useDismiss,
+  useInteractions,
+} from '@floating-ui/react'
+import { usePrefersReducedMotion } from '@ultraviolet/animations'
+import { useState, useRef, useCallback } from 'react'
+import type { CSSProperties } from 'react'
+import type { TooltipPlacement } from '.'
```
lisalupi: I'd put it in a type.ts file to avoid a circular dependency, even if it's just type

packages/ui/src/components/Tooltip/TooltipChildren.tsx
```diff
@@ -0,0 +1,99 @@
...
+  :not([tabindex="-1"])
+`.replace(/\s+/g, '')
+
+type TooltipChildrenProps = {
+  tooltip: ReturnType<typeof useTooltip>
+  tooltipId: string
+  isTooltipVisible: boolean
+} & Pick<TooltipProps, 'children' | 'tabIndex' | 'relation' | 'containerFullHeight' | 'containerFullWidth'>
+
+export function TooltipChildren({
```
lisalupi:
```suggestion
export const TooltipChildren = ({
```

packages/ui/src/components/Tooltip/styles.css.ts
```diff
@@ -1,6 +1,98 @@
-import { style } from '@vanilla-extract/css'
+import { theme } from '@ultraviolet/themes'
+import { style, styleVariants } from '@vanilla-extract/css'
+import { recipe } from '@vanilla-extract/recipes'
+import { fadeIn, fadeOut } from '../../utils'
+import { tagStyle } from '../Tag/styles.css'
+import { tagListStyle } from '../TagList/styles.css'
+import { textStyle } from '../Text/style.css'
+
```
lisalupi: you can remove those import from `Popup` style I think since the tooltip doesn't use `Popup` anymore and the components use `Tooltip`

packages/ui/src/components/Tooltip/index.tsx
```diff
@@ -1,82 +1,193 @@
...
-      style={style}
-      tabIndex={tabIndex}
-      text={text}
-      visible={visible}
-    >
-      {children}
-    </Popup>
-  ),
-)
+export const Tooltip = ({
```
lisalupi, line 118: no more ref & forwardRef ?

packages/ui/src/components/Tooltip/__tests__/index.test.tsx
```diff
@@ -1,135 +1,125 @@
...
+      const { asFragment } = renderWithTheme(
+        <Tooltip text="test">
+          <div className="some complex component">
+            <button type="button">interactive element</button>
+          </div>
+        </Tooltip>,
+        consoleLightTheme,
+      )
+      expect(asFragment()).toMatchSnapshot()
+    })
```
lisalupi, line 33: you don't really test the tabIndex here, don't you? 🤔 Do you just rely on the snapshot?

packages/ui/src/components/Tag/__tests__/index.test.tsx
jsulpis: I had an issue with the snapshots which did not render the same markup in local and on the preview and I couldn't find a reliable fix. Since the test has other assertions I got rid of the snapshot

packages/ui/src/components/Tooltip/__stories__/A11y.mdx
jsulpis: Alexandre said that it could be good to keep track of what we did to improve the accessibility. But I agree that it looks a bit verbose. Maybe we could hide the whole section in a detail/summary ?

packages/ui/src/components/Tooltip/styles.css.ts
jsulpis: yes that's what I have
image

packages/ui/src/components/Tooltip/__tests__/a11y.test.tsx
jsulpis: absolutely not 😬 but I used the right color tokens.

I would say we can add those tests as if they worked and when we [switch to a browser mode ](https://github.com/scaleway/ultraviolet/pull/6686) we will see if something breaks

packages/ui/src/components/Tooltip/__tests__/a11y.test.tsx
jsulpis: probably copied from another old test from the other test file, I removed it

packages/ui/src/components/Tooltip/__stories__/A11y.mdx
lisalupi: yes because it is a lot of "unnecessary" info, since it's just the component's history

packages/ui/src/components/Tooltip/useTooltip.ts
jsulpis: ✅

packages/ui/src/components/Tooltip/styles.css.ts
lisalupi: on your screenshot i see a 4px top/bottom padding

packages/ui/src/components/Tooltip/styles.css.ts
jsulpis: the `minWidth` was necessary to preserve the text ellipsis on the children (tags), but I replaced all the styles with a `maxWidth: "100%"` and it seems to work

packages/ui/src/components/Tooltip/index.tsx
jsulpis: I removed the two refs, `innerRef` and `tooltipRef`, they didn't work anyway so I guess they were not used. And it was more complicated to handle with the Floating UI hooks

packages/ui/src/components/Tooltip/__tests__/index.test.tsx
jsulpis: yes for once I relied on the snapshot, but you're right I added a check to be more explicit

packages/ui/src/components/Tooltip/styles.css.ts
jsulpis: oh sorry then where do you see this padding ? On Figma I have 4px/8px
image

packages/ui/src/components/Tooltip/TooltipChildren.tsx
jsulpis: ✅

packages/ui/src/components/Tooltip/styles.css.ts
lisalupi: https://ultraviolet.scaleway.com/6dd9b5c45/p/5919b1-tooltip/b/221400/i/298132406 (first illustration) when you click on "open in figma" but i just saw that elsewhere it is indeed 0.5/1

packages/ui/src/components/Tooltip/styles.css.ts
lisalupi: No sorry what I meant was that in `packages/ui/src/components/Popup/styles.css.ts` there is still
```js
import { tagStyle } from '../Tag/styles.css'
import { tagListStyle } from '../TagList/styles.css'
import { textStyle } from '../Text/style.css'
```
which is probably not necessary anymore since Tooltip doesn't use Popup anymore and this style was added because of the tooltip behavior on those components

jsulpis: oh right, fixed

packages/ui/src/components/Tooltip/styles.css.ts
jsulpis: As seen with Mathieu, this was an old design on Zeroheight, the right padding is 0.5/1

lisalupi (approved):

PR #6785, opened 2026-08-12
fix: update components with new Tooltip behavior

jsulpis:
## Type

- Enhancement

### Summarize concisely:

Fix the issues brought by the new Tooltip

#### What is expected?

No issue with our components using tooltips

Fix tooltip issues (position / tabindex) in our components

| url  | image | image |


packages/ui/src/components/NumberInput/components/Controls.tsx
```diff
@@ -63,6 +63,7 @@ export const Controls = ({ controls, direction, size, localRef, max, min, isDisa
         sentiment="neutral"
         size={size === 'small' ? 'xsmall' : 'small'}
         variant="ghost"
+        tabIndex={-1}
```
lisalupi, line 66: you can update the a11y audit talking about that 😄

packages/ui/src/components/Slider/styles.css.ts
```diff
@@ -116,34 +116,6 @@ const thumbStyle = recipe({
...
-        selectors: {
-          '&::-webkit-slider-thumb': {
-            top: `-${THUMB_SIZE / 2}px`,
-          },
-        },
-      },
-    },
-    isDouble: {
-      false: {
-        selectors: {
```
lisalupi, line 130: it broke something on firefox 

Image

Image


Image

packages/ui/src/components/TagInput/index.tsx
```diff
@@ -188,78 +188,97 @@ export const TagInput = ({
...
-                  value={input}
-                />
-              )}
-            </div>
-            {computedClearable || success || error ? (
-              <div className={tagInputStyle.stateContainer}>
-                {computedClearable ? (
-                  <Button
-                    aria-label="clear value"
+          {tooltipProps => (
```
lisalupi: Image

not sure about tooltip placement here

lisalupi (changes requested):

packages/ui/src/components/NumberInput/components/Controls.tsx
jsulpis: ✅

packages/ui/src/components/TagInput/index.tsx
jsulpis: I reverted the changes to keep the tooltip on the container element, which may be wrong but we will fix that when auditing the component

packages/ui/src/components/Slider/styles.css.ts
jsulpis: it looks like the slider thumb can't be positioned in Firefox, so I removed even more styles and now it works

image

lisalupi (approved):

PR #6787, opened 2026-08-13
chore: optimize icons generation

jsulpis:
## Type

- Enhancement

#### What is expected?

Improve the script that generates the icons so that we don't have to manually add or fix the classNames in the svg files. This allows to have a more aggressive svgo optimization since we don't rely on the className anymore.

- automatically add fill classNames on {Product/Category}Icons during generation based on the fill color
- log errors if there is an invalid fill color in an input svg
- update svgo config to keep only what is really necessary
- run the import script

| Icon file  | 287kB | 205kB (-28%) |
| ProductIcon file  | 457kB | 406 kB (-11%) |
| Logo file | 471kB | 305 kB (-35%) |
| Flag file | 53.7kB | 40 kB (-25%) |
| CategoryIcon file | 43.8kB | 37.8 kB (-13%) |

The only visual difference I saw is on the integration-service CategoryIcon which was broken in prod

| Before |   After |      Figma |
| image  | image |  image


lisalupi (changes requested):
Image breaks EyeIcon

lisalupi (approved):

PR #6798, opened 2026-08-17
feat(Flags): udpate sizes

lisalupi:
## Type

- Enhancement

#### What is expected?
Update flags size.

Add small variant to flags svg and update script


jsulpis (commented):
Sorry it's a bit late but if the small flags just have a different padding I think it would be better to set it in CSS instead of doubling the number of assets. We already ship a huge amount of javascript code that just renders svg, it's not ideal

jsulpis (changes requested):
the flags are not correctly centered in their box, the issue comes from the small icons in Figma (the large icons are correct)

Image

Image

jsulpis (approved):

PR #6799, opened 2026-08-17
chore: correctly group all @testing-library packages

jsulpis:
## Type

- Bug

### Summarize concisely:

We had two rules for @testing-library packages:
- one that extends `packages:jsUnitTest` and match all `devDependencies` of test libs
- on that enforced a `devDeps` scope on the testing-library packages, which are used as `dependencies` in the package `utils/test`

Since the testing-library packages are declared as `dependencies` in the package `utils/test`, they didn't match the first rule in the config, which triggers 2 distinct PRs for testing-library (see [one here](https://github.com/scaleway/ultraviolet/pull/6796)). These PRs always fail the CI because there is a mismatch in libs versions in the monorepo.

#### What is expected?

All testing-library updates are grouped in the same PR

- extend the first rule to include the `dependencies` and force a `devDeps` scope to include the testing-library from utils/test
- remove the second rule

This change assumes that no other test lib is a `dependency`, or that we accept that all test libs are marked as `devDeps`

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -16,8 +16,6 @@
     "**/__fixtures__/**"
   ],
   "prConcurrentLimit": 5,
-  "reviewers": ["team:console"],
-  "reviewersSampleSize": 2,
   "labels": ["dependencies"],
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -16,8 +16,6 @@
     "**/__fixtures__/**"
   ],
   "prConcurrentLimit": 5,
-  "reviewers": ["team:console"],
```
jsulpis, line 19: I think we can spare them all those PRs ?

lisalupi (approved):

PR #6803, opened 2026-08-18
docs(CheckboxGroup): audit accessibility

lisalupi:
## Type

- Documentation

#### What is expected?
Audit `CheckboxGroup` accessibility

### DIFF
--- a/packages/ui/src/components/CheckboxGroup/__stories__/A11y.mdx
+++ b/packages/ui/src/components/CheckboxGroup/__stories__/A11y.mdx
@@ -1,0 +1,50 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="UI/Data Entry/CheckboxGroup/A11y" />
+
+# CheckboxGroup - Accessibility
+
+## Description
+
+A form component that groups a set of checkboxes inside a native `<fieldset>` with an optional `<legend>`, helper/error text, and required indication. Each option is rendered through the `CheckboxGroup.Checkbox` sub-component which wraps the `Checkbox` component.
+
+## Summary
+
+1. ❌ **Perceivable**: Grouping relationship is provided natively via `fieldset`/`legend` (1.3.1), but the group inherits the Checkbox focus indicator contrast failure (1.4.11)
+2. ❌ **Operable**: Inherits Checkbox focus-visible contrast failure (2.4.7); independent checkboxes require no arrow-key navigation
+3. ❌ **Understandable**: Group error/helper text is associated only with the `<fieldset>`, not with the individual checkbox inputs, so options announce "invalid" without the descriptive message (3.3.1)
+4. ❌ **Robust**: A `fieldset` without `legend` has no accessible name (4.1.2)
+
+## ARIA Pattern
+
+[ARIA Checkbox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/) - individual options. For grouping, the native HTML `fieldset`/`legend` elements are the recommended mechanism and already used here.
+
+## Rules
+
+Enforced by the component:
+
+- ✅ [WCAG 1.3.1 - Info and Relationships (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) - The group is wrapped in a native `<fieldset>` with an optional `<legend>` (rendered via `Label as="legend"`), programmatically conveying that the checkboxes belong together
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/CheckboxGroup/__stories__/A11y.mdx
```diff
@@ -0,0 +1,50 @@
...
+# CheckboxGroup - Accessibility
+
+## Description
+
+A form component that groups a set of checkboxes inside a native `<fieldset>` with an optional `<legend>`, helper/error text, and required indication. Each option is rendered through the `CheckboxGroup.Checkbox` sub-component which wraps the `Checkbox` component.
+
+## Summary
+
+1. ❌ **Perceivable**: Grouping relationship is provided natively via `fieldset`/`legend` (1.3.1), but the group inherits the Checkbox focus indicator contrast failure (1.4.11)
+2. ❌ **Operable**: Inherits Checkbox focus-visible contrast failure (2.4.7); independent checkboxes require no arrow-key navigation
```
jsulpis: > independent checkboxes require no arrow-key navigation

what does that mean ? is it good or bad ?

packages/ui/src/components/CheckboxGroup/__stories__/A11y.mdx
```diff
@@ -0,0 +1,50 @@
...
+- ❌ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) - A `<fieldset>` rendered without a `legend` has no accessible name; the user-supplied `aria-describedby` value is also reused as the Description id, which can orphan the helper/error text
+
+To apply when using the component:
+
+- [WCAG 1.3.1 - Info and Relationships (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) - Always provide a `legend` so the group has an accessible name; a bare `<fieldset>` without a legend conveys no grouping information to assistive technologies
+- ⚠️ [RFC #6585 - Accessible Icons and Hidden Labels](https://github.com/scaleway/ultraviolet/discussions/6585) - When the `legend` is omitted but a visually-hidden name is still needed, provide an `accessibleLegend` prop that renders a `<legend>` hidden via the `VisuallyHidden` component (or `sr-only` class), so the `<fieldset>` keeps an accessible name without visible text
```
jsulpis: Could we use `accessibleLabel` like the other components for consistency ? It will also be easier with our types

packages/ui/src/components/CheckboxGroup/__stories__/A11y.mdx
```diff
@@ -0,0 +1,50 @@
...
+## Accessibility issues
+
+**High:**
+
+- Group error text is associated only with the `<fieldset>` via `aria-describedby`, while each `CheckboxGroup.Checkbox` input gets only `aria-invalid`/styling. Screen reader users tabbing through options hear "invalid" without the message. Fix: associate the group error/helper `id` with each checkbox input via `aria-describedby` (the underlying `Checkbox` already supports it), keeping the fieldset association as a complement.
+- `legend` is optional, but a `<fieldset>` without a `<legend>` has no accessible name. When `legend` is omitted the group still renders a `<fieldset>` (index.tsx:64) with no name, failing to convey the grouping relationship required by 1.3.1 / 4.1.2. Add an `accessibleLegend` prop that renders a hidden `<legend>` via the `VisuallyHidden` component (per RFC #6585) to name the `fieldset`.
+
+**Medium:**
+
+- The `aria-describedby` prop is overloaded: when a consumer supplies it, it is used both as the `fieldset`'s `aria-describedby` and as the `Description` element `id`. This can collide with the id of the external element it points to (duplicate id in the document).
```
jsulpis, line 46: didn't we have this issue with other components ? They should all behave the same.

I think we should always generate the id of the `Description` element and if an `aria-describedby` id is provided, use both ids in the `aria-describedby` attribute on the `input`, separated by a space.

It should not happen often that a user provides `aria-describedby` _and_ a helper / error message, but with that we could handle the case properly

jsulpis (approved):

packages/ui/src/components/CheckboxGroup/__stories__/A11y.mdx
lisalupi: i added this because radios in groups need arrow-key navigation but not checkboxes. this is a neutral statement, but not really necessary. I can remove it

packages/ui/src/components/CheckboxGroup/__stories__/A11y.mdx
lisalupi: ah ok if you prefer

iManu (approved):
Note that Storybook CheckboxGroup check (using space key or click) seems not working anymore. Btw, unlike the radio elem, the focus "outline" style is great

PR #6811, opened 2026-08-20
fix(Navigation): fix CSS specificity issues

jsulpis:
### Summary

Fix CSS specificity issues in `Navigation` that were breaking styles on the separators and collapsed items. The styles now use doubled `&&` selectors to ensure they win over other rules with equal/higher specificity.

### Screenshots

This was visible on the console, not on Storybook

|   Before   |      After |
| :--------: | ---------: |
| image | image |

### DIFF
--- a/packages/ui/src/compositions/Navigation/components/items.css.ts
+++ b/packages/ui/src/compositions/Navigation/components/items.css.ts
@@ -11,9 +11,13 @@ export const itemRelative = style({ position: 'relative' })
 export const itemPadded = style({ paddingLeft: theme.space[1] })
 
 export const itemCollapsed = style({
-  height: theme.sizing['400'],
-  width: theme.sizing['400'],
-  padding: 0,
+  selectors: {
+    '&&': {
+      height: theme.sizing['400'],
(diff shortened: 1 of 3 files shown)


packages/ui/src/compositions/Navigation/components/items.css.ts
```diff
@@ -11,9 +11,13 @@ export const itemRelative = style({ position: 'relative' })
 export const itemPadded = style({ paddingLeft: theme.space[1] })
 
 export const itemCollapsed = style({
-  height: theme.sizing['400'],
-  width: theme.sizing['400'],
-  padding: 0,
+  selectors: {
+    ['&&']: {
```
lisalupi:
```suggestion
    '&&': {
```
i don't think that the [ ] are necessary

lisalupi (approved):

packages/ui/src/compositions/Navigation/components/items.css.ts
jsulpis: ✅

PR #6815, opened 2026-08-21
fix(Icon): improve accessibility

lisalupi:
## Type

- Enhancement
- Documentation

#### What is expected?

All components (`Icon`, `ProductIcon`, `CategoryIcon`, `Flag` and `Logo`):
-  Remove default `title`
-  New prop `accessibleLabel`;
- `aria-hidden = true` by default;
- Informative icons: when an icon has an `accessibleLabel` (or `aria-label`) set, `aria-hidden` switches to `false` and the svg has `role="img"`;
- `Icon`: deprecate `aria-label` prop. Use `accessibleLabel` instead.

Why 600+ files changed? Regenerated all icon 500 components (to remove generated title) + updated all tests. 

Updated some stories showcasing icon-only component to document `VisuallyHidden` usage


packages/icons/src/components/Icon/Icon.tsx
```diff
@@ -43,30 +47,35 @@ export const Icon = forwardRef<SVGSVGElement, IconProps>(
...
     return (
       <svg
-        aria-hidden={ariaHidden}
+        aria-hidden={ariaHidden ?? !informative}
         aria-label={ariaLabel}
+        aria-labelledby={ariaLabelledBy}
         className={cn(className, icon({ disabled, prominence, sentiment, size }))}
         cursor={cursor}
         data-testid={dataTestId}
+        focusable={false}
```
jsulpis: it looks like this attribute was necessary on old browsers like Internet Explorer, do we still need it now ?

packages/icons/src/components/Icon/Icon.tsx
```diff
@@ -43,30 +47,35 @@ export const Icon = forwardRef<SVGSVGElement, IconProps>(
...
         ref={ref}
+        role={informative ? 'img' : undefined}
         stroke={stroke}
         strokeWidth={strokeWidth}
         style={style}
         viewBox={typeof size === 'string' && ['xsmall', 'small'].includes(size) ? '0 0 16 16' : '0 0 20 20'}
         width={defaultHW}
       >
-        <title>{title}</title>
+        {accessibleLabel || ariaLabel ? <title>{accessibleLabel ?? ariaLabel}</title> : null}
```
jsulpis: for consistency with the other conditions
```suggestion
        {accessibleLabel || ariaLabel ? <title>{accessibleLabel || ariaLabel}</title> : null}
```

utils/scripts/generate-icons-file.ts
```diff
@@ -74,13 +74,13 @@ const templateIcon = (iconName: string, svg: string, svgSmall?: string, svgDisab
     svgSmall
       ? `export const ${iconName} = ({
     ...props
-  }: Omit<IconProps, 'children' | 'title'>) => (
-      <Icon {...props} title="${iconName}">{typeof props.size === 'string' && ['medium', 'large', 'xlarge', 'xxlarge'].includes(props.size) ? ${wrapSvg(svg)} : ${wrapSvg(svgSmall)}}</Icon>
+  }: Omit<IconProps, 'children'>) => (
```
jsulpis: I don't know if there is a performance overhead but it's not useful to destructure the props here (and in the other components)

packages/icons/src/components/Icon/__stories__/Label.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+        from
+        <Text as="span" variant="code">
+          @ultraviolet/ui
+        </Text>
+        and
+        <Text as="span" variant="code">
+          aria-labelledBy
+        </Text>
+        : <VisuallyHidden id="example-1">See</VisuallyHidden>
+        <EyeIcon {...args} aria-labelledby="example-1" />
```
jsulpis, line 23: the svg has aria-hidden="true" so no label

Image

packages/icons/src/components/Icon/__stories__/Label.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+        Using
+        <Text as="span" variant="code">
+          Tooltip
+        </Text>
+        from
+        <Text as="span" variant="code">
+          @ultraviolet/ui
+        </Text>
+        :
+        <Tooltip relation="label" text="See">
```
jsulpis: I think we should not encourage this usage since it should be a popover instead of the tooltip, so either use a popover if it works correctly, or remove this line

jsulpis (approved):

packages/icons/src/components/Icon/Icon.tsx
iManu: I think no, while some old browsers were incorrectly treated SVG as tab stops in keyboard navigation, we should not continue to fix this bug

iManu (approved):

philibea (approved):

PR #6822, opened 2026-08-24
fix(Text): correctly recompute the overflow for the Tooltip when the text is resized

jsulpis:
### Summary

Fix the computation of the overflow in the Text component.
- use a `ResizeObserver` instead of the `resize` event listener on the window because:
  - the window could be resized without the Text to be resized (unnecessary computation)
  - the Text could be resized without the window to be resized (visual bug)
- register the `ResizeObserver` _only_ if the Text is `oneLine` since we don't need it otherwise

| resize handlers in the `Text / Showcase` story  | image | image |
| performance during resize  | image | image|

### DIFF
--- a/packages/ui/src/components/Text/index.tsx
+++ b/packages/ui/src/components/Text/index.tsx
@@ -3,7 +3,6 @@
 import type { TextVariant } from '@ultraviolet/themes'
 import { cn } from '@ultraviolet/utils'
 import { assignInlineVars } from '@vanilla-extract/dynamic'
-import { useRef } from 'react'
 import type { CSSProperties, AriaRole, ElementType, ReactNode } from 'react'
 import recursivelyGetChildrenString from '../../helpers/recursivelyGetChildrenString'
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/Text/index.tsx
```diff
@@ -65,13 +64,12 @@ export const Text = ({
   'aria-hidden': ariaHidden,
   style,
 }: TextProps) => {
-  const elementRef = useRef(null)
-  const isOverflowing = useIsOverflowing(elementRef)
+  const [setElement, isOverflowing] = useIsOverflowing({ enabled: oneLine === true })
```
jsulpis, line 67: I used a React state (`setElement`) instead of a ref because when the Tooltip re-renders it can destroy and recreate the children, which makes the `ref` stale. With `setElement` it will trigger the `useEffect` in `useIsOverflowing` to recreate the `ResizeObserver` on the new element.

It's not very pretty but this hook is not publicly exposed so we can always update this later if needed.

lisalupi (approved):

PR #6824, opened 2026-08-24
fix(TagList): fix and improve overflow behavior

jsulpis:
### Summary

Redesign the overflow in the `TagList` to fix visual bugs and improve the behavior 

| `TagList/Parent With Defined Width`  | image | image |
| `TagList/Parent With Defined Width`, multiline=true  | image | image |

### DIFF
--- a/packages/ui/src/components/TagList/__stories__/ParentWithWidth.stories.tsx
+++ b/packages/ui/src/components/TagList/__stories__/ParentWithWidth.stories.tsx
@@ -1,20 +1,16 @@
 import type { StoryFn } from '@storybook/react-vite'
-import { useState } from 'react'
 import { TagList } from '..'
-import { Slider } from '../../Slider'
 import { Stack } from '../../Stack'
 
 export const ParentWithDefinedWidth: StoryFn<typeof TagList> = args => {
-  const [width, setWidth] = useState(298)
-
   return (
     <Stack gap={2}>
-      <Slider label="Change container width" max={500} min={2} onChange={setWidth} value={width} />
       <div
         style={{
           border: '1px solid gray',
           padding: '10px',
-          width: `${width}px`,
+          overflow: 'hidden',
+          resize: 'horizontal',
         }}
       >
         <TagList {...args} />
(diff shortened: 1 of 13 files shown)


packages/ui/src/components/TagList/__stories__/ParentWithWidth.stories.tsx
```diff
@@ -1,20 +1,16 @@
...
 import { TagList } from '..'
-import { Slider } from '../../Slider'
 import { Stack } from '../../Stack'
 
 export const ParentWithDefinedWidth: StoryFn<typeof TagList> = args => {
-  const [width, setWidth] = useState(298)
-
   return (
     <Stack gap={2}>
-      <Slider label="Change container width" max={500} min={2} onChange={setWidth} value={width} />
```
jsulpis, line 12: the overflow update only worked in the story because this Slider caused the TagList to re-render. But nothing listened to changes of the actual component width, so the TagList could be broken by a resize without re-render.

The new `resize: horizontal` in CSS assures that there is no external re-render.

packages/ui/src/components/TagList/index.tsx
```diff
@@ -88,261 +87,163 @@ export const TagList = ({
...
-  // after the first render ( to know if we should add ellipsis to the last visible tag
-  // or readjust the tags when joined with the popover trigger might overflow the parent )
-  // and this causes some flickering
-  const [isReady, setIsReady] = useState(false)
+  const popoverTriggerRef = useRef<HTMLButtonElement>(null)
 
   const [isPopoverVisible, setIsPopoverVisible] = useState(false)
-  const [popoverTriggerWidth, setPopoverTriggerWidth] = useState(0)
-  const [visibleTags, setVisibleTags] = useState<TagType[]>([])
+  const [visibleTags, setVisibleTags] = useState<TagType[]>([tags[0]])
```
jsulpis, line 93: we initially render the first tag which will end-up in the DOM in any case, to avoid a vertical layout shift that we had before (clearly visible in the story `ParentWithDefinedWidth` with CPU slowdown)

lisalupi (approved):

PR #6833, opened 2026-08-26
fix(TagList): empty tags

lisalupi:
## Type

- Bug

### Summarize concisely:

`TagList`: avoid adding empty tags when tags.length = 0

| Page |   After   |      Before |
| TagList, tags=[]  |Capture d’écran 2026-08-26 à 16 10 07 (no empty tag) | Capture d’écran 2026-08-26 à 16 10 34 (empty tag)|

### DIFF
--- a/packages/ui/src/components/TagList/index.tsx
+++ b/packages/ui/src/components/TagList/index.tsx
@@ -185,11 +185,6 @@ export const TagList = ({
       sentiment: sentiment,
     }
 
-    // Avoid displaying an empty tag
-    if (!tag) {
-      return
-    }
-
     if (isKeyValueTag(tag)) {
       return <Tag keyValue={tag} {...commonProps} key={`${tagLabel}-${index}`} />
     }
--- a/packages/ui/src/components/TagList/index.tsx
+++ b/packages/ui/src/components/TagList/index.tsx
@@ -90,7 +90,8 @@ export const TagList = ({
   const popoverTriggerRef = useRef<HTMLButtonElement>(null)
 
   const [isPopoverVisible, setIsPopoverVisible] = useState(false)
-  const [visibleTags, setVisibleTags] = useState<TagType[]>([tags[0]])
+  // Avoid adding undefined to default visibleTags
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/TagList/index.tsx
```diff
@@ -90,7 +90,8 @@ export const TagList = ({
   const popoverTriggerRef = useRef<HTMLButtonElement>(null)
 
   const [isPopoverVisible, setIsPopoverVisible] = useState(false)
-  const [visibleTags, setVisibleTags] = useState<TagType[]>([tags[0]])
+  // Avoid adding undefined to default visibleTags
+  const [visibleTags, setVisibleTags] = useState<TagType[]>(tags.length > 0 ? [tags[0]] : [])
```
jsulpis: I think this is enough to not render any tag when the list is empty no ? the `visibleTags.map` should return nothing if the list is empty, so no need for the condition in `renderTag`

this change is because your version would create a list with one undefined element if the tags is [undefined]
```suggestion
  const [visibleTags, setVisibleTags] = useState<TagType[]>(tags[0] ? [tags[0]] : [])
```

lisalupi: I wanted to add it to make sure there is no issue, because with [undefined] it did iterate on the array and it returned the empty tag

jsulpis: yes because `[undefined].length` is 1 so it iterates, but with the change I suggested it should be `[]` instead of `[undefined]`, and `[].length` is 0 so it should not iterate. You can just try without the `if` in `renderTag` to see if the test still passes

jsulpis (approved):

packages/ui/src/components/TagList/index.tsx
lisalupi: yes yes I know but i think that leaving the check in renderTag is another protection. But I can remove it if you prefer

jsulpis: I prefer to avoid unnecessary code but that's no big deal

PR #6837, opened 2026-08-27
fix(SelectInput): isLoading vs empty state

lisalupi:
### Concisely:
`SelectInput`: empty state should no longer override `isLoading`.

### Behavior change
When `isLoading` is set, the dropdown now renders a loading skeleton as the sole content. For **ungrouped** options this means `AddOption` / `SelectAll` are no longer shown above the skeleton while loading (previously they were rendered alongside it). This makes ungrouped consistent with grouped, which already showed only the skeleton during loading.

## Type

- Bug

### DIFF
--- a/packages/ui/src/components/SelectInput/components/Dropdown/Content.tsx
+++ b/packages/ui/src/components/SelectInput/components/Dropdown/Content.tsx
@@ -110,6 +110,19 @@ export const CreateDropdown = ({
     </Stack>
   )
 
+  if (isLoading) {
+    return (
+      <Stack
+        className={selectInputStyle.dropdownContainer}
+        id="select-dropdown"
+        onKeyDown={handleKeyDownSelect}
+        role="listbox"
+      >
+        <Skeleton variant="block" />
+      </Stack>
+    )
+  }
+
   if (isEmpty && !addOption) {
     return (
       <Stack alignItems="center" className={selectInputStyle.dropdownEmptyState} gap={2}>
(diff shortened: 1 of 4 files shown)


etienne-scaleway (approved):

philibea (approved):

packages/ui/src/components/SelectInput/components/Dropdown/Content.tsx
```diff
@@ -110,6 +110,19 @@ export const CreateDropdown = ({
     </Stack>
   )
 
+  if (isLoading) {
```
jsulpis, line 113: Behavior change: this early return now renders only `<Skeleton>`, so during loading the **ungrouped** case no longer shows `AddOption`/`SelectAll` (previously they rendered above the skeleton in the ungrouped branch). This makes ungrouped consistent with grouped, but it's still a visible change — can you confirm it's intended and mention it in the PR body?

packages/ui/src/components/SelectInput/components/Dropdown/Content.tsx
```diff
@@ -110,6 +110,19 @@ export const CreateDropdown = ({
...
+    return (
+      <Stack
```
jsulpis, line 115: `data-grouped` is now dropped during loading: the grouped container still sets `data-grouped`, but this shared loading block doesn't. The snapshot confirms `data-grouped="true"` is removed from the grouped loading case. Anything relying on that attribute while loading?

packages/ui/src/components/SelectInput/components/Dropdown/Content.tsx
```diff
@@ -110,6 +110,19 @@ export const CreateDropdown = ({
...
+        className={selectInputStyle.dropdownContainer}
+        id="select-dropdown"
+        onKeyDown={handleKeyDownSelect}
+        role="listbox"
+      >
+        <Skeleton variant="block" />
```
jsulpis, line 121: The listbox container has `role="listbox"` but no `aria-busy`/`aria-live` here. Previously the skeleton wrapper carried `aria-busy="true"` and `aria-live="polite"`. An empty listbox that will be populated later should probably stay `aria-busy` so screen readers don't announce an empty list.

packages/ui/src/components/SelectInput/components/Dropdown/Content.tsx
```diff
@@ -183,73 +192,67 @@ export const CreateDropdown = ({
...
-            option={{
-              label: addOptionLabel,
-              searchText: searchInput,
-              value: `${addOption?.text} ${searchInput}`,
-            }}
-            searchable={searchable}
-            textVariant={textVariant}
-          />
-          <SelectAll textVariant={textVariant} />
+      <AddOption
```
jsulpis, line 195: Nit: the listbox `Stack` wrapper (this block vs the loading early return at ~115–122 and the ungrouped block at ~152–158) and the `AddOption`/`SelectAll` markup are duplicated across the three branches. Extracting the shared wrapper/items into a variable or component would reduce drift risk — the snapshot churn here is largely a symptom of that duplication.

jsulpis (commented):
Missing regression test: both existing loading tests use **non-empty** data (`dataGrouped`, `dataUnGrouped`). This PR fixes "empty state overriding isLoading", but that exact scenario (`isLoading` + empty `options={[]}`) isn't covered — please add a test asserting the skeleton renders (not "No options") when `isLoading` with empty data.

jsulpis (approved):

PR #6852, opened 2026-08-31
fix(deploy): min scale preview to 1, prod to 2

philibea:
## Type

- Enhancement

#### What is expected?

Fix output url for custom domain.
Set min scale to 1.

### DIFF
--- a/.github/workflows/cleanup_stale_containers.yml
+++ b/.github/workflows/cleanup_stale_containers.yml
@@ -1,0 +1,71 @@
+name: Cleanup Stale Containers
+
+# Deletes Scaleway Serverless Containers that were not torn down properly
+# (e.g. a PR was closed but the teardown workflow was cancelled or failed).
+#
+# Uses the native `type: cleanup` mode of scw-serverless-container-action,
+# which lists containers in the namespace and deletes the ones matching the
+# filters below. The `cleanup_name_pattern` regex targets PR preview
+# containers (ending with `-storybook`) so `main` and `beta` are never touched.
+
+on:
+  schedule:
+    # Every day at 03:00 UTC
+    - cron: '0 3 * * *'
+
+  # uncommented to force cleanup on merge request
+  # pull_request:
+
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        description: 'List stuck containers without deleting them'
+        required: false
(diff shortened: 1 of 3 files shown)


.github/workflows/cleanup_stale_containers.yml
```diff
@@ -0,0 +1,71 @@
...
+# filters below. The `cleanup_name_pattern` regex targets PR preview
+# containers (ending with `-storybook`) so `main` and `beta` are never touched.
+
+on:
+  schedule:
+    # Every day at 03:00 UTC
+    - cron: '0 3 * * *'
+
+  # uncommented to force cleanup on merge request
+  # pull_request:
```
jsulpis: why ?

.github/workflows/cleanup_stale_containers.yml
```diff
@@ -0,0 +1,71 @@
+name: Cleanup Stale Containers
+
+# Deletes Scaleway Serverless Containers that were not torn down properly
+# (e.g. a PR was closed but the teardown workflow was cancelled or failed).
+#
+# Uses the native `type: cleanup` mode of scw-serverless-container-action,
+# which lists containers in the namespace and deletes the ones matching the
+# filters below. The `cleanup_name_pattern` regex targets PR preview
```
jsulpis: there is a `cleanup_exclude_names` below but no `cleanup_name_pattern`, is it two different parameters ? Since we can craft a regex to exclude main and beta, I think `cleanup_name_pattern` alone would be enough

jsulpis (commented):
I'm not sure the summary works because [this workflow](https://github.com/scaleway/ultraviolet/actions/runs/33391603442) deleted many containers and the summary is empty (and with empty quotes)

Image

.github/workflows/cleanup_stale_containers.yml
philibea: in my first iteration i made a `cleanup_name_pattern` but i found that it's more simple to have a whitelist.

.github/workflows/cleanup_stale_containers.yml
philibea: just to check that everything work as expect. We can remove it now :)

jsulpis (approved):

PR #6853, opened 2026-08-31
fix(TagList): avoid issues with duplicated tag labels in tests

jsulpis:
### Summary

- add a `shuffle` function in `@ultraviolet/utils`
- use it to shuffle the labels of hidden tags in the measurement container to avoid issues with duplicate labels in tests


.changeset/seven-trains-brake.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ultraviolet/ui": patch
+---
+
+`TagList`: 
+- avoid issues with duplicated tag labels in tests
+- set `data-testid` on visible tags (including the popover) instead of the hidden ones
```
lisalupi, line 7: should I close #6841 then?

packages/ui/src/components/TagList/__tests__/index.test.tsx
```diff
@@ -5,6 +5,15 @@ import { afterAll, describe, expect, it, vi } from 'vitest'
 import { TagList } from '..'
 import type { TagType } from '..'
 
+vi.mock('@ultraviolet/utils', async importOriginal => {
```
lisalupi, line 8: why do you need to mock this?

philibea (approved):

lisalupi (approved):

.changeset/seven-trains-brake.md
jsulpis: Yes it's a duplicate, I needed the fix in this PR

packages/ui/src/components/TagList/__tests__/index.test.tsx
jsulpis: it's a way to have a deterministic "shuffle" in order to have stable snapshots

philibea (approved):

PR #6861, opened 2026-09-02
feat(Button): add accessible labeling and tooltip relation props

jsulpis:
### Summary

Add new `tooltipLabel`, `tooltipDescription` and `accessibleLabel` props to `Button` so accessible names and descriptions can be provided via a localizable hidden label and a tooltip with an explicit relation. Also remove the Tooltip opening delay in tests to make them faster and less flaky.


packages/ui/src/helpers/isTestEnvironment.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Check if the code is running in a test environment (jsdom, happy-dom, ...).
+ * This is useful to skip animations or delays that would make tests flaky.
+ *
+ * The detection relies on the `navigator.userAgent` set by test DOMs, so it
+ * survives library bundling (unlike `process.env.NODE_ENV` which is inlined at
+ * build time).
+ */
+export const isTestEnvironment =
```
jsulpis: not very robust but at least it allows to avoid delays on the Tooltip in our repo. In consumer projects, in the worst case this will be false and the Tooltip will have a delay like before. 

Afterthought: maybe `isHeadlessTestEnvironment` ? Because it would be false in browser tests

packages/ui/src/components/Tooltip/useTooltip.ts
```diff
@@ -58,11 +57,8 @@ export const useTooltip = ({ visible, placement, delay, onOpenChange }: UseToolt
     placement: (placement.replace(/auto-?/, '') || 'top') as Placement,
   })
 
-  // remove the transition when prefers reduced motion, mostly to avoid waiting in tests
-  const prefersReducedMotion = usePrefersReducedMotion()
```
philibea, line 62: why this change ?

jsulpis: the setting `prefersReduceMotion` was used only to remove the transition (fade in/out), to "avoid waiting in tests". But the opening delay was preserved. So I used instead `isTestingEnvironment` to remove both the transition and the delay in tests, so we don't need `prefersReduceMotion` anymore (users with the setting enabled will have the fade in/out transition but I think it's ok)

packages/ui/src/helpers/isTestEnvironment.ts
philibea: i prefer isHeadlessTestEnvironment also

jsulpis: ✅

philibea (approved):

PR #6868, opened 2026-09-03
feat(agent): add unit testing skill

philibea:
## Summary

add unit testing skill

### DIFF
--- a/.agents/skills/unit-testing/SKILL.md
+++ b/.agents/skills/unit-testing/SKILL.md
@@ -81,6 +81,19 @@ The line is thin, but split files by intent:
 If a test's real question is "is this accessible?", it's an a11y test. If it's
 "does this work?", it's a regular test. Same query tools, different intent.
 
+### Accessible-name/description matchers
+
+Prefer `jest-dom` matchers over raw attribute checks — attributes are the
+implementation, accessible names/descriptions are the behavior:
+
+- `expect(radio).toHaveAccessibleName('Agree')` — over `toHaveAttribute('aria-label', ...)`
+- `expect(radio).toHaveAccessibleDescription('Invalid value')` — over checking
+  `aria-describedby`/`title` wiring
+- `expect(button).toBeDisabled()` — over `toHaveAttribute('disabled')`
+
+These compute the name/description from the whole source chain (label, aria-label,
+aria-labelledby, title), so they pass when the label works and fail when it breaks.
+
 ## Snapshot policy
 
 A snapshot only locks the initial DOM — it passes until someone clicks "update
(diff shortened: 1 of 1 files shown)


.agents/skills/unit-testing/SKILL.md
```diff
@@ -0,0 +1,104 @@
...
+   3. `getByPlaceholderText` — placeholder is not a label, only use if that's all there is.
+   4. `getByText` — for non-interactive elements (div, span, p) outside forms.
+   5. `getByDisplayValue` — for filled-in form values.
+2. **Semantic queries** — HTML5/ARIA selectors, but inconsistent across browsers/screen readers.
+   1. `getByAltText` — for `img`, `area`, `input`.
+   2. `getByTitle` — title is not read consistently and not visible to sighted users.
+3. **Test IDs** — `getByTestId` only when you can't match by role or text and it doesn't
+   make sense (e.g. dynamic text). Users can't see them; a testid is a last resort.
+
+### Query variants
```
jsulpis: I think this section is not necessary, the models already know this

.agents/skills/unit-testing/SKILL.md
```diff
@@ -0,0 +1,104 @@
...
+| `findAllBy...`  | throws    | array   | array      | **yes** |
+
+- `queryBy...` → assert something is **not** present.
+- `findBy...` → content appears **asynchronously** (await it).
+- Prefer regex `/submit/i` over string + `{ exact: false }`.
+
+### By component type
+
+- Interactive (button, link, tab, switch): `getByRole`
+- Form fields: `getByLabelText` / `getByDisplayValue`
```
jsulpis: I can't find the page where playwright listed the locators by priority, I am not sure if we should prioritize `getByLabel` or `getByRole` for form fields 🤔

jsulpis (commented):
should we add the instruction to gather all accessibility tests in a separate `a11y.test.tsx` file ?

.agents/skills/unit-testing/SKILL.md
philibea: Both work for a labeled input (`getByLabel('Password')` and `getByRole('textbox', { name: 'Password' })` resolve to the same element)


In case of form, all ours input should have "label" define so i think it's better to enforce this pattern ? wdyt ?

.agents/skills/unit-testing/SKILL.md
philibea: yes we can  remove if necessary, still find to remind the model just in case

jsulpis: the tests will fail if it uses the wrong locator so IMO it's just unnecessary instructions that fill the agent's context

.agents/skills/unit-testing/SKILL.md
jsulpis: `getByRole` looks more precise to me because it checks the role in addition to the label, whereas `getByLabel` only checks the label. I think it would be easier to just use getByRole whenever we can, without wondering if it's a form field or not

philibea: changed

.agents/skills/unit-testing/SKILL.md
philibea: changed

jsulpis (approved):

PR #6871, opened 2026-09-03
test(Radio, RadioGroup): replace snapshot-only tests with behavior assertions

### LINKED ISSUE #6863: [TEST] Rewrite snapshot-only tests: Radio, RadioGroup
Rewrite snapshot-only unit tests to behavior assertions, following the `unit-testing` skill (.agents/skills/unit-testing/SKILL.md).

## Scope
Rewrite `packages/ui/src/components/Radio/__tests__/index.test.tsx` and `packages/ui/src/components/RadioGroup/__tests__/index.test.tsx`.

## Rules
- Keep ONE snapshot of the default render to lock the initial DOM.
- Replace every variant/prop snapshot with behavior assertions (screen.getByRole, userEvent, toHaveBeenCalled, toHaveAttribute).
- Delete redundant variant snapshots.
- Use renderWithTheme; prefer screen.getByRole; interact with userEvent.
- Run pnpm run test:unit, pnpm run lint, and typecheck before finishing.
- Open a PR titled "test(Radio, RadioGroup): replace snapshot-only tests with behavior assertions".

philibea:
Closes #6863

Rewrites the snapshot-only unit tests for `Radio` and `RadioGroup` into behavior assertions, following the `unit-testing` skill (one default-render snapshot per file + behavior assertions for every prop/variant).

## Radio
- Keeps one default-render snapshot to lock the initial DOM.
- Replaces variant/prop snapshots with behavior assertions:
  - name/value forwarded to the input (`toHaveAttribute`)
  - disabled → input disabled + `aria-disabled`
  - checked → `toBeChecked()`
  - error description rendered as `role="status"` and linked to the input via `aria-describedby`
  - helper description linked via `aria-describedby` (rendered at `size="small"` to exercise the small path)
  - tooltip exposed on hover (`userEvent.hover` → `findByRole('tooltip')`)
  - `onChange` fired on click (`userEvent.click`)

## RadioGroup
- Keeps one default-render snapshot to lock the initial DOM.
- Replaces variant/prop snapshots with behavior assertions:
  - legend + fieldset + option radios rendered, with the selected option checked
  - `direction="row"` layout keeps options queryable
  - helper description linked to the fieldset via `aria-describedby`
  - error description linked to the fieldset via `aria-describedby` (rendered at `size="small"` to exercise the small path)
  - `onChange` fired when an option is selected
  - context guard: `RadioGroup.Radio` outside a `RadioGroup` throws

## Notes
- The `size` prop only changes visual styling (gap/padding) with no DOM/behavior difference observable in jsdom, so it is exercised alongside a real behavior assertion (helper/error + `aria-describedby` link) rather than via a standalone snapshot.
- The fieldset's `<legend>` is nested inside layout `Stack`s rather than a direct child of `<fieldset>`, so the group role is not named from the legend; the fieldset is therefore queried directly. (This is a pre-existing component concern, out of scope here.)

## Verification
- `pnpm vitest --run src/components/Radio src/components/RadioGroup` — 15 passed
- `pnpm run lint` — clean (no new warnings)
- `pnpm --filter @ultraviolet/ui run typecheck` — clean


packages/ui/src/components/Radio/__tests__/index.test.tsx
```diff
@@ -1,57 +1,62 @@
...
-    const { asFragment } = renderWithTheme(
-      <Radio checked label="Choice" name="radio" onChange={() => {}} value="choice" />,
-    )
-    expect(asFragment()).toMatchSnapshot()
+  it('renders the error description and links it to the input', () => {
+    renderWithTheme(<Radio error="Invalid value" label="Choice" name="radio" onChange={() => {}} value="choice" />)
+    const radio = screen.getByRole('radio', { name: 'Choice' })
+    const description = screen.getByRole('status')
+    expect(description).toHaveTextContent('Invalid value')
+    expect(radio).toHaveAttribute('aria-describedby', description.id)
```
jsulpis: should be `expect(radio).toHaveAccessibleDescription('Invalid value')`

I think this could be added to the skill, use existing matchers whenever possible instead of checking attributes (testing the implementation and not the behavior)

Same in the other tests and in RadioGroup

jsulpis (changes requested):

packages/ui/src/components/Radio/__tests__/index.test.tsx
philibea: @jsulpis https://github.com/scaleway/ultraviolet/pull/6871#issuecomment-5539140457

jsulpis (approved):

PR #6873, opened 2026-09-04
docs: add color contrast page

jsulpis:
### Summary

Add a page to check color contrast for all text/backgrounds pairings.

### Relevant logs and/or screenshots

image

image


pnpm-lock.yaml
```diff
@@ -2804,18 +2804,6 @@ packages:
     peerDependencies:
       react: ^16.14 || ^17.0 || ^18.0 || ^19.0
```
philibea, line 2809: why do we have a pnpm-lock change ? there is no package.json changed

jsulpis: ✅

utils/stories/src/theme/colorContrast/ContrastChecker.tsx
```diff
@@ -0,0 +1,315 @@
...
+  const l2 = relativeLuminance(bg)
+  return (Math.max(l1, l2) + 0.05) / (Math.min(l1, l2) + 0.05)
+}
+
+// --- Thresholds & levels ---
+
+const AA_THRESHOLD = 4.5
+type ContrastLevel = 'pass' | 'fail'
+
+const getContrastLevel = (ratio: number): ContrastLevel => (ratio >= AA_THRESHOLD ? 'pass' : 'fail')
```
lisalupi: since disabled components don't need to meet requirements, do we need to mark them as a fail (when they do)?

utils/stories/src/theme/colorContrast/ContrastChecker.tsx
```diff
@@ -0,0 +1,315 @@
...
+  </Stack>
+)
+
+const SummaryBar = ({ counts }: { counts: { total: number; pass: number; fail: number } }) => (
+  <Stack className={contrastStyle.summaryBar} direction="row" gap={2} alignItems="center" wrap>
+    <Text as="span" prominence="strong" sentiment="neutral" variant="bodySmallStrong">
+      {counts.total} intended pairings
+    </Text>
+    <Badge sentiment="success">
+      {'\u2713'} {counts.pass} pass
```
lisalupi: why not use uv icons here?

utils/stories/src/theme/colorContrast/ContrastChecker.tsx
jsulpis: indeed, I wanted to keep them visible because we rely a lot on disabled button so they should be readable, but we can remove the pass/fail for disabled pairings

lisalupi: you could add a third state (neutral or whatever sentiment) with "non applicable" maybe?

jsulpis: yep, did that

utils/stories/src/theme/colorContrast/ContrastChecker.tsx
jsulpis: fixed

utils/stories/src/theme/colorContrast/ContrastChecker.tsx
```diff
@@ -0,0 +1,328 @@
...
+
+// Module-scope persistence: the theme switcher remounts the story tree, so plain
+// useState is wiped on each light/dark/darker toggle. Keep it here to survive remounts.
+const persist = {
+  highlightFailures: false,
+  hideDisabled: false,
+  scrollY: 0,
+}
+
+// --- WCAG contrast computation ---
```
philibea: can be put outside of the jsx inside a custom helpers.ts ?

lisalupi (approved):

utils/stories/src/theme/colorContrast/ContrastChecker.tsx
jsulpis: ✅

philibea (approved):

PR #6879, opened 2026-09-04
ci: add pkg-pr-new workflow for continuous preview releases

philibea:
## What

Adds a [`pkg.pr.new`](https://github.com/stackblitz-labs/pkg.pr.new) workflow that publishes preview builds of all `./packages/*` on every pull request. Preview packages are installable via `npm i https://pkg.pr.new/scaleway/ultraviolet/<pkg>@<sha>` without waiting for an npm release.

## Changes

- `.github/workflows/pkg-pr-new.yml` — new workflow triggered on `pull_request`, with concurrency to cancel superseded runs. Builds the workspace then runs `pnpm exec pkg-pr-new publish --pnpm --commentWithSha './packages/*'`.
- `package.json` / `pnpm-lock.yaml` — add `pkg-pr-new` as a workspace devDependency (the docs warn against `npx`/`pnpm dlx`/`bunx` in CI; run from the lockfile via `pnpm exec`).

## Notes

- `--pnpm` is required because this workspace uses pnpm `catalog:` — `npm pack` cannot resolve it.
- `--commentWithSha` makes the bot comment link to the commit SHA so each push gets a stable URL.
- One-time setup outside this repo: install the [pkg.pr.new GitHub App](https://github.com/apps/pkg-pr-new) on `scaleway/ultraviolet` before the first run.

### DIFF
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -1,0 +1,25 @@
+name: pkg-pr-new
+
+on:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  publish:
+    runs-on: ubuntu-24.04
+    env:
+      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+      TURBO_CACHE: 'remote:rw'
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,25 @@
+name: pkg-pr-new
+
+on:
+  pull_request:
```
jsulpis, line 4: I think it could be a manual workflow no ? We don't need it on every PR. I don't know if we can do manual jobs like on gitlab, that would be ideal (just trigger a job from a PR page without having to go trigger a workflow in the actions page)

jsulpis (changes requested):

.github/workflows/pkg-pr-new.yml
philibea: @jsulpis 
should work with the label  `pkg-pr-new`  only now

jsulpis: perfect, I'm not sure about the label, maybe `publish-pr` or `publish-commits` would be more explicit and not linked to the implementation ?

jsulpis (approved):

PR #6894, opened 2026-09-07
feat(storybook): distinguish unaudited components in a11y audit

jsulpis:
### Summary

- Mark components without an `A11y.mdx` audit as unaudited by setting their `a11yStatus` flags to `undefined` instead of `false`. This differentiates components that haven't been audited from those that fail all checks.
- In the Accessibility Audit page, add a new `unknown` a11y level: unaudited components now display **"Not audited"**.

### Screenshots

|   Before   |   After    |
| :--------: | :--------: |
| image |image |

### DIFF
--- a/packages/ui/src/components/Action/CopyButton/__stories__/index.stories.tsx
+++ b/packages/ui/src/components/Action/CopyButton/__stories__/index.stories.tsx
@@ -14,10 +14,10 @@ export default {
   title: 'UI/Action/CopyButton',
   parameters: {
     a11yStatus: {
-      perceivable: false,
-      operable: false,
-      understandable: false,
-      robust: false,
+      perceivable: undefined,
+      operable: undefined,
+      understandable: undefined,
+      robust: undefined,
     },
   },
 } as Meta<typeof CopyButton>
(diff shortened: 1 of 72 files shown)


packages/ui/src/components/Row/__stories__/index.stories.tsx
```diff
@@ -6,10 +6,10 @@ export default {
   title: 'UI/Layout/Row',
   parameters: {
     a11yStatus: {
-      perceivable: true,
-      operable: true,
-      understandable: true,
-      robust: true,
+      perceivable: undefined,
```
lisalupi: everything was set to "true" but the component hasn't been audited 🤔 same for `Stack`

lisalupi (approved):

lisalupi (approved):

packages/ui/src/components/Row/__stories__/index.stories.tsx
jsulpis: good catch, I reverted

philibea (approved):

PR #6895, opened 2026-09-07
fix: update Flag and Submit props

jsulpis:
### Summary

- **Submit** (`@ultraviolet/form`): spread all `Button` props directly, adding `accessibleLabel`, `tooltipLabel` and `tooltipDescription` instead of listing a subset individually.
- **Flag (Icon)** (`@ultraviolet/icons`): `accessibleLabel` is now a `string` (the flag's accessible name) instead of a boolean, and `className`/`aria-hidden` are now accepted via `SVGProps`.

### Screenshots

_No visual change._

### DIFF
--- a/packages/form/src/components/Submit/index.tsx
+++ b/packages/form/src/components/Submit/index.tsx
@@ -13,7 +13,17 @@ type SubmitButtonProps = {
 } & Partial<
   Pick<
     ComponentProps<typeof Button>,
-    'size' | 'sentiment' | 'variant' | 'tooltip' | 'fullWidth' | 'onClick' | 'disabled' | 'className'
+    | 'size'
+    | 'sentiment'
+    | 'variant'
+    | 'tooltip'
+    | 'accessibleLabel'
+    | 'tooltipLabel'
+    | 'tooltipDescription'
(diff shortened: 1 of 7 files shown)


philibea (approved):

packages/form/src/components/Submit/index.tsx
```diff
@@ -13,41 +13,34 @@ type SubmitButtonProps = {
...
 
 type SubmitProps<TFieldValues extends FieldValues> = RHFBase<TFieldValues> & SubmitButtonProps
 
 export const Submit = <TFieldValues extends FieldValues>({
   children,
-  className,
-  disabled = false,
-  size,
-  variant = 'filled',
-  sentiment = 'primary',
```
philibea, line 28: this is not define by default or we let the default of the button ?

jsulpis: oops I didn't see that but the Button has the same default values and I think it's unlikely to change, so I think we can keep it like that if it's ok for you

philibea (approved):

lisalupi (approved):

PR #6900, opened 2026-09-09
fix(Tooltip): render relation="label" tooltips inline and move to portal on open

jsulpis:
### Summary

- `Tooltip` with `relation="label"` now renders the tooltip **inline** in the DOM, visually hidden until hover/focus, and only moves it to the portal target (`body` by default) when it opens. This allows to:
  1. not flood the body with tooltip elements like we had before
  2. append the tooltip element _after_ existing portaled elements (popover) in order to render the tooltip on top of them
- `relation="description"` behavior is unchanged.
- `Button.tooltipLabel` keeps using the `label` relation (accessible name via `aria-labelledby`).

### Tests
- `Tooltip`: label tooltip renders inline when closed and moves into the portal target on hover.

### DIFF
--- a/packages/ui/src/components/Action/Button/__tests__/index.test.tsx
+++ b/packages/ui/src/components/Action/Button/__tests__/index.test.tsx
@@ -148,10 +148,19 @@ describe('button', () => {
     ))
 
   describe('tooltipLabel and tooltipDescription', () => {
-    it('renders tooltipLabel as the accessible name (label relation)', () => {
-      renderWithTheme(<Button tooltipLabel="More info">Hello</Button>)
+    it('renders tooltipLabel as the accessible name (label relation)', async () => {
+      renderWithTheme(
+        <Button tooltipLabel="More info">
+          <PencilOutlineIcon />
+        </Button>,
+      )
 
       const trigger = screen.getByRole('button', { name: 'More info' })
+      expect(screen.getByText('More info')).toBeInTheDocument()
+      expect(trigger).not.toHaveAccessibleDescription()
+
+      await userEvent.hover(trigger)
+      expect(trigger).toHaveAccessibleName('More info')
       expect(trigger).not.toHaveAccessibleDescription()
     })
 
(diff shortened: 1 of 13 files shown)


packages/ui/src/components/Action/Button/__tests__/index.test.tsx
```diff
@@ -148,10 +148,19 @@ describe('button', () => {
...
+          <PencilOutlineIcon />
+        </Button>,
+      )
 
       const trigger = screen.getByRole('button', { name: 'More info' })
+      expect(screen.getByText('More info')).toBeInTheDocument()
+      expect(trigger).not.toHaveAccessibleDescription()
+
+      await userEvent.hover(trigger)
+      expect(trigger).toHaveAccessibleName('More info')
```
lisalupi: maybe check this before hovering the tooltip to be sure that the accessible name doesn't come from the tooltip? not sure if this is relevent

packages/ui/src/components/Action/Button/index.tsx
```diff
@@ -170,8 +170,8 @@ export const Button = forwardRef<Element, BaseButtonProps>(
     }, [sentiment, theme, variant])
 
     const tooltipText = tooltipLabel || tooltipDescription || tooltip
-    const tooltipRelation = tooltipLabel ? 'label' : 'description'
-    const computedAccessibleLabel = accessibleLabel ?? ariaLabel
+    const tooltipRelation = tooltipLabel ? 'none' : 'description'
+    const computedAccessibleLabel = accessibleLabel ?? tooltipLabel ?? ariaLabel
```
lisalupi: is it possible to either document priorities (if both are defined, the computed name of the button will be `accessibleLabel`, not `tooltipLabel`) or use XOR for type definition so that `accessibleLabel` & `tooltipLabel` are not defined simulatenously?

philibea (approved):

lisalupi (approved):

PR #6901, opened 2026-09-09
fix(OrderSummary): display non undefined details

lisalupi:
### Summary

`OrderSummary`: display categories, sub-categories and details when values are **defined** instead of **truthy** (categories/sub-categories/details with title=0 or title="" can now be displayed)

### DIFF
--- a/packages/ui/src/compositions/OrderSummary/__tests__/index.test.tsx
+++ b/packages/ui/src/compositions/OrderSummary/__tests__/index.test.tsx
@@ -1,6 +1,6 @@
 import { screen } from '@testing-library/react'
 import { userEvent } from '@testing-library/user-event'
-import { renderWithTheme, shouldMatchSnapshot } from '@utils/test'
+import { renderWithTheme } from '@utils/test'
 import { describe, expect, it } from 'vitest'
 import { OrderSummary } from '..'
 import {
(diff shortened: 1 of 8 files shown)


packages/ui/src/compositions/OrderSummary/__tests__/index.test.tsx
```diff
@@ -284,4 +320,22 @@ describe('orderSummary', () => {
 
     expect(asFragment()).toMatchSnapshot()
   })
+
+  it('should work with falsy titles', async () => {
```
jsulpis: the combo "should work" + snapshot that we have in this test suite tells absolutely nothing about the expected behavior, so if a test fails we won't know what broke and how to fix it.

One of the purposes of tests is to document the expected behavior, it's clear in the test you added but it's even better to have it in plain text in the title ;)

```suggestion
  it('should display titles that are defined and falsy', async () => {
```

packages/ui/src/compositions/OrderSummary/components/Category.tsx
```diff
@@ -19,7 +19,7 @@ export const CategoryName = ({ category }: { category: ItemsType }) => {
     totalPriceWithDiscount: 0,
   }
 
-  return category.category ? (
+  return category.category !== undefined ? (
```
jsulpis: I think we don't want to render `null` right ? Same for the other occurrences of `!== undefined`

```suggestion
  return category.category != undefined ? (
```

philibea: or you can use `isNullOrUndefined`

jsulpis (approved):

PR #6903, opened 2026-09-09
feat(Filters): hide clear all button

lisalupi:
### Summary

`Filters`: new prop `hideClearAll` to hide the "Clear All" button on the main row (new property of `layout` prop)


jsulpis (approved):
there is a little blank space next to the "all filters" button, but I think the clear all button should be hidden when the row is not full so this should not be a problem right ?

Image

PR #6906, opened 2026-09-10
docs(Dialog): audit accessibility

lisalupi:
### Summary

Audit `Dialog` accessibility

### DIFF
--- a/packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
+++ b/packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
@@ -1,0 +1,69 @@
+import { Meta } from '@storybook/addon-docs/blocks'
+
+<Meta title="UI/Overlay/Dialog/A11y" />
+
+# Dialog - Accessibility
+
+## Description
+
+A confirmation/message dialog that presents a short title with a sentiment-based icon and content requiring user interaction. Built on top of `Modal`, it renders a `<dialog>` with focus trap, Escape-to-close, and a set of sub-components (`Button`, `CancelButton`, `Buttons`, `Stack`, `Text`) for composing confirmations.
+
+## Summary
+
+1. ✅ **Perceivable**: Content is text-based; sentiment is conveyed by the title/message, so meaning does not depend on color alone.
+2. ⚠️ **Operable**: Keyboard support (Tab, Escape, focus trap) present, but initial focus lands on the `<dialog>` container instead of the least destructive action (Cancel), and focus is not restored for programmatic (non-disclosure) usage.
+3. ✅ **Understandable**: Title rendered as a semantic `h2` heading, clear button labels.
+4. ❌ **Robust**: Accessible name defaults to the generic `"modal"` instead of the visible title; the visible `h2` title is not linked to the dialog via `aria-labelledby`.
+
+## ARIA Pattern
+
+[Dialog (Modal) Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)
+
+## Rules
+
+Enforced by the component:
+
+- ✅ [WCAG 2.1.1 - Keyboard (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html) - All interactions (open via disclosure, close, confirm/cancel) are operable via keyboard
+- ✅ [WCAG 2.1.2 - No Keyboard Trap (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/no-keyboard-trap.html) - Focus trap via `handleFocusTrap` keeps focus cycling inside while allowing Escape to exit
+- ⚠️ [WCAG 2.4.3 - Focus Order (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html) - Initial focus on the `<dialog>` element (via `tabIndex={0}`), not the least destructive action; incomplete focus restoration for programmatic modals
+- ✅ [WCAG 2.4.6 - Headings and Labels (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/headings-and-labels.html) - Title is a descriptive heading
+- ❌ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) - Accessible name defaults to generic `"modal"` (`aria-label`) and is not derived from the visible title; no `aria-labelledby` relationship
+
+To apply when using the component:
+
+- [WCAG 1.1.1 - Non-text Content (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html) - The sentiment icon (Check/AlertCircle) is `aria-hidden`; ensure the title or message conveys the meaning
+- [WCAG 1.4.1 - Use of Color (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html) - Sentiment is conveyed via icon/color; do not rely on color alone, include meaningful text
+- [WCAG 1.3.1 - Info and Relationships (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) - Maintain semantic structure (headings, lists) within dialog content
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
```diff
@@ -0,0 +1,69 @@
...
+- ⚠️ [WCAG 2.4.3 - Focus Order (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html) - Initial focus on the `<dialog>` element (via `tabIndex={0}`), not the least destructive action; incomplete focus restoration for programmatic modals
+- ✅ [WCAG 2.4.6 - Headings and Labels (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/headings-and-labels.html) - Title is a descriptive heading
+- ❌ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) - Accessible name defaults to generic `"modal"` (`aria-label`) and is not derived from the visible title; no `aria-labelledby` relationship
+
+To apply when using the component:
+
+- [WCAG 1.1.1 - Non-text Content (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html) - The sentiment icon (Check/AlertCircle) is `aria-hidden`; ensure the title or message conveys the meaning
+- [WCAG 1.4.1 - Use of Color (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html) - Sentiment is conveyed via icon/color; do not rely on color alone, include meaningful text
```
jsulpis: a bit redundant with the rule above no ? _"ensure the title or message conveys the meaning"_

packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
```diff
@@ -0,0 +1,69 @@
...
+- [WCAG 1.3.1 - Info and Relationships (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) - Maintain semantic structure (headings, lists) within dialog content
+- [WCAG 1.4.3 / 1.4.11 - Contrast (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html) - Ensure text and interactive elements inside have sufficient contrast
+- [WCAG 2.4.4 - Link Purpose (In Context) (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context.html) - Use descriptive link text within dialog content
+- [WCAG 3.3.2 - Labels or Instructions (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/labels-or-instructions.html) - Label any form inputs placed inside the dialog
```
jsulpis: the last 3 rules look very generic, they should be applied everywhere so I don't think they are necessary in this component audit, wdyt ?

packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+## Accessibility issues
+
+### Critical
+
+- **Accessible name is the generic `"modal"` instead of the title**: The `<dialog>` gets its name from `aria-label`, which defaults to `'modal'`. The visible `<h2>` title rendered by the `title` prop is not programmatically associated (`aria-labelledby`), so screen reader users hear "modal" rather than e.g. "Discard changes?". The title should drive the accessible name of the dialog.
+
+### High
+
+- **Incorrect initial focus placement**: Focus lands on the `<dialog>` element itself (`tabIndex={0}`) rather than an element inside. For a confirmation dialog, focus should move to the **least destructive action** (e.g. the Cancel button) per the Dialog pattern.
```
jsulpis: quick toughts:
- while it's probably not encouraged, since the Dialog buttons are in a slot it's technically possible to have no cancel button
- I think it's acceptable and it would probably be easier to focus the close button since it's not in the slot, but it's also possible that there is no close button with the prop `isClosable=false`
- the focus is managed by the Modal component which can have no cancel button, so the solution should work in any case
- in the most extreme case I can imagine a Modal for a loading state with no button inside, so we can only focus the dialog element

So I think we should use fallbacks in the Modal component: focus the close button OR the first focusable element (which could be the cancel button in a dialog without close button) OR the dialog element. wdyt ?

packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+- **No focus restoration for programmatic usage**: When opened with the `open` prop (no `disclosure`), focus is not returned to a logical element on close. Focus should return to the invoking element.
+
+### Medium
+
+- **Visible `title` not associated with the dialog**: The `title` prop renders visible text but it is not wired to the `<dialog>` via `aria-labelledby`. Auto-generate an `id` on the title and reference it so the accessible name matches the visible title, and expose an `ariaLabelledBy` prop for advanced cases referencing an external title.
```
jsulpis: Do you think it would be possible to have a visible label of the Dialog _outside_ of the Dialog ? I don't see how, I would not add this prop

packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+### Low
+
+- **No `aria-describedby` support**: The description (`Dialog.Text`) is not associated with the dialog; wiring it up would help announce context for simple dialogs.
```
jsulpis: it's optional in the Dialog pattern and should be omitted if there are complex content or multiple paragraph, and it would be complex to setup because the Text is in a slot. So I think we can skip this

jsulpis (changes requested):

packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
lisalupi: i'd say the pattern should be: least destructive > first focusable element > close button > modal itself

jsulpis (approved):

philibea (approved):

packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
```diff
@@ -0,0 +1,63 @@
...
+# Dialog - Accessibility
+
+## Description
+
+A confirmation/message dialog that presents a short title with a sentiment-based icon and content requiring user interaction. Built on top of `Modal`, it renders a `<dialog>` with focus trap, Escape-to-close, and a set of sub-components (`Button`, `CancelButton`, `Buttons`, `Stack`, `Text`) for composing confirmations.
+
+## Summary
+
+1. ✅ **Perceivable**: Content is text-based; sentiment is conveyed by the title/message, so meaning does not depend on color alone.
+2. ⚠️ **Operable**: Keyboard support (Tab, Escape, focus trap) present, but initial focus lands on the `<dialog>` container instead of the least destructive action (Cancel), and focus is not restored for programmatic (non-disclosure) usage.
```
iManu, line 14: Note focus is restored using Esc key, but not on other buttons (cancel, validate, close cross)

packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
```diff
@@ -0,0 +1,63 @@
...
+3. ✅ **Understandable**: Title rendered as a semantic `h2` heading, clear button labels.
+4. ❌ **Robust**: Accessible name defaults to the generic `"modal"` instead of the visible title; the visible `h2` title is not linked to the dialog via `aria-labelledby`.
```
iManu, line 16: True, le aria-label sur dialog devrait être supprimé et utiliser mettre un ref aria avec l'id du h2

iManu (approved):

packages/ui/src/components/Overlay/Dialog/__stories__/A11y.mdx
iManu: Noter que VoiceOver retourne bien le h2, il faudra vérifier qu'il n'existe pas de répétition

Capture d’écran 2026-09-16 à 14 17 27
Capture d’écran 2026-09-16 à 14 17 39

PR #6909, opened 2026-09-11
perf(CodeEditor): lazy load all languages

jsulpis:
### Summary

`CodeEditor` previously bundled every language grammar (total ~530kB minzipped) via `@uiw/codemirror-extensions-langs`.

Now languages are lazy-loaded on demand:
- `@codemirror/language-data` resolves languages by extension and `import()`s them only when used.
- `nix`, `svelte`, `solidity` (missing from language-data) added via `@replit/codemirror-lang-*`.
- `extensions` is now a typed `LanguageName` union; unknown values fall back to `sh` (unchanged).

Dropped `@uiw/codemirror-extensions-langs` and updated the renovate CodeMirror matcher.

**Note**: the only functional change should be that the `python` extension is no longer recognized (other python extensions like `py` or `pyw` still work). It was mapped by `@uiw/codemirror-extensions-langs` but it's not in `@codemirror/language-data`. I checked in the console that only `.py` is used.

### How to test

Baseline in prod:
1. go to https://storybook.ultraviolet.scaleway.com/?path=/story/compositions-codeeditor--playground
2. open the network tab
3. see a huge CodeEditor js chunk
4. when changing the language in the playground ("rs", "py", "lua"...) you should not see new js files downloaded because all languages are already in the initial chunk

Lazy-load in preview:
1. go to https://ultravioletjxvhzygy-codemirrorperfstorybook.functions.fnc.fr-par.scw.cloud/?path=/story/compositions-codeeditor--playground
2. open the network tab
3. see a _much_ smaller CodeEditor js chunk
4. when changing the language in the playground ("rs", "py", "lua"...) you should see new js files downloaded on demand

You can check the `solidity`, `nix` and `svelte` languages that are loaded differently from the others


packages/ui/src/compositions/CodeEditor/type.ts
```diff
@@ -1,11 +1,239 @@
-import type { langs } from '@uiw/codemirror-extensions-langs'
 import type { ReactCodeMirrorProps } from '@uiw/react-codemirror'
 import type { CSSProperties, ReactNode } from 'react'
 
+type CodeEditorLanguageName =
```
philibea, line 4: question(non-blocking): This should come from  the package directly or we should build it with type from this package for futur update ?
i wonder how can we stay sync with the package in a futur update

jsulpis: that's a big downside but `@codemirror/language-data` types the `extensions` as `string` so we can't derive this type from the package 😕
I think (hope) that any change to the supported languages would be stated as a breaking change so we would see it and update our type

packages/form/src/compositions/CodeEditorField/__tests__/index.test.tsx
```diff
@@ -1,15 +1,24 @@
+import { screen, waitFor } from '@testing-library/react'
 import { describe, expect, it } from 'vitest'
 import { CodeEditorField } from '..'
 import { renderWithForm } from '../../../__tests__/helpers'
 
 describe('optionSelectorField', () => {
```
lisalupi:
```suggestion
describe('codeEditorField', () => {
```
not a new issue, but since you edited the file...

lisalupi (approved):

packages/ui/src/compositions/CodeEditor/type.ts
lisalupi: I wouldn't count on them disclosing every breaking change, given a few releases we had to deal with 🥲

jsulpis: Yes there might be a risk but I think it's worth the 500kB of JS saved. The worst that can happen is to loose syntax highlighting (or by chance it kind of works with the fallback sh language) on a language that they would drop without announcing it

packages/form/src/compositions/CodeEditorField/__tests__/index.test.tsx
jsulpis: ✅

philibea (approved):

PR #6918, opened 2026-09-14
fix: add missing compositions files in the `dist` folder

jsulpis:
### Summary

- since `@ultraviolet/ui@3.21.2` the files `compositions/index.js` and `compositions/styles.js` are no longer in the `dist` file and they are referenced in the `package.json` so the exports are broken
- `src/composition.ts` looks unused but it's probably what pulled the 2 files because of Rolldown's `preserveModules` option that we have enabled. The version of vite 8.2.0 probably changed that.

This PR removes the unused `composition.ts` file and properly declare the two files as build entries.


packages/ui/vite.config.ts
```diff
@@ -7,10 +7,9 @@ export const config = mergeConfig(defineConfig(defaultConfig), {
   build: {
     lib: {
       entry: {
-        'compositions/Navigation/index': resolve(import.meta.dirname, 'src/compositions/Navigation/index.tsx'),
-        'compositions/EstimateCost/index': resolve(import.meta.dirname, 'src/compositions/EstimateCost/index.tsx'),
         'compositions/Filters/index': resolve(import.meta.dirname, 'src/compositions/Filters/index.ts'),
```
jsulpis, line 10: `Filters` is the only component which is not defined in an `index.tsx` file and without this line the `index.ts` file is omitted from the bundle. The two other components have their `index.js` emitted.

alexandre-combemorel (approved):

lisalupi (approved):

PR #6919, opened 2026-09-14
feat: don't hash css variables

lisalupi:
### Summary

> The following changes apply _only_ to `ThemeProvider` from `@ultraviolet/themes`, since the one from `@ultraviolet/ui` is deprecated and won't be updated anymore.

`ThemeProvider`: do not hash css variable names.

The name of the CSS variables defined in the ThemeProvider are not hashed anymore. This has a couple of consequences:

- clearer naming
- possibility to use the variables without importing css files (`@ultraviolet/themes/light.css` and other variants)
- reduces FOUC

Update generated `.css` files to match the `.js` ones (missing "sizing" variables, mismatch between font-weight value)

I also updated the README with those changes as well as the ones in #6908 

⚠️⚠️ **Breaking change**: If `@ultraviolet/themes/light.css` (and other variants) is imported, any custom theme defined in the theme provider **will be** overwritten by the .css file (see `examples/next/src/pages/_app.tsx`: to apply the custom *primary text* , `@ultraviolet/themes/light.css` is imported in a layer to reduce its specificity).⚠️⚠️

### Screenshots

_If the PR introduces visual changes, in the components or Storybook, you can show the changes with screenshots. Otherwise, you can remove this section._

|   Before   |   After    |
| :--------: | :--------: |
| https://github.com/user-attachments/assets/3f7baa7c-75d7-486a-a15f-5611c3ed0aee | https://github.com/user-attachments/assets/162e5cdc-ec38-4ac8-98bd-53f0af1a6d9a |
| Capture d’écran 2026-09-15 à 15 15 12 | Capture d’écran 2026-09-15 à 15 16 09|

This is the next example, which imports css files. Without them, there is still a flash while loading since the theme is defined at runtime: 

https://github.com/user-attachments/assets/d493e520-3a5c-4b4e-92cd-326aeb671a7c


examples/next/src/pages/_app.tsx
```diff
@@ -71,7 +71,7 @@ const App = ({ Component, pageProps }: AppProps) => {
   }, [])
 
   return (
-    <ThemeProvider theme={theme === 'light' ? localLightTheme : localDarkTheme} cssLayer="next-app">
+    <ThemeProvider theme={theme === 'light' ? localLightTheme : localDarkTheme}>
```
jsulpis: did you remove this on purpose or is it a rebase artifact ? If it's on purpose, remove the comment in `global.css`

packages/themes/public/style/dark.css
```diff
@@ -386,7 +408,7 @@
   --typography-body-font-family:
     Inter, system-ui, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Helvetica, Arial, sans-serif;
   --typography-body-font-size: 1rem;
-  --typography-body-font-weight: 400;
```
jsulpis: why did you change that ? `Regular`, `Medium` etc are not valid `font-weight` values. In fact most of the time we use the other `weight` variable which has a number, so these `font-weight` variables seem to be useless duplicates (but the name "font-weight" is consistent with "font-size" so maybe we should remove the "weight" variables)

packages/themes/src/ThemeProvider.tsx
```diff
@@ -31,32 +31,61 @@ type ThemeProviderProps = {
...
+    const cssString = Object.entries(missingVars)
       .map(([key, value]) => `${key}: ${value};`)
-      .join(' ')} }
-      
-      body {
+      .join(' ')
+
+    if (cssString) {
+      const css = `:root { ${cssString} }
+       body {
```
jsulpis, line 46: this rule could now be in `globalStyle.css` and we can remove the cssLayer prop because projects can just apply a layer to the whole `globalStyle.css` if needed. But you can do that in another PR if you prefer

examples/next/styles/global.css
```diff
@@ -1,3 +1,6 @@
+@import'@ultraviolet/themes/light.css' layer(ultraviolet);
+@import'@ultraviolet/themes/light.css' layer(ultraviolet);
```
jsulpis:
```suggestion
@import'@ultraviolet/themes/dark.css' layer(ultraviolet);
```

packages/themes/src/ThemeProvider.tsx
```diff
@@ -31,32 +31,38 @@ type ThemeProviderProps = {
...
-    const cssVars = assignInlineVars(themeContract, theme)
-    const cssString = `:root { ${Object.entries(cssVars)
+    const cssString = Object.entries(cssVars)
       .map(([key, value]) => `${key}: ${value};`)
-      .join(' ')} }
-      
-      body {
+      .join(' ')
+
+    if (cssString) {
```
jsulpis: this `if` was added when filtering the variables but now it's no longer necessary

examples/next/styles/global.css
jsulpis: - this import syntax with the layer doesn't work with webpack. Maybe we can remove the flag and use turbopack since it's what they use in the websites ?
- missing dark and darker ?

examples/next/styles/global.css
jsulpis: you removed the cssLayer, I'm not sure what result is expected

.changeset/eighty-pandas-think.md
```diff
@@ -0,0 +1,15 @@
...
+"@ultraviolet/themes": minor
+---
+
+`ThemeProvider`: do not hash css variable names.
+The name of the CSS variables defined in the ThemeProvider are not hashed anymore. This has a couple of consequences:
+- clearer naming
+- possibility to use the variables without importing css files (`@ultraviolet/themes/light.css` and other variants)
+- reduces FOUC
+
+⚠️⚠️ **Breaking change**: If `@ultraviolet/themes/light.css` (and other variants) is imported, any custom theme defined in the theme provider **will be** overwritten by the .css file (see `examples/next/src/pages/_app.tsx`: to apply the custom *primary text* , `@ultraviolet/themes/light.css` is imported in a layer to reduce its specificity).⚠️⚠️
```
jsulpis, line 11: When I remove the layer syntax in the import in `global.css` I see the custom `colors.primary.text` correctly applied so I don't understand what you mean here

Image

examples/next/styles/global.css
lisalupi: - ok i'll fix it
- wrong copy/paste + darker is not needed the example only switches between dark and light

.changeset/eighty-pandas-think.md
lisalupi: that was not the case on my end when i tried

packages/themes/src/ThemeProvider.tsx
lisalupi: `globalStyle.css` is just a normalization file. Applying a layer to it will not impact ultraviolet tokens

jsulpis: yes but if we move the rule with the background-color on the body in globalStyle.css, then the website can put it in a layer and override the background-color with their style. It's simpler than injecting the css rule at runtime with a cssLayer prop on the ThemeProvider

.changeset/eighty-pandas-think.md
jsulpis: I checked again with your newest changes and the custom theme overrides the ones in the css files without needing the layer. Could you see if you can reproduce the issue you had on your side ? Just to avoid announcing a breaking change that is not one

packages/themes/package.json
```diff
@@ -36,8 +36,22 @@
       "import": "./dist/themes/console/*/index.js",
       "require": "./dist/themes/console/*/index.js"
     },
-    "./global": "./dist/themes.css",
-    "./theme.css": "./dist/themes.css"
+    "./global": {
+      "types": "./dist/global/globalStyle.css.d.ts",
+      "default": "./dist/global/globalStyle.css"
+    },
+    "./normalize": {
```
jsulpis: do you think it's necessary to export `/normalize` in addition to `/normalize.css` ? I would say one syntax is enough and the latter is more explicit

packages/themes/src/index.ts
```diff
@@ -1,5 +1,3 @@
-import './vanilla/globalStyle.css'
```
jsulpis, line 1: you can delete the file I think it's unused (and it duplicates `public/global/globalStyle.css`)

packages/themes/README.md
```diff
@@ -50,17 +50,55 @@ export const App = () => (
 )
 ```
 
+> **Note**:
+> The static CSS imports above are **not** required for the CSS variables to work as long as `ThemeProvider` is used because the provider injects them at runtime. Importing the theme CSS file only **reduces FOUC**: it inlines the variables into the initial HTML, whereas without it the theme is applied client-side after first paint.
+
+##### Wrap the injected styles in a CSS layer
```
jsulpis: this section can be removed

packages/themes/README.md
```diff
@@ -50,17 +50,55 @@ export const App = () => (
...
+```
+
+#### Global style
+
+For a default background-color and text-color, and a `visually-hidden` class that visually hides a component while keeping it accessible to screen readers, you can import a `global` style instead of `normalize`.
+
 ```tsx
 import '@ultraviolet/themes/global'
+// OR
+import '@ultraviolet/themes/global.css'
```
jsulpis: yeah the naming is not very consistent 😕

```suggestion
import '@ultraviolet/themes/theme.css'
```

jsulpis (changes requested):

packages/themes/package.json
lisalupi: i did it to have the same behavior as "global"

jsulpis (approved):

philibea (approved):

PR #6920, opened 2026-09-15
docs(CopyButton): audit accessibility

lisalupi:
### Summary

Audit `CopyButton` accessibility


packages/ui/src/components/Action/CopyButton/__stories__/A11y.mdx
```diff
@@ -0,0 +1,51 @@
...
+## ARIA Pattern
+
+[ARIA Button Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/button/)
+
+## Rules
+
+Enforced by the component:
+
+- ✅ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html): Renders a semantic native `<button>`; the accessible name is provided via `aria-labelledby` pointing to the tooltip label ("Copy" / "Copied!"). Even in icon-only mode the button has an accessible name.
+- ❌ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html): When `children` text is provided (e.g., "Copy id"), the tooltip label ("Copy") is applied via `aria-labelledby` (relation `label`), which overrides the visible text as the accessible name. The announced name therefore differs from the visible label.
```
jsulpis: Under [WCAG Success Criterion 2.5.3 (Label in Name)](https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html), the accessible name can be different from the visible label but then it must _contain_ (ideally start with) the visible label. So I think this criterion should be in the section "To apply when using the component" and not considered a failure

packages/ui/src/components/Action/CopyButton/__stories__/A11y.mdx
```diff
@@ -0,0 +1,51 @@
...
+- ✅ [WCAG 2.1.1 - Keyboard (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html): Native button activation with Enter and Space.
+- ✅ [WCAG 2.4.7 - Focus Visible (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html): Native focus ring preserved by the `Button` component.
+- ✅ [WCAG 1.1.1 - Non-text Content (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html): Icons (`CopyContentIcon`/`CheckIcon`) default to `aria-hidden="true"` and are purely decorative; the state is conveyed by the label.
+- ❌ [WCAG 4.1.3 - Status Messages (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html): The success feedback ("Copied!") is communicated only by changing the button's accessible name — there is no `role="status"` or live region, so the outcome is not announced to screen reader users.
+
+To apply when using the component:
+
+- [WCAG 3.2.4 - Consistent Identification (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/consistent-identification.html): Keep the announced accessible name aligned with the visible label.
```
jsulpis: The criterion "consistent identification" applies to multiple instances of the same component across pages, for the consistency between accessible name and visible label it is 2.5.3, cf my previous comment

packages/ui/src/components/Action/CopyButton/__stories__/A11y.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+## Accessibility issues
+
+**High:**
+
+- No success/status announcement. After activating the button, the "Copied!" confirmation is conveyed only by swapping the accessible name and icon. Screen reader users receive no announcement that the copy succeeded, because the change is not exposed through a live region or `role="status"`. Consider adding an `aria-live`/`role="status"` region (e.g., a visually hidden status element) that announces the copy result on activation.
```
jsulpis: maybe add an `aria-live` property on the Tooltip instead of adding another DOM element ?

packages/ui/src/components/Action/CopyButton/__stories__/A11y.mdx
```diff
@@ -0,0 +1,51 @@
...
+- Accessible name overrides visible text when `children` is used. Because `tooltipLabel` maps to `relation="label"`, the tooltip text becomes the accessible name via `aria-labelledby`, so a button showing "Copy id" is announced as "Copy". Fix: switch between the two tooltip props depending on `children`: use `tooltipDescription` when there is `children` (so the visible text remains the accessible name), and use `tooltipLabel` when there is no `children` (so the icon-only button still has an accessible name).
```
jsulpis: To remove if we agree that it's the user's responsibility to set an accessible name that _contains_ the visible label.

I'm thinking that maybe we could remove the default "copy" tooltip when there is `children`, because 
1. as you said it would be invalid to have a visible "Copy id" and a "Copy" tooltip (the accessible name doesn't contain the visible text)
2. it would be weird to have a tooltip "Copy id" on top of a button with visible "Copy id"

I think if we can just display a temporary "Copied!" tooltip it would provide a clear feedback. But we will have to test carefully using a screen reader because all this logic is not obvious

jsulpis (changes requested):

packages/ui/src/components/Action/CopyButton/__stories__/A11y.mdx
```diff
@@ -0,0 +1,53 @@
...
+# CopyButton - Accessibility
+
+## Description
+
+A button that copies a given value to the clipboard. It renders as a native `<button>` (via the shared `Button` component) with a copy icon and, optionally, text children. Its accessible name and success feedback are both driven by a tooltip label that switches from "Copy" to "Copied!" (with a checkmark icon) for 2.5 seconds after activation.
+
+## Summary
+
+1. ✅ **Perceivable**: Icons are `aria-hidden` by default; color is not the only indicator of state (icon + label change together)
+2. ✅ **Operable**: Native `<button>` with full keyboard support (Enter, Space) and visible focus indicator
```
iManu, line 14: Operable: button is not reachable using keyboard in the CodeEditor block

lisalupi: what do you mean? It is a native `<button>`, tab works to focus it (+ default keyboard interactions)

packages/ui/src/components/Action/CopyButton/__stories__/A11y.mdx
```diff
@@ -0,0 +1,53 @@
...
+3. ✅ **Understandable**: Accessible name reflects the purpose and the copied state
+4. ❌ **Robust**: Copy success/failure is not announced to assistive technologies (no status message)
```
iManu: Robust: it seems ok for me, VoiceOver return "Copied!"

Image

iManu (approved):

packages/ui/src/components/Action/CopyButton/__stories__/A11y.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+**Medium:**
+
+- Copy failures are silently swallowed (`catch(() => null)`). If the clipboard write fails, there is no feedback at all, so users cannot tell the action did not succeed. Expose the failure through a status message as well.
+- Recommended tooltip strategy to enforce correct naming (see [4.1.2](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value), [2.5.3](https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html) and [1.1.1](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content)):
+  - With `children`: do not show a default tooltip (a "Copy" tooltip over "Copy id" both violates 2.5.3 and is redundant). Show only a temporary "Copied!" tooltip as feedback via `tooltipDescription` so the visible `children` text stays the accessible name.
+  - Icon-only (no `children`): keep `tooltipLabel` (`aria-labelledby`) so the tooltip text ("Copy" / "Copied!") provides the accessible name.
+  - Deprecate `copyText` prop: replace with `accessibleName` which displays a tooltip: this would enforce a correct usage (no tooltip for buttons with children, tooltip for icon-only buttons) while keeping the `copiedText` as a description (with aria-live)
```
jsulpis: > replace with `accessibleName`

shouldn't it be `tooltipLabel` like on the Button ?

jsulpis (approved):

packages/ui/src/components/Action/CopyButton/__stories__/A11y.mdx
lisalupi: yes

PR #6924, opened 2026-09-17
feat: tagLink wip

lisalupi:
### Summary

- New component `TagLink`
- Add sentiments to CopyButton

BLOCKED : need update of Link (missing sentiments) #6435


packages/ui/src/components/Badges/TagLink/styles.css.ts
```diff
@@ -0,0 +1,89 @@
...
+const idStackBase = style({})
+
+const idStack = styleVariants(
+  Object.fromEntries(
+    SENTIMENTS.map(sentiment => [
+      sentiment,
+      {
+        backgroundColor: theme.colors[sentiment][sentiment === 'neutral' ? 'backgroundStrong' : 'background'],
+
+        padding: `${theme.space['0.5']} ${theme.space[1]}`,
```
jsulpis: the padding and height don't depend on the sentiment, they should be in the`idStackBase`

packages/ui/src/components/Badges/TagLink/styles.css.ts
```diff
@@ -0,0 +1,89 @@
...
+  default: {},
+})
+
+const copyButton = styleVariants(
+  Object.fromEntries(
+    SENTIMENTS.map(sentiment => [
+      sentiment,
+      {
+        borderRadius: 0,
+        borderRight: '1px solid transparent',
```
jsulpis: you could do a `recipe` and put these two properties in the base to avoid duplicating them in all sentiments ?

packages/ui/src/components/Action/CopyButton/index.tsx
```diff
@@ -44,7 +45,7 @@ export const CopyButton = ({
 
   return (
     <Button
-      className={className}
+      className={cn(className, 'copyButton')}
```
jsulpis: why this hardcoded class ?

packages/ui/src/components/Badges/Tag/styles.css.ts
```diff
@@ -81,6 +81,21 @@ const container = recipe({
         },
       },
     },
+    tagLink: {
```
jsulpis: it's a bit unconventional to have a style for the tagLink here, couldn't it be defined in the tagLink styles ?

packages/ui/src/components/Badges/TagLink/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+            className={cn(onClose ? tagLinkStyle.copyButton[sentiment] : '')}
+          />
+        ) : null}
+      </Stack>
+      {onClose ? (
+        <button
+          data-testid="close-tag"
+          onClick={onClose}
+          type="button"
+          className={tagStyle.container({ isButton: true, sentiment, tagLink: true })}
```
jsulpis: it seems wrong to use `tagStyle.container` here. Could it be just a Button ghost with a Separator on its left ?

packages/ui/src/components/Badges/TagLink/styles.css.ts
```diff
@@ -0,0 +1,89 @@
...
+        selectors: {
+          '&:active': {
+            boxShadow: `inset ${theme.shadows[`focus${capitalize(sentiment)}`]}`,
+          },
+          '&:hover': {
+            borderRight: `1px solid ${theme.colors[sentiment].border}`,
```
jsulpis: so there is:
- a border-left on the close button
- a margin-left: -1px on the close button which makes the copy button paint on top of the border
- we add another border here to compensate for the hidden border

this seems very hacky, I don't know if it comes from the Tag but could we do better on the TagLink ?

packages/ui/src/components/Badges/TagLink/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </Link>
+        {copiable ? (
+          <CopyButton
+            size="xsmall"
+            sentiment={sentiment}
+            value={link}
+            copiedText={copiedText}
+            copyText={copyText}
+            onCopy={onCopy}
+            className={cn(onClose ? tagLinkStyle.copyButton[sentiment] : '')}
```
jsulpis: why is the style conditioned to onClose ?

packages/ui/src/components/Badges/TagLink/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+  loading ? (
+    <Tag isLoading sentiment={sentiment}>
+      {loading}
+    </Tag>
+  ) : (
+    <Text
+      as="span"
+      className={cn(tagLinkStyle.tagLink, tagStyle.container({ sentiment }), className)}
+      variant={variant === 'code' ? 'code' : 'caption'}
+      oneLine
```
jsulpis: I think the wrapper should not be a oneLine Text, it doesn't work when the content is too large. This might also be the reason why the wrapper has overflow hidden which truncates the active state of the copy button 

Image

packages/ui/src/components/Badges/TagLink/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    >
+      <Stack
+        direction="row"
+        gap="0.5"
+        alignItems="center"
+        justifyContent="center"
+        className={cn(tagLinkStyle.idStack[sentiment], tagLinkStyle.idStackBase)}
+      >
+        {icon}
+        {label}
```
jsulpis: in Figma it's called "key", we should have the same naming. Maybe "key" for the text next to the icon and "label" for the text of the link ? (or keep "link" if you prefer and if it's not confusing with the href)

packages/ui/src/components/Badges/TagLink/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </Stack>
+      {onClose ? (
+        <button
+          data-testid="close-tag"
+          onClick={onClose}
+          type="button"
+          className={tagStyle.container({ isButton: true, sentiment, tagLink: true })}
+        >
+          <VisuallyHidden>
+            Close tag {label}: {link}
```
jsulpis: To align with what you did in the Tag ? (if so, update A11y.mdx)

```suggestion
            Remove tag {label}: {link}
```

packages/ui/src/components/Badges/TagLink/__stories__/A11y.mdx
```diff
@@ -0,0 +1,45 @@
...
+# TagLink - Accessibility
+
+## Description
+
+A badge-style component that pairs an identifier (a `label`, with optional `icon`) with a `link`, and can additionally render a copy button and a close button, plus a `loading` state.
+
+## Summary
+
+1. ✅ **Perceivable**: decorative icons are `aria-hidden` by default; any text/icon contrast violation needs to be fixed by a tokens update (see [Ultraviolet's color constrast checker](https://storybook.ultraviolet.scaleway.com/?path=/docs/customization-color-contrast--docs)).
+2. ✅ **Operable**: all interactions use native `<a>`/`<button>` elements which are keyboard accessible with visible focus indicators (default) and a logical focus order.
```
jsulpis: the focus indicator on the copy button is a bit hidden, I think ideally we don't want an overflow hidden on the container to keep the focus indicator fully visible 

Image

jsulpis (changes requested):

packages/ui/src/components/Badges/TagLink/index.tsx
lisalupi: well it needs the style of a tag and it's not really a ghost button, on figma it doesn't seem to use the Button component either

packages/ui/src/components/Badges/TagLink/index.tsx
lisalupi: i'm not sure about `key`  since it is already used in react

PR #6928, opened 2026-09-18
fix(Tag): accessibility

lisalupi:
### Summary

Fix `Tag` accessibility


packages/ui/src/components/Badges/Tag/index.tsx
```diff
@@ -118,20 +120,24 @@ export const Tag = ({
...
         {onClose ? (
           <button
-            aria-label="Close tag"
             data-testid="close-tag"
             disabled={disabled}
             onClick={onClose}
             type="button"
             className={tagStyle.container({ disabled, isButton: true, sentiment })}
           >
+            <VisuallyHidden>Remove tag {stringChildren}</VisuallyHidden>
```
philibea, line 133: missing translation for ?

packages/ui/src/components/Badges/Tag/__tests__/a11y.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  })
+
+  describe('disabled state', () => {
+    it('announces the disabled state on a non-copiable tag', () => {
+      const { container } = renderWithTheme(<Tag disabled>test</Tag>)
+
+      expect(container.querySelector('[aria-disabled="true"]')).toHaveTextContent('test')
+    })
+
+    it('does not expose aria-disabled when not disabled', () => {
```
jsulpis, line 63: this is true when `disabled` is undefined, but if `disabled={false}` then the attribute `aria-disabled="false"` is present. We can see it in the snapshots from the `TagInput` which sets the disabled prop to false. 

I personally like to avoid `aria-disabled="false"` if it adds no value (`aria-disabled={disabled ? 'true' : undefined}`) but that's no big deal and since we have this in many components we can work on that later.

But in any case, the name of the test is misleading

packages/ui/src/components/Badges/Tag/index.tsx
```diff
@@ -118,20 +120,24 @@ export const Tag = ({
...
             <CloseIcon size="small" />
           </button>
         ) : null}
+        {isCopiable ? (
+          // oxlint-disable-next-line jsx-a11y/prefer-tag-over-role
+          <VisuallyHidden role="status">{isCopied ? copiedText : null}</VisuallyHidden>
```
jsulpis, line 139: I'm not sure if it would work but could we instead add an `aria-live` prop to the Tooltip so that tooltip updates are vocalized ? I wonder if it should even be by default on the Tooltip

packages/ui/src/components/Badges/Tag/__stories__/A11y.mdx
```diff
@@ -11,50 +11,58 @@ A metadata component that displays a short text description of an item (category
 ## Summary
 
 1. ✅ **Perceivable**: decorative icons are `aria-hidden` (resolved by the Icon component fix); sentiment contrast is correct.
-2. ❌ **Operable**: the disabled copiable button remains focusable even though it performs no action.
-3. ❌ **Understandable**: the copiable button lacks `type="button"` and may unexpectedly submit a parent form.
-4. ❌ **Robust**: `aria-disabled` is placed on a decorative element instead of the interactive button; the "Copied!" status message is not announced (4.1.3).
+2. ✅ **Operable**: the disabled copiable button remains focusable even though it performs no action.
```
jsulpis, line 14: this line and the next should be updated to reflect the new correct behavior

packages/ui/src/components/Badges/Tag/index.tsx
lisalupi: I can add a prop for that yes

packages/ui/src/components/Badges/Tag/index.tsx
lisalupi: I added `aria-live` to the tooltip, but left it as `undefined` by default

.changeset/humble-groups-film.md
```diff
@@ -0,0 +1,9 @@
+---
+"@ultraviolet/ui": patch
+---
+
+`Tag`: fix accessibility issues
+
+`Separator`: new prop `aria-hidden`
+
+`Tooltip`: new prop `aria-polite`
```
jsulpis:
```suggestion
`Tooltip`: new prop `aria-live`
```

packages/ui/src/components/Badges/Tag/__stories__/A11y.mdx
```diff
@@ -11,50 +11,58 @@ A metadata component that displays a short text description of an item (category
 ## Summary
 
 1. ✅ **Perceivable**: decorative icons are `aria-hidden` (resolved by the Icon component fix); sentiment contrast is correct.
-2. ❌ **Operable**: the disabled copiable button remains focusable even though it performs no action.
-3. ❌ **Understandable**: the copiable button lacks `type="button"` and may unexpectedly submit a parent form.
-4. ❌ **Robust**: `aria-disabled` is placed on a decorative element instead of the interactive button; the "Copied!" status message is not announced (4.1.3).
+2. ✅ **Operable**: the disabled copiable button is not focusable.
+3. ✅ **Understandable**: the copiable button has `type="button"` to avoid unexpected submitting a parent form.
+4. ✅ **Robust**: the "Copied!" status message is announced via a `role="status"` live region (4.1.3).
```
jsulpis:
```suggestion
4. ✅ **Robust**: the "Copied!" status message is announced via a live region in the tooltip element.
```

packages/ui/src/components/Badges/Tag/__stories__/A11y.mdx
```diff
@@ -11,50 +11,58 @@ A metadata component that displays a short text description of an item (category
...
-- ❌ [WCAG 4.1.3 - Status Messages (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html) - The "Copied!" feedback is only shown via a tooltip, which is not announced by screen readers.
+- ✅ [WCAG 3.2.2 - On Input (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/on-input.html) - The copiable `<button>` has `type="button"` attribute.
+- ✅ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) - The removable (close) button's has an explicit label that includes the visible tag text.
+- ✅ [WCAG 4.1.2 - Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) - The copiable button's accessible name is the tag text, and the tooltip is wired as its accessible description. On focus, the action is conveyed as a description, so the button's purpose is programmatically determinable.
+- ✅ [WCAG 4.1.3 - Status Messages (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html) - The "Copied!" feedback is announced to screen readers through a visually hidden `role="status"` live region rendered for copiable tags.
```
jsulpis:
```suggestion
- ✅ [WCAG 4.1.3 - Status Messages (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html) - The "Copied!" feedback is announced to screen readers through a live region in the tooltip.
```

packages/ui/src/components/Badges/Tag/index.tsx
```diff
@@ -65,16 +67,15 @@ export const Tag = ({
...
     variant: variant === 'code' ? 'code' : 'caption',
     oneLine: true,
     sentiment: disabled ? 'neutral' : sentiment,
-    disabled: disabled,
+    disabled,
   } as const
 
   return (
-    <Tooltip text={isCopiable ? copyTextTooltip : null}>
+    <Tooltip text={isCopiable ? copyTextTooltip : null} relation="description" aria-live="polite">
```
jsulpis: VoiceOver says 2 times "Value copied Value copied" and then "Click to copy Click to copy", is it the same for you ? Maybe the Tooltip element should be in the DOM from the beginning, or maybe the aria-live isn't the right approach I'm not sure 😕

lisalupi: Ah yes indeed, but it only says it once on Safari... 
I'll see what I can do, but i'll revert to VisuallyHidden if aria-live doesn't work without needing major work on Tooltip

packages/ui/src/components/Badges/Tag/__stories__/A11y.mdx
lisalupi: reverted to visually hidden

packages/ui/src/components/Badges/Tag/__stories__/A11y.mdx
lisalupi: reverted to visually hidden

packages/ui/src/components/Badges/Tag/index.tsx
lisalupi: reverted to visually hidden

PR #6929, opened 2026-09-18
chore(oxlint): remove react warn rules — batch 5

philibea:
### Summary

Resolve all `react/*` warnings from batch 5 of the oxlint warn-rule cleanup plan, then remove the `warn` overrides from `oxlint.config.ts` so the base config (`error` severity) applies with zero violations.

Rules resolved:
- `react/no-object-type-as-default-prop` (1) — replaced array default prop `keyvalues = []` with `undefined` + local `keyvaluesList` variable
- `react/static-components` (2) — inlined conditional rendering in `SelectableCard` instead of creating a component inside render; moved `SubComponent` out of story render in `Table/Context.stories`
- `react/hooks` (2) — fixed by moving `SubComponent` out of the story render function
- `react/preserve-manual-memoization` (3) — fixed `useMemo`/`useCallback` dependency arrays to match React Compiler inference (refs are stable, use whole `value` object instead of partial access)


packages/ui/src/components/Data Entry/KeyValueInput/index.tsx
```diff
@@ -32,19 +32,21 @@ export const KeyValueInput = ({
...
   addButton,
   readOnly,
-  keyvalues = [],
+  keyvalues,
   maxSize = 100,
   fieldErrors,
 }: KeyValueInputProps) => {
   const errorId = useId()
 
+  const keyvaluesList = keyvalues ?? []
```
jsulpis: see the [doc of the rule](https://oxc.rs/docs/guide/usage/linter/rules/react/no-object-type-as-default-prop.html), the goal is to avoid redefining default values at each render and use a stable ref instead, this is not what was done here. You should instead create a constant out of the composant and use it as the default value

packages/ui/src/components/Data Entry/SelectInput/components/Dropdown/Dropdown.tsx
```diff
@@ -134,7 +134,8 @@ export const Dropdown = ({
     ) {
       setWidth(refSelect.current.getBoundingClientRect().width)
     }
-  }, [refSelect, maxWidth])
+    // oxlint-disable-next-line react-hooks/exhaustive-deps -- refSelect is a stable ref
```
jsulpis: is it harmful to keep it ? I would rather avoid a disable comment each time we need a ref in a dependency array

oxlint.config.ts
```diff
@@ -194,8 +194,8 @@ export default defineConfig({
     'react/display-name': 'off',
     'react/exhaustive-effect-dependencies': 'warn', // react errors (30)
     'react/forbid-component-props': 'off',
+    'react/function-component-definition': 'warn', // react errors (964)
```
jsulpis: why was this added ?

jsulpis (changes requested):

packages/ui/src/components/Data Entry/SelectInput/components/Dropdown/Dropdown.tsx
philibea: this was related to this rule in fact```// oxlint-disable-next-line  react/preserve-manual-memoization```

lisalupi (approved):

packages/ui/src/components/Data Entry/KeyValueInput/index.tsx
```diff
@@ -32,19 +34,21 @@ export const KeyValueInput = ({
   inputValue,
   addButton,
   readOnly,
-  keyvalues = [],
+  keyvalues,
```
jsulpis: is it possible to use this syntax ? This avoids to redefine another variable `keyvaluesList`, in the component there is a risk of using the wrong variable

```suggestion
  keyvalues = DEFAULT_KEY_VALUE_LIST,
```

packages/ui/src/components/Data Entry/SelectInput/components/Dropdown/Dropdown.tsx
```diff
@@ -134,6 +134,7 @@ export const Dropdown = ({
     ) {
       setWidth(refSelect.current.getBoundingClientRect().width)
     }
+    // oxlint-disable-next-line  react/preserve-manual-memoization
```
jsulpis: could we replace `refSelect` with `refSelect.current` as the warning message suggests ?

oxlint.config.ts
```diff
@@ -172,7 +172,7 @@ export default defineConfig({
     'react/display-name': 'off', // react errors (4)
     'react/forbid-component-props': 'off', // react errors (229)
     'react/jsx-max-depth': ['error', { max: 10 }], // base sets max: 8
-    'react/jsx-no-useless-fragment': 'off', // react errors (3)
+    // 'react/jsx-no-useless-fragment': 'off', // react errors (3)
```
jsulpis: delete the commented rules, in this file and the oxlint-fix config

jsulpis (approved):

PR #6930, opened 2026-09-18
chore(oxlint): remove typescript warn rules — batch 6

philibea:
### Summary

Resolve all `typescript/*` warnings from batch 6 of the oxlint warn-rule cleanup plan, then remove the `warn` overrides from `oxlint.config.ts` so the base config (`error` severity) applies with zero violations.

Rules resolved:
- `typescript/unified-signatures` (1) — merged two call overloads into a single signature
- `typescript/parameter-properties` (1) — moved constructor parameter property to explicit class field
- `typescript/no-invalid-void-type` (1) — kept `void` in union with inline disable (needed for `onSubmit` callbacks that don't return a value)
- `typescript/no-unnecessary-boolean-literal-compare` (1) — removed `=== true`
- `typescript/prefer-optional-chain` (1) — used optional chain `?.`
- `typescript/prefer-regexp-exec` (1) — switched `String#match()` to `RegExp#exec()`
- `typescript/no-misused-promises` (1) — wrapped async `onChange` handler with `.catch()`
- `typescript/consistent-indexed-object-style` (3) — replaced index signatures with `Record`


packages/ui/src/components/Data Display/List/index.tsx
```diff
@@ -97,13 +97,9 @@ const BaseList = forwardRef<HTMLTableElement, NewListProps | LegacyListProps>(
  * List is a component that displays a list of items based on the columns you provide and the data you pass.
  */
 type ListType = {
-  // biome-ignore  lint/style/useUnifiedTypeSignatures: ok
-  (props: NewListProps & RefAttributes<HTMLTableElement>): ReactNode
-  /**
-   * @deprecated Use `colMode="strict"`
-   */
-
```
lisalupi, line 105: don't we lose this deprecation message?

philibea: nice catch. Change to right place :)

lisalupi (approved):

PR #6931, opened 2026-09-18
fix(lint): restore arrow-function options for react/function-component-definition

philibea:
### Summary

The `react/function-component-definition` override in `oxlint.config.ts` was set to `'warn'` without options, which caused oxlint to drop the `{ namedComponents: 'arrow-function', unnamedComponents: 'arrow-function' }` options inherited from `@scaleway/oxlint-config`. As a result, oxlint fell back to its defaults (`function-declaration` / `function-expression`) and flagged all 964 arrow-function components in the codebase as violations — even though arrow functions are the enforced convention.

Removing the override lets the base config's `error`-level rule with the correct `arrow-function` options take effect. Two source files that genuinely used `function` declarations were converted to arrow functions to comply:

- `.storybook/preview.tsx` — `DottedBackground`
- `packages/ui/src/components/Data Entry/SelectInput/__tests__/index.test.tsx` — `ControlledSelect`

Result: 0 `react/function-component-definition` warnings (down from 964).


oxlint.config.ts
```diff
@@ -226,7 +225,7 @@ export default defineConfig({
 
     'typescript/consistent-indexed-object-style': 'warn', // typescript errors (3)
     'typescript/consistent-return': 'warn', // typescript errors (8)
-    'typescript/consistent-type-definitions': 'warn', // typescript errors (310)
+    'typescript/consistent-type-definitions': ['warn', 'type'], // typescript errors (0)
```
lisalupi: if 0 error, it can be removed

philibea: @lisalupi removed

jsulpis (approved):

lisalupi (approved):

PR #6932, opened 2026-09-18
chore(oxlint): remove jsx-a11y warn rules — batch 3

philibea:
### Summary

Resolve all `jsx-a11y/*` warnings from batch 3 of the oxlint warn-rule cleanup plan, then remove the `warn` overrides from `oxlint.config.ts` so the base config (`error` severity) applies with zero violations.

Rules resolved:
- `jsx-a11y/no-autofocus` (1) — inline disable in UnitInput (intentional focus behavior)
- `jsx-a11y/no-noninteractive-element-interactions` (1) — inline disable in PieChart Legends (legend items sync hover/focus to chart segments)
- `jsx-a11y/no-noninteractive-tabindex` (1) — inline disable in TagInput
- `jsx-a11y/no-static-element-interactions` (6) — inline disables in TagInput, Popup, Dialog, Avatar, BarStack (elements with conditional roles or overlay/presentation layers)


packages/ui/src/components/Data Entry/UnitInput/index.tsx
```diff
@@ -123,6 +123,7 @@ export const UnitInput = ({
           <input
             aria-describedby={ariaDescribedBy || (hasHelperText(helper, error, success) ? helperId : undefined)}
             aria-invalid={!!error}
+            // oxlint-disable-next-line jsx_a11y/no-autofocus
```
jsulpis: if we want to authorize autofocus we might as well disable the rule no ?

jsulpis (approved):

packages/ui/src/components/Data Display/BarStack/index.tsx
```diff
@@ -107,6 +107,7 @@ export const BarStack = ({
                 <Tooltip text={legend === 'outside' ? text : undefined} visible={hoveredBarId === id}>
                   {tooltip ? (
                     <Tooltip id={`tooltip-${id}`} text={tooltip ?? text}>
+                      {/* oxlint-disable-next-line jsx_a11y/no-static-element-interactions */}
```
lisalupi: if we disable it everywhere instead of fixing it, why not remove it?

lisalupi (approved):

philibea:
@jsulpis @lisalupi i think we should keep the rule active and comment when necessary. We have the information like this with a possible comment if it's normal or not handle by the component but we should keep this as error to have the information that doesn't respect some of oxlint rules wdyt ?

PR #6933, opened 2026-09-18
chore(oxlint): remove unicorn warn rules — batch 4

philibea:
### Summary

Resolve all `unicorn/*` warnings from batch 4 of the oxlint warn-rule cleanup plan, then remove the `warn` overrides from `oxlint.config.ts` so the base config (`error` severity) applies with zero violations.

Rules resolved:
- `unicorn/prefer-top-level-await` (1) — replaced promise chain with top-level `await import` in `.storybook/preview.tsx`
- `unicorn/no-array-reverse` (2) — `.reverse()` → `.toReversed()`
- `unicorn/prefer-spread` (2) — `.split('')` → `[...str]`
- `unicorn/prefer-number-properties` (3) — `parseInt` → `Number.parseInt`
- `unicorn/no-array-sort` (6) — `.sort()` → `.toSorted()`


packages/ui/src/components/Overlay/Modal/components/Dialog.tsx
```diff
@@ -110,7 +110,7 @@ export const Dialog = ({
   // We need to reverse the array as the last opened modal should be the first to be with normal size
   // while the first opened modal should shrink
   const realPosition = [...openedModals].findIndex(object => object.id === id)
-  const position = [...openedModals].reverse().findIndex(object => object.id === id) // reverse method mutate array so we need to create a new array
+  const position = [...openedModals].toReversed().findIndex(object => object.id === id)
```
jsulpis: there is no mutation so no need to create a new array, and same thing on the line before
```suggestion
  const position = openedModals.toReversed().findIndex(object => object.id === id)
```

packages/utils/src/__tests__/shuffle.test.ts
```diff
@@ -16,7 +16,7 @@ describe(shuffle, () => {
     const result = shuffle(input)
 
     expect(result).toHaveLength(input.length)
-    expect([...result].sort()).toStrictEqual([...input].sort())
+    expect([...result].toSorted()).toStrictEqual([...input].toSorted())
```
jsulpis, line 19: same l.33

```suggestion
    expect(result.toSorted()).toStrictEqual(input.toSorted())
```

jsulpis (approved):

lisalupi (approved):

packages/utils/src/__tests__/shuffle.test.ts
philibea: toSorted doesn't exist for a string. The idea is to compare `abcdef` which has be shuffle so [...'abcdef'] => ['a','b'...etc] and at the end that will compare that we have the same "string"  after the shuffle.

package.json
```diff
@@ -154,8 +154,9 @@
   },
   "engines": {
     "node": ">=20.x",
-    "pnpm": ">=10.x"
+    "pnpm": ">=12.x"
   },
+  "packageManager": "pnpm@12.5.1",
```
jsulpis: why those changes ?

packages/utils/src/__tests__/shuffle.test.ts
jsulpis: you updated one of the four occurrences

package.json
philibea: worktree issue => pr comming

PR #6934, opened 2026-09-18
chore(oxlint): remove node+oxc warn rules — batch 2

philibea:
### Summary

Resolve all `node/*` and `oxc/*` warnings from batch 2 of the oxlint warn-rule cleanup plan, then remove the `warn` overrides from `oxlint.config.ts` so the base config (`error` severity) applies with zero violations.

Rules resolved:
- `node/callback-return` (2) — added `return` before callback calls in `useIsOverflowing`; renamed `callback` param in `useOnFieldChange` (not a Node-style error-first callback)
- `oxc/no-map-spread` (1) — replaced spread-in-map with direct object construction in `TagInput`
- `oxc/branches-sharing-code` (6) — hoisted shared code out of `if/else` blocks in `Dropdown`, `SearchBar`, `ToggleGroupField`, `Group`, and `Row`


lisalupi (approved):

packages/ui/src/components/Data Entry/TagInput/index.tsx
```diff
@@ -29,7 +29,7 @@ type StatusValue = (typeof STATUS)[Keys]
 const convertTagArrayToTagStateArray = (tags?: TagInputProp) =>
   (tags ?? [])?.map((tag, index) =>
     typeof tag === 'object'
-      ? { ...tag, index: getUUID(`tag-${index}`) }
+      ? { label: tag.label, index: getUUID(`tag-${index}`) }
```
jsulpis: would be simpler and less error prone I think
```suggestion
     { label: typeof tag === 'object' ? tag.label : tag, index: getUUID(`tag-${index}`) }
```

jsulpis (approved):

